PR #48, opened 2012-08-21
Add Ashima Group to Users list

dsheets:


Chris00:
Thanks for your patch.  Could you please provide a small description of what you do with OCaml — and possibly how it was beneficial to your company?

PR #57, opened 2012-09-27
Replace "News from INRIA" by a small code snippet

gasche:
The "News from INRIA" feed was relatively outdated and, I think, used free space on the front page for something that wasn't really worth it (if one want to underline INRIA's role in OCaml development, I'm sure there are better ways to do, such as putting more emphasis on the research around OCaml).

I replaced it with a "minimum viable snippet" that shows a small example of OCaml code that is, I think, a reasonable compromise between "simple to understand" and "interesting". I think it's important to have such a snippet somewhere on the front page, but I'm not insisting on either this location on the page, or this particular snippet (which could be replaced by a simpler one).

In the process of putting this snippet, I've added a few options to the "binding" for OCaml code highlighting (having the toplevel output around like for all other code examples would consume too much space).

### DIFF
--- a/src/app/build.ml
+++ b/src/app/build.ml
@@ -133,13 +133,31 @@ module Toc = struct
 
 end
 
+let safe_bind binding name renderer =
+  (* We display the returned expression on stderr, but not in the
+     rendered page. The exception may leak information and, more
+     importantly, the errors we get are often of the form "unable to
+     parse [this huge HTML bloat we got from a remote server]" and
+     displaying this is not very helpful. *)
+  let safe_renderer args ~content page =
+    try renderer args ~content page 
+    with exn ->
+      let call = 
+        sprintf "%s %s" name
+          (String.concat " " args) in
+      let msg = sprintf "Error on binding %S" call in
+      prerr_endline (sprintf "%s : %s" msg (Printexc.to_string exn));
+      [Data (sprintf "<span class=\"binding_error\">%s</span>\n" msg)]
+  in
+  Weberizer.Binding.fun_html binding name safe_renderer
+
 let () =
   let b = Weberizer.Binding.make() in
-  Weberizer.Binding.fun_html b "rss" Render_rss.of_urls;
-  Weberizer.Binding.fun_html b "news" Render_rss.news;
-  Weberizer.Binding.fun_html b "opml" Render_rss.OPML.of_urls;
-  Weberizer.Binding.fun_html b "toc" Toc.make;
-  Weberizer.Binding.fun_html b "ocaml" Code.eval_ocaml;
+  safe_bind b "rss" Render_rss.of_urls;
+  safe_bind b "news" Render_rss.news;
+  safe_bind b "opml" Render_rss.OPML.of_urls;
+  safe_bind b "toc" Toc.make;
+  safe_bind b "ocaml" Code.eval_ocaml;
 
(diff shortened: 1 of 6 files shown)


agarwal:
Regarding commit ac6429:
I agree with this. Indeed the Inria feed is subscribed in the planet, so it
will show up there. Or does it? Christophe did you filter it out because we
once discussed that it was poor to have it appear in both places?

Regarding commit 362f06:
I like it!

Regarding commit 197fdd and 691ce4:
This seems good to me, but Christophe knows this part best, so let's get
his input.

gasche:
Hm, I'm not exactly sure why github automatically added this commit I just pushed on my repo; it would rather make sense as a separate pull request as it's an orthogonal matter, but indeed I'd also request for its inclusion upstream.

Chris00:
This is the way github works.  You have to create different branches to submit different pull requests.  In any case, I pulled (before https://github.com/gasche/ocaml.org/commit/782f71b1d5668053ec09c3f2c1819d25ed0412c0) from your repo to integrate your commits, so the two will be indeed treated as separate merges.

Chris00:
@gasche I have integrated your first pull request.

Chris00:
@gasche I have handled the exceptions raised by bindings in Weberizer (mostly commit https://github.com/Chris00/weberizer/commit/e4038ea509f8d6bf0240012a042d58cfb69574a3).
I have adapted the code of ocaml.org accordingly https://github.com/ocaml/ocaml.org/commit/57a37c23c5de5e9725b059e84c809a602d25d1ac

gasche:
Thanks. I think having it in Weberizer directly is good, but then you should make sure that the users are aware that they may get error messages in the web page rendering.

On a tangent matter, I'm not fond of your phantom type API for `Context`. It smells over-engineering, and it's a tad complex for the users. I would be in favor of exposing the record directly (abstract product type don't actually strengthen against adding new fields), but if I understand correctly you want to be able to have different set of fields in different cases. What about plain objects, used as extensible records? `Fun_html of (< content : html; page : html > -> ...) -> ...`.

Chris00:
On Sat, 29 Sep 2012 10:50:24 -0700, gasche wrote:

> Thanks. I think having it in Weberizer directly is good, but then
> you should make sure that the users are aware that they may get
> error messages in the web page rendering.

At some point I was wondering about exiting the "build" program with a
value $B!b(B0 if an exception was raised during the compilation of pages $B!=(B
so that Make stops with an error.  However, the images will not then
be copied into www which is not completely satisfactory.  Another
"solution" is to prominently display a message at the end of the build
if an exn was raised.  What do you think?

> On a tangent matter, I'm not fond of your phantom type API for
> Context. It smells over-engineering, and it's a tad complex for the
> users. I would be in favor of exposing the record directly (abstract
> product type don't actually strengthen against adding new fields),
> but if I understand correctly you want to be able to have different
> set of fields in different cases. What about plain objects, used as
> extensible records? Fun_html of (< content : html; page : html > ->
> ...) -> ....

The problem is that if you are in a string context, then getting the
content of the HTML tag is meaningless.  I want the type system to
ensure that accesses are meaningful $B!=(B so I will not be exposing the
record.  Objects may be an alternative solution.  Do you think it
would be easier to decipher?  Anyway, I will experiment with objects.
If I am happy with the result, I may keep that solution.

Thanks for your comments,
C.

Chris00:
I have moved to objects.

PR #59, opened 2012-09-28
Remove two references to an old ocaml.org repository

pquentin:
https://github.com/agarwal/ocamlweb gives a 404, so I used https://github.com/ocaml/ocaml.org instead.

### DIFF
--- a/src/html/companies.html
+++ b/src/html/companies.html
@@ -404,7 +404,7 @@
       <small>Appearance of a company's name here does not necessarily
       imply endorsement by that company of OCaml or of the
       descriptions provided here. Company representatives should <a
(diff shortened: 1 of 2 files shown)


agarwal:
Fixed in 807d0b9d89da3aa98f8777be914f1300a223a1a6. `OCamlWeb_Main.ml` is generated from `OCamlWeb_Main.html`.

pquentin:
Oops, I missed those ml files. I didn't understand why I couldn't find the footer, which is the initial reason for this patch.

agarwal:
I'll try to add some documentation about how the source code is organized,
and what does what. The point is that weberizer processes the html files,
and adds/modifies some things. Content contributors usually don't have to
worry about this and can just modify the html files. And thanks for
contributing!

PR #70, opened 2012-12-20
Fixed an omitted '#' for a directive call. Updated weberizer executable name.

xandkar:


Chris00:
The version of weberizer in OPAM seems to use the old name.  I'll make a new release soon.

Chris00:
Please submit a pull request for the missing “#” only.

xandkar:
Not sure that I understand - which is the outdated name, 'weberizer_compile' or 'weberizer'? Both, the latest OPAM build and a fresh clone & build of weberizer repo, seem to be using 'weberizer_compile' name.

agarwal:
The new name is 'weberizer'. This is what it's called in the latest head of
the weberizer repo (you must not be getting the latest head). But this
version is not yet released formally in opam.

xandkar:
Got it. Just submitted another pull request for "#" only.

However, weberizer issue is still a mystery - I see your commit that changed the name, but the build process still produces an executable named "weberizer_compile.native". What else can I be missing?

agarwal:
That's the executable in your working directory, but oasis renames it at
install time. So do `make install`. Where it gets installed depends on what
--prefix option you used when doing `ocaml setup.ml -configure`. I
recommend setting --prefix to the root of your ocaml installation. Then all
your ocaml related stuff gets installed under one directory.

Or better yet, if you're using opam (which you should be), just do this:

opam pin weberizer /path/to/local/weberizer
opam install weberizer

and opam will do its magic! When Christophe updates the opam-repository,
you'll do

opam pin weberizer none
opam update
opam upgrade

and you'll no longer be using your local copy.

Chris00:
Weberizer 0.7.2 is now in opam-repository.

xandkar:
Cool. Just updated.

Thank you for the detailed help, Ashish!

Still, all of that just to get a predictable filename? My spidey senses are tingling... :)

PR #71, opened 2012-12-20
Updated info about Mr. Number and Skydeck

mjambon:
Skydeck is now the name of our mobile app for meeting new people.


Chris00:
Thanks for your contribution Martin.   Maybe you could also add a description (what the companies do, what they use OCaml for)?  That would be really nice.

Issue #74, opened 2012-12-22
tabs in install.html page are so far down the screen you don't see them


mk270:
I didn't realise they were there. I only became aware of them when checking a diff of a pull request from several months ago, as they were off the bottom of the screen.

We should just get rid of them, pretty as they may be, and split this stuff into separate pages.

agarwal:
Duplicate of Issue #75. Closing.

Issue #39, opened 2012-08-18
download.html: add more information on releases


agarwal:
Content of [latest](http://caml.inria.fr/ocaml/release.en.html), [previous](http://caml.inria.fr/ocaml/release-prev.en.html), and [older](http://caml.inria.fr/pub/distrib/) releases should be imported directly into releases.html.

agarwal:
Content of download.html on current Inria site duplicates information on individual release pages, which have now been ported fully. Thus, will not port download.html itself.

PR #75, opened 2012-12-22
make the installation options findable on the page

mk270:
should go some way to resolving #74

### DIFF
--- a/src/html/faq.html
+++ b/src/html/faq.html
@@ -59,6 +59,19 @@
 	</div>
 
 	<div class="Q">
+	  Under what licensing terms is the OCaml software available?
+	</div>
+	<div class="A">
+	  The OCaml system is open source software: the compiler is
+      distributed under the terms of the Q Public License, and its
+      library is under LGPL; please read the 
+	  <a href="license.html">license</a> document for more
+      details.  A BSD-style license is also available for a fee
+      through the <a href="http://caml.inria.fr/consortium/index.en.html">
+	  OCaml Consortium</a>.
+	</div>
+
+	<div class="Q">
 	  What is the meaning of the name &#8220;OCaml&#8221;
 	</div>
 	<div class="A">
--- a/src/html/install.html
+++ b/src/html/install.html
@@ -11,29 +11,7 @@
 
     <h1>Install OCaml</h1>
 
-    <div class="pull-right" style="margin-left: 1em" >
-      <div class="span3" >
-	<div class="big-announcements" >
-	  <div class="hero-unit">
-	    See <a href="whats_new.html"><strong>what's new</strong></a>
-	    in the latest version!
-	  </div>
-	</div>
-
-	<div class="hero-unit" >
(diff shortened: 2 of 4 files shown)


Chris00:
Your purpose is no clear to me.  You remove a bunch of information but do not propose to put them elsewhere.  I do not understand for example why you remove the link to the licence.

mk270:
Ok, so I _literally_ did not find most of the content on the page, the first seven or eight times I looked at it, and only discovered it by accident when browsing the revision log.

Here's a screenshot of what it looks like on my screen: http://mk.ucant.org/media/screenshot.jpg

That's in 1280x800 on a normal laptop.

So, I tried to remove information which a user didn't need to know to install OCaml: previous versions, what is new in the _current_ version(!), the licensing arrangements (which they can consult at their own initiative if so inclined), and a bunch of adverts for alternative Javascript targets. This stuff really just is not that important - it can be linked from other pages: the installation are complicated enough by themselves without having two non-installation boxes on the right hand side about the latest features and the terms under which you can redistribute the compiler.

mk270:
In case it wasn't clear: removing some of the information above the tabs meant that the information below the menu bar was now visible, meaning it got noticed. A bunch of tabs at the bottom of the page was something I mistook for boilerplate rather than content.

mk270:
Ok, so it seems the licence and the advertisement for the BSD licensing arrangement aren't present elsewhere on the site(!) .. I didn't realise this

mk270:
Ok, I've re-added the licensing info - had assumed it was duplicated elsewhere

agarwal:
Thanks for the input. I see your point. The tabs are the main content but
not very prominent. This page has mixed up two separate items: listing the
releases available, and providing installation instructions. Let's do this:
- Create a new Releases page, which is related to Issue #39 and partially
  undoes commit 2fda77b972002e684f244211237ccc2081ba90a0.
- Simplify the opening text as Martin proposed, moving text about different
  releases to the above page. I don't want to just delete the text;
  js_of_ocaml, ocamljava, and older releases should be mentioned somewhere.
- Move the licensing link to the new page too. The license is too important
  to be only in the FAQ.
- With that done, we can again look at the Install page and decide if the
  tabs are worth it. It maybe better just as flat text with headers and
  sub-headers. A table of contents can be auto-generated (e.g.
  http://ocaml.org/taste.html), which might be even more clear than the tabs.

Chris00:
On Sat, 22 Dec 2012 11:57:54 -0800, Ashish Agarwal wrote:

> Thanks for the input. I see your point. The tabs are the main
> content but not very prominent. This page has mixed up two separate
> items: listing the releases available, and providing installation
> instructions. Let's do this: [...]

I meant to reply something similar.  I agree with these points — we do
not have to loose information for readability, we want both!

mk270:
Ok, is that better?

agarwal:
Looks good. I've merged all your changes.

Issue #90, opened 2013-01-10
Broken links


voila:
I've run the W3C link checker on the website 

http://validator.w3.org/checklink?uri=ocaml.org&hide_type=all&depth=&check=Check

and here's the list of broken links:
- [x] http://ocaml.org/tutorials/hashtbl.html
- [x] http://ocaml.org/tutorials/map.html
- [x] http://ocaml.org/tutorials/set.html
- [x] http://ocaml.org/tutorials/comparison_of_standard_containers.html
- [x] http://ocaml.org/caml.inria.fr/pub/docs/manual-ocaml/libref/Pervasives.html (via beaf7a696b33311febebac06574287c27819837c)
- [x] http://ocaml.org/caml.inria.fr/pub/docs/manual-ocaml/libref/Printf.html (via beaf7a696b33311febebac06574287c27819837c)
- [x] http://ocaml.org/caml.inria.fr/pub/ml-archives/caml-list/2003/06/2732fbf1bed1ae50fefb6ea410ea9955.en.html (via c77d3ae0fcafc62731260eab57e20c169389c01d)
- [x] http://www.noir-medical.com/graphics/diode_graph.gif (via 67ca5e6b4beb8b5a19342880878b283ddaa12d77)
- [x] http://ocaml.org/tutorials/ocaml_and_the_web/discussion.html (via c6edb48cf6423f18a3502924da36db6b04f6bc48)
- [x] http://ocaml.org/caml.inria.fr/pub/ml-archives/caml-list/2008/09/2bc9b38171177af5dc0d832a365d290d.en.html (via c77d3ae0fcafc62731260eab57e20c169389c01d)
- [x] http://ocaml.org/caml.inria.fr/pub/docs/manual-ocaml/manual029.html (via beaf7a696b33311febebac06574287c27819837c)
- [x] http://www.princeton.edu/~sacm/humor/proof.html (via 69a17c4bfe6fec16a6d9e3f9850f319ef3f4628e)
- [x] http://cocan.org/tips/single_line_comment_syntax redirected to http://www.cocan.org/tips/single_line_comment_syntax (_non existent_)
- [x] http://ocaml.org/tutorials/compiling_with_gnu_make.html
- [x] http://ocaml.org/caml.inria.fr/cgi-bin/hump.en02dc.html?sort=1&browse=55 (_non existent_)
- [x] http://ocaml.org/tutorials/compiling_with_omake.html
[...]

PR #94, opened 2013-01-12
Issue #90

voila:
I've added 4 missing pages in tutorials: hashtbl.html, set.html, map.html, comparison_standard_containers.html.

I've copied them from http://mirror.ocamlcore.org/ocaml-tutorial.org/.

### DIFF
--- a/src/html/tutorials/comparison_of_standard_containers.html
+++ b/src/html/tutorials/comparison_of_standard_containers.html
@@ -1,0 +1,99 @@
+<title>Comparison of Standard Containers</title>
+<body>
+
+<div class="pull-right hero-unit" style="margin-left: 1em" >
+      <em>Table of contents</em>
+      <div ml:replace="toc 2 2"></div>
+    </div>
+
+    
+<h1><span>Comparison of Standard Containers</span></h1>
+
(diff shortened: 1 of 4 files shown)


agarwal:
These are not listed on the frontpage of that website. How did you find
them? Are we perhaps missing more still?

agarwal:
I missed your reference to #90. I see you found these from the broken
links. Maybe we can see the full contents of the mirror directly on the
server by contacting Sylvain.

voila:
Some of the broken links seems to be related to content coming from http://mirror.ocamlcore.org/ocaml-tutorial.org/, I assumed the way to go about this was to add the missing pages to ocaml.org. 
Was that the right thing to do ?

agarwal:
Yes, that's great! There are several more it seems.

voila:
Yes, they are easy to fix. Some others are more tricky (external content which has disappeared), we'll have to agree on how to fix these...

agarwal:
Hi Sylvain,

Any chance we can get a tarball of the ocaml-tutorial mirror? We've noticed
many tutorials are missing from the main list [1]. Some have been tracked
down by following broken links, but maybe we're still missing some. We
could make sure by looking at a complete tarball.

[1] http://mirror.ocamlcore.org/ocaml-tutorial.org/index.html

Thanks.

PR #95, opened 2013-01-14
Issue #90

voila:
The last of the missing tutorials from http://mirror.ocamlcore.org/ocaml-tutorial.org/

### DIFF
--- a/src/html/tutorials/camlp4_3.10/foreach_tutorial.html
+++ b/src/html/tutorials/camlp4_3.10/foreach_tutorial.html
@@ -18,7 +18,7 @@
 
 <h2><a name="Camlp4_3_10_Motivtion">Motivation</a></h2>
 
-<p>One of the greatest strengths of Objective CAML is Camlp4, a modular pre-processor pretty-printer that lets you add syntactic sugar to the language without rewriting the entire grammar from scratch. A syntactic sugar is a language construct that can be decomposed to an existing construct. For example, O'Caml doesn't have a "for-each" syntax, which can be illustrated by the following Python code:</p>
+<p>One of the greatest strengths of Objective CAML is Camlp4, a modular pre-processor pretty-printer that lets you add syntactic sugar to the language without rewriting the entire grammar from scratch. A syntactic sugar is a language construct that can be decomposed to an existing construct. For example, OCaml doesn't have a "for-each" syntax, which can be illustrated by the following Python code:</p>
 
 <pre>
 a_list = ["hello", "world"]
@@ -33,7 +33,7 @@ hello
 world
 </pre>
 
-<p>However, that doesn't mean O'Caml can't support iterators! As it turns out, many modules in O'Caml that define a data structure also provide an "iter" higher-order function that works similarly:</p>
+<p>However, that doesn't mean OCaml can't support iterators! As it turns out, many modules in OCaml that define a data structure also provide an "iter" higher-order function that works similarly:</p>
 
 <pre ml:content="ocaml noeval">
 let a_list = ["hello"; "world"] in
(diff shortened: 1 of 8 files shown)


src/html/tutorials/camlp4_3.10/foreach_tutorial.html
```diff
@@ -0,0 +1,318 @@
...
+</pre>
+
+<p>This code outputs:</p>
+
+<pre>
+hello
+world
+</pre>
+
+<p>However, that doesn't mean O'Caml can't support iterators! As it turns out, many modules in O'Caml that define a data structure also provide an "iter" higher-order function that works similarly:</p>
```
Chris00: The official name of the language is "OCaml".  Please change.

src/html/tutorials/camlp4_3.10/foreach_tutorial.html
```diff
@@ -0,0 +1,318 @@
...
+<pre>
+fun patt1 ->
+  ...
+    fun pattN ->
+      seq
+</pre>
+
+<p class="first_para">which is the desugared form for <code>fun patt1 ... pattN -> seq</code>.</p>
+<p>We can verify that pattern works:</p>
+<pre ml:content="ocaml noeval">
```
Chris00, line 156: Ideally, the toplevel interaction below should be generated through the toplevel itself.  One may behe to add "pa_foreach.ml" to the repository and compile it though...

src/html/tutorials/camlp4_3.10/foreach_tutorial.html
```diff
@@ -0,0 +1,318 @@
...
+         (* ... *)
+      ] ]
+  ;
+END
+</pre>
+
+<p class="first_para">Once the "for" keyword is matched, the parser proceeds to match the next token, but <code>ipatt</code> also matches <code>a_LIDENT</code> (a lowercase identifier is also a legal pattern). At this point, the parser fixes on a rule (whichever comes first) and no longer backtracks to the alternative option.</p>
+<p>The solution is to make the original for loop always match <code>ipatt</code>, but refuse anything except a lowercase identifier.</p>
+<p>We can now verify that everything works as expected.</p>
+<pre  ml:content="ocaml noeval">
```
Chris00, line 278: Same remark as above: ideally, generate outputs at compile time using the toplevel.

Chris00:
Thanks for your efforts.  Some remarks attached to lines of code.

src/html/tutorials/camlp4_3.10/foreach_tutorial.html
voila: Sure, sorry I missed this.

src/html/tutorials/camlp4_3.10/foreach_tutorial.html
voila, line 156: Ok, I'll give this a try

voila:
I have added the pa_foreach1/2/3.ml and compiled them. 
But now the site refuses to build (it seems, because of the presence of .cmo files):

$ make
oasis setup -setup-update dynamic
ocaml setup.ml -build 
Finished, 0 targets (0 cached) in 00:00:00.
SANITIZE: a total of 6 files that should probably not be in your source tree
  has been found. A script shell file
  "/home/manu/Work/ocaml.org.manu/_build/sanitize.sh" is being created. Check
  this script and run it to remove unwanted files or use other options (such
  as defining hygiene exceptions or using the -no-hygiene option).
IMPORTANT: I cannot work with leftover compiled files.
ERROR: Leftover OCaml compilation files:
  File pa_foreach3.cmo in src/html/tutorials/camlp4_3.10 has suffix .cmo
  File pa_foreach2.cmo in src/html/tutorials/camlp4_3.10 has suffix .cmo
  File pa_foreach1.cmo in src/html/tutorials/camlp4_3.10 has suffix .cmo
  File pa_foreach1.cmi in src/html/tutorials/camlp4_3.10 has suffix .cmi
  File pa_foreach2.cmi in src/html/tutorials/camlp4_3.10 has suffix .cmi
  File pa_foreach3.cmi in src/html/tutorials/camlp4_3.10 has suffix .cmi
Exiting due to hygiene violations.
Compilation unsuccessful after building 0 targets (0 cached) in 00:00:00.
E: Failure("Command ''/home/manu/.opam/4.00.1/bin/ocamlbuild' src/lib/ocamlorg.cma src/lib/ocamlorg.cmxa src/lib/ocamlorg.a src/lib/ocamlorg.cmxs src/app/build.byte -tag debug' terminated with error code 1")
make: **\* [build] Error 1

Is there a place out of /src/html/ I should put these files in ?
If so, how do I reference them from the top level ?

agarwal:
I don't find any ml files in your repo. Did you commit this change? Also,
what are these files are for? Is this related to Christophe's comment to
eval code through the toplevel?

Chris00:
I created a branch in my repo to try to fix that.  One can extract the code from the web page (I'll try to avoid duplicating it — I'll likely will update weberizer for that).  As the site grows (and its rendering is slowed down by many toplevel eval), I am thinking to add a cache system — I'll have to think more about dependencies though.

voila:
@agarwal I didn't commit the files, since the site didn't build anymore. Yes, it's related to using the toplevel, for interactive code snippets.

I was wondering, could it become problematic to rely on one version of the toplevel to run these snippets?
How likely would it be, to have snippets of old OCaml code, that does not run anymore with the OCaml version we are using to build the site ?

agarwal:
OCaml is pretty good about maintaining backwards compatibility, but it's a
good general question. If we run into this, we might have to inline the
generated output for older versions and henceforth not evaluate that code
automatically.

I'm not quite following the technical issue on this thread. Christophe,
please merge these commits as you see fit. Thanks. From what I can see, the
current commits in voila/master are okay.

voila:
The issue is that some of the interactive snippets load a .cmo file, and at the moment including these files prevent the site from building. A workaround is to mark the snippet as no-eval, so they are not run by the toplevel and we don't need to include the .cmo files (the current commits)

agarwal:
Thanks for the explanation. I get it now. :)

PR #105, opened 2013-02-03
remove a misleading discussion on single-line comments

iartarisi:
The discussion contains a personal (from whom?) suggestion about OCaml style and is inappropriate in a tutorial aimed at people who see the language for the first time.

### DIFF
--- a/src/html/tutorials/basics.html
+++ b/src/html/tutorials/basics.html
@@ -12,8 +12,7 @@
  *)
 </pre>
 
-<p class="first_para">In other words, the commenting convention is very similar to original C (<code>/* ... */</code>).</p>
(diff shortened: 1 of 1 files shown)


Chris00:
The tutorials were imported as they were and may need to be revised.  I agree that personal comments should be removed from the tutorials.  It would be nice however to add a page with some tricks for editors as indeed some may wonder how to alleviate the absence of single line comments (the comment was not mine in case you wonder).

iartarisi:
Sure, but I think that building a page with editor tricks is a separate issue and cleaning up this section a bit is still a good step on its own. After all, this tutorial looks to be about learning the language, not its development environment.

Chris00:
OK, I have opened a separate issue for that.

iartarisi:
Cool, thanks!

PR #120, opened 2013-03-19
Add some news about CUFP 2013

samoht:
I'm not sure that's the right place to put that news, but I feel that's important to have it on the front page :-)

### DIFF
--- a/src/html/index.html
+++ b/src/html/index.html
@@ -101,6 +101,14 @@
 	</div>
 
 	<div class="hero-unit">
+	  <h4>Commercial Users of Functional Programming 2013</h4>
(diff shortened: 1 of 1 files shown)


agarwal:
Too much whitespace gets added. The general issue is we need a place for
slow-aging news items. For now, we can perhaps make the Download icon
smaller to fit this item in.

Alternatively, we can remove the OUD 2012 hero unit, but we'll just defer
the issue as OUD 2013 will need to be added at some point soon.

samoht:
Right, I have removed the OUD 2012 item. Indeed we should get a proper "long-term" news on the front page (but I guess this is part of the ocaml.org new design process)

PR #126, opened 2013-04-01
install on mageia

fccm:
Add a section about Mageia Linux on the install.html page

### DIFF
--- a/src/html/install.html
+++ b/src/html/install.html
@@ -43,6 +43,7 @@
 	  <li><a href="#macosx" data-toggle="tab">Mac OS X</a></li>
 	  <li><a href="#suse"   data-toggle="tab">SuSE</a></li>
 	  <li><a href="#ubuntu" data-toggle="tab">Ubuntu</a></li>
+	  <li><a href="#mageia" data-toggle="tab">Mageia</a></li>
 	  <li><a href="#windows" data-toggle="tab">Windows</a></li>
 	</ul>
       </li>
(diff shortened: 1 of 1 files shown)


Chris00:
It would be nice to confirm that OPAM is working on that distribution (it should).

fccm:
Chris00 wrote:

> It would be nice to confirm that OPAM is working on that distribution (it should).

I haven't allocated much time yet to play with it (I would play more with it if it could be a solution to manage libs in Cygwin too), so I can only say that the sources do compile out of the box in Mageia 2 and 3.
But can't compile the docs because we don't have pandoc, so I'm just providing the .md files in the rpm:
http://svnweb.mageia.org/packages/cauldron/opam/current/SPECS/opam.spec?view=markup

I've only tested the basic commands found in the tutorials, and everything seems the work as expected.
After installing a lib, how are we supposed to compile with it?
Should we use the full -I ~/.opam/the/path/to/thelib/ in our command line compilation? Or is there some shortcut similar to findlib?
I've also tested creating my own package, but it seems that Opam is not happy with .tar.lzma source tarballs.

If you want more informations you can ask me some commands to test, to check some different features.

Chris00:
On Tue, 16 Apr 2013 07:35:00 -0700, blue-prawn wrote:

> I've only tested the basic commands found in the tutorials, and
> everything seems the work as expected.

Good.  I think it is enough to advertise it on that platform —
especially if no many OCaml packages are available as native packages.

> After installing a lib, how are we supposed to compile with it?
> Should we use the full -I ~/.opam/the/path/to/thelib/ in our command
> line compilation? Or is there some shortcut similar to findlib?

Actually, (almost) all packages in OPAM use findlib.  So you do not
need to (and should not) put the paths to your dependencies in your
build system.

> I've also tested creating my own package, but it seems that Opam is
> not happy with .tar.lzma source tarballs.

I do not think they are supported for now — but you can file a bug
report.  However, it is no big deal to use tar.gz IMHO.

PR #140, opened 2013-04-27
Update functional_programming.html

ShawnOtis:
Grammar and syntax modifications. Note that the word alright was not modified for the word all right, even though the former is not an "actual word", as using all right would not have the same "ring" to it, and thus would not give justice to the sentence. Also, the link at the end is broken and does not lead anywhere: http://ocaml-tutorial.org/_bin/upload_image_form.cmo?name=boxedarray.png


agarwal:
According to the builtin dictionary on my Mac,

> There is no logical reason for insisting that **all right** be two words when other single-word forms such as **altogether** have long been accepted. Nevertheless, although found widely, **alright** remains nonstandard.

I interpret the above to mean it's okay to use **alright**.

Thanks for the fixes.

PR #143, opened 2013-05-05
fixed broken links in IDEAS.org

fccm:
the .org extension was replaced by .html (so breaking these external links), just replace by .org/ seems to work around this.

### DIFF
--- a/IDEAS.org
+++ b/IDEAS.org
@@ -114,7 +114,7 @@
 
 * Blog
   - aggregate related blogs, already provided by [[http://planet.ocamlcore.org/][planet]] but we would
-    like to display healines only on the front page of the site.
+    like to display headlines only on the front page of the site.
   - one new blog where major announcements are made, show this one on
     front page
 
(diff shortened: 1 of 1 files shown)


Chris00:
Looking at the diff, I do not understand what links you fixed.

fccm:
Yes sorry I should have fixed only the broken links, and no formating at all at the same time that makes it difficult to compare.

go on this page:
https://github.com/ocaml/ocaml.org/blob/master/IDEAS.org
https://github.com/ocaml/ocaml.org/blob/c1ab30fe61d4d6aab4b1220018861776af1f4ad3/IDEAS.org

go at the paragraph "Design" at the end of this page,
then try the links that are there:
- examples of sites for inspiration: [PostgreSQL](http://www.postgresql.html/), [mongoDB](http://www.mongodb.html/), [Drupal](http://drupal.html/), [Perl](http://www.perl.html/)

Chris00:
Got it.  This is a GitHub bug — the links were fine when using Emacs [org mode](http://orgmode.org/) which I why I did not figure it out initially.

PR #144, opened 2013-05-19
Improve a taste of ocaml

lefessan:
The section "A Taste of OCaml" gives only one complex example. This
replacement proposes a pull-down menu with 4 examples (Hello World,
Binary Trees, Fibonacci, List/Array ops), Hello World being the default.
The framework is easy to improve with more examples.

### DIFF
--- a/src/html/index.html
+++ b/src/html/index.html
@@ -21,7 +21,7 @@
 	  benefits from one of the most advanced type systems and
 	  supports functional,
 	    imperative and object-oriented styles of programming.
-      
+
 	    <a href="description.html"
 	       class="btn btn-large pull-right"
 	       >Read more...</a></p>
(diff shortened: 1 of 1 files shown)


agarwal:
Thanks. I like these additional examples. Few comments:
- There is already the "[a stronger taste](http://ocaml.org/taste.html)" link, which has additional examples. What example should be within the drop down list vs on taste.html?
- The formatting doesn't work in all cases. If your window is small enough, the code box overlaps the sentence below.
- Instead of a drop down list, we could have a new example randomly shown every time the page is refreshed. We do something similar on the [Companies](http://ocaml.org/companies.html) page.

lefessan:
Hi !

1/ I think all short examples (less than 10 lines with comments) should appear on both pages. The drop down list gives a faster access than clicking on a link.

2/ I set the window height in the code, it can be removed. It should probably done differently, but (a) I thought a better expert of CSS would do it and/or (b) I thought that it would only appear before the new website is up, so it wasn't worth the trouble.

3/ My primary target is the newbie, coming for the first time. I think we should not let the choice be done randomly, but choose what he is expecting, i.e. the hello world example, leaving him the possibility to see more thanks to either a drop down list or a "prev/next" button set.

pw374:
Does this still make sense? If not, perhaps this PR should be closed...

Chris00:
Agreed.  One can always reopen it if needed.

PR #147, opened 2013-05-22
added gitorious.org

fccm:
Added gitorious.org link after the links to Github and Bitbucket in the "use" menu.

### DIFF
--- a/src/html/menu.html
+++ b/src/html/menu.html
@@ -46,6 +46,8 @@
 	   >GitHub</a></li>
     <li><a href="https://bitbucket.org/repo/all?name=ocaml"
 	   >Bitbucket</a></li>
+    <li><a href="http://gitorious.org/search?page=1&q=ocaml"
+	   >Gitorious</a></li>
   </ul>
 </li>
 <li class="dropdown">


agarwal:
We need to also update the french and italian versions of the menu. Or, my preference, get rid of them.

fccm:
We could also make languages independent versions.
This solution would avoid the difficulty to get all translations synchronised.
This is how it works on Wikipedia, Wikibooks, Wiktionary, etc.
We could also invite other language communities to contribute to ocaml.org,
like for example the Japaneese camelids.

agarwal:
> We could also make languages independent versions.

How do you do that?

> We could also invite other language communities to contribute to ocaml.org

Yes, but we still have to provide a framework supporting this, which is not completely trivial. Some features we'd want are:
- an "include" mechanism to e.g. reuse code snippets (you don't want to copy/paste identical text)
- you want to easily know if the non-English pages are out of sync with the English pages
- and there have to be enough contributors interested in supporting the other languages

Chris00:
On Sat, 25 May 2013 06:11:58 -0700, Ashish Agarwal wrote:

> Yes, but we still have to provide a framework supporting this, 
> which is not completely trivial. Some features we'd want are:
> - an "include" mechanism to e.g. reuse code snippets (you don't 
>   want to copy
>   /paste identical text)

This is already possible but not used because the pages were copied 
from the wiki which had the code duplicated.  Note that the comments 
in the code are also usually translated, so an include will not be 
completely satisfactory.

PR #154, opened 2013-06-24
Replace Jane St logos

amirmc:
Note: I can't build the site to check that these work/look ok (I'm opening a separate issue about that).  Since I'm replacing files, I'm confident they'll be fine so I'll merge this in within the hour.  If someone happens to see this beforehand and _can_ run a build, please do comment here either way.


agarwal:
The support.html page looks fine with the new `janestreet.png` logo.

`jane-street-blue.jpg` is no longer the right filename since the new logo is not blue. Also, I don't think we need it. It is used only on the success.html page, but maybe we can simply use `janestreet.png` there too. In either case, sizing needs to be set. It looks too big right now.

amirmc:
I just fixed my problems with `make` and noticed how big it was.  I was reluctant to change the filename as I wasn't sure how many other places the same file was referenced and replacing the file seemed like the quickest way to make sure the change was global.

agarwal:
I used this command to search for links:

```
$ find src/html -exec grep -H "jane-street-blue.jpg" {} \; 2> /dev/null
```

amirmc:
Thanks!  So it's only in use on the success page.  Ill rename the file and then merge this for now.  The png is one I had to produce from the jpg I have. It doesn't look that great as I made the background transparent by hand (again, wasn't sure where the logo was in use, but I see that the background was white anyway).  

I'll ask Jane Street to send me a proper png file and I'll switch out the current files then.

Chris00:
@amirmc If you use Emacs, `M-x grep-find` or equivalently `M-x find-grep` presents you a command that you just have to complete.  Then you can simply click on the file names in the result to go to the occurrences — this makes this can of work really fast.

PR #165, opened 2013-07-18
Clarify module usage from the OCamL command line.

ghost:
It took me a while to figure out why the commands were not working from the prompt so I felt it prudent to elaborate in the tutorial, just in case people try this without compiling.


Chris00:
I ported slightly modified versions of the changes I found were still pertaining (e9e794e, 9f5650c).

PR #170, opened 2013-08-02
replaced the old fashioned favicon from the old website for a more modern favicon

fccm:
I'm not the author of this favicon,
it comes from the link on opam.ocamlpro.com to ocaml.org
http://opam.ocamlpro.com/ext/img/ocaml.png
(I've edited it a little in Gimp though to optimise it a little
for a very small favicon size.)


Chris00:
What is the license?

amirmc:
This is the OCaml icon that I've worked on as part of the redesign work through OCaml Labs. 

There's probably some more tweaking to do but using this one is fine for now. I'm already using it on the OCaml GitHub account. The final icon will be released under an open licence. 

Best wishes,
Amir

## 

sent via mobile

Issue #203, opened 2013-11-06
Platform doesn't exist yet, so how to handle the Platform page


agarwal:
The [Platform landing page](http://preview.ocaml.org/platform/) has lots of dummy text, by necessity since the Platform isn't actually released yet. How should we handle this page in the mean time? Perhaps add "Coming Soon" banner at the top, and direct people to go to "Install" for now.

amirmc:
We shouldn't link to this page from the front page or navigation yet. We should keep it but link to it from elsewhere. 
That makes it less prominent but we can still send people to it. I'll submit a PR for consideration tonight.

PR #207, opened 2013-11-06
Changes to how Platform page is linked

amirmc:
This is one way to deal with #203

@pw374, please can you check the nav bar change I've made is sufficient (i.e doesn't break anything else).  If it's ok, then I'm happy for either you or @agarwal to merge this.

### DIFF
--- a/src/site/index.md
+++ b/src/site/index.md
@@ -35,8 +35,8 @@
                 <section class="span4 home-feature">
                     <a href="/platform">
                         <img src="/img/platform-large.png" alt="Platform">                    </a>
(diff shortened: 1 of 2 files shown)


pw374:
I checked: it's fine.
But I have to ask about `site/packages/core.html`: should it be removed?

amirmc:
Ah, yes.  the whole packages folder can be removed from this repo. In fact, cleaning up any unused files is a good thing, include those in the root.  If we're not ready to delete them, then we should at least move them to a folder out of the way ('archived'?)

PR #208, opened 2013-11-06
fix broken links

amirmc:


amirmc:
Ok, so this clashes with @pw374 recent commits so I'll leave it here for him to see the diff.  Close it whenever you're ready.

PR #274, opened 2013-11-14
Improve body text contrast.

avsm:
The current choice of slightly gray text on white is very hard to read on some
devices.  This fixes the colors to be black on white, to make large bodies of
text and the tutorials easier on the eyes.

(I made this same change to realworldocaml.org after several requests from
reviewers)

### DIFF
--- a/site/css/bootstrap.css
+++ b/site/css/bootstrap.css
@@ -216,7 +216,7 @@ body {
   font-family: Domine, serif;
   font-size: 14px;
   line-height: 24px;
-  color: #404040;
+  color: #000000;
   background-color: #ffffff;
 }
 a {
@@ -830,7 +830,7 @@ pre {
   padding: 0 3px 2px;
   font-family: Monaco, Menlo, Consolas, "Courier New", monospace;
   font-size: 12px;
-  color: #404040;
+  color: #000000;
   -webkit-border-radius: 3px;
   -moz-border-radius: 3px;
   border-radius: 3px;
@@ -889,7 +889,7 @@ legend {
   margin-bottom: 24px;
   font-size: 21px;
   line-height: 48px;
-  color: #404040;
+  color: #202020;
   border: 0;
   border-bottom: 1px solid #e5e5e5;
 }


Chris00:
Merged manually.

dsheets:
I don't know how many edits we've made to bootstrap.css (could check history...) but perhaps we should consider either 1.) editing the bootstrap LESS/CASS/whatever sources or 2.) using the CSS cascade to change properties. If we modify the bootstrap CSS, we make upgrade and maintenance harder.

Just my 2p...

avsm:
Yep; I was just following the crowd here.

pw374:
`<h1>` is now black, but  `<h1><a>` isn't. Is this intentional?

Chris00:
On what page?  The titles "Learn",... on the front page are `<h1><a>` and are still black.

Chris00:
@dsheets I agree.  Somebody should put the original LESS files into the repo and write the `make` rules to generate the CSS.

agarwal:
@dsheets I had the same concern and we had some discussion about this previously over email. @pw374 did feel he could do back-port his changes into pure LESS. If I understood correctly @amirmc felt this wasn't a priority because future design work could possibly involve a full change, like switching even out of Bootstrap.

However, as we can see, we are continuing to work on the design even now. Thus, I think we should reconsider this.

amirmc:
In general, I'm keen to push changes 'upstream' as much as practicable but I don't see the value in this specific case. Please allow me to explain.

### Upgradability

Yes, making edits directly to CSS means upgrading is more complicated but let's take a step back and think what upgrading actually means.  We're using Bootstrap 2.3.x and the next one up is Bootstrap 3.x.x.  V3 has lots of backward breaking changes so would require [a lot of work to upgrade](http://bootply.com/bootstrap-3-migration-guide) anyway.  In that sense, we're not saving ourselves any work by using LESS directly.  In any case, if we're not looking for specific features or bug-fixes from the new version then I see no reason to upgrade any time soon.  When we eventually do decide to upgrade (either to Bootstrap 3.x.x or even [Zurb Foundation](http://foundation.zurb.com)) we won't have saved ourselves any future work by _back-porting_ changes now.

Before anyone asks, we didn't use Bootstrap 3 for the new site as it was still under active development during the design phase.

### Maintainablity

The CSS file is large so changes made to it are harder to track.  For this reason I'm much more keen on keeping changes in a separate file and make use of the cascade (as David mentioned).  This was originally what was meant to happen but for whatever reason we're here now.  I'd suggest any new CSS changes are added to [bootstrap_mod.css](https://github.com/ocaml/ocaml.org/blob/redesign/site/css/bootstrap_mod.css) or [ocamlorg.css](https://github.com/ocaml/ocaml.org/blob/redesign/site/css/ocamlorg.css) going forward, so we avoid further edits to bootstrap.css. At least this way, we will know what defaults we're overriding (w.r.t to the LESS/CSS the designers gave us, not the original bootstrap).

### LESS compilation

So far, no-one has suggested how we practically get from LESS to CSS, despite it cropping up numerous times.  I have no experience with LESS but the options I find tend to involve using [Node or Ruby](http://lesscss.org), and I don't see the point in adding either of these to our toolchain just for the sake of generating CSS.  Perhaps we could run `less.js` on the server-side but that doesn't seem like a good idea either. Perhaps I'm misunderstanding this compilation step entirely but no-one's corrected me yet.

As I mentioned to Ashish in an email thread, the existing ocaml.org site doesn't make use of LESS files directly and it's not been in any way detrimental to us.  I've also had a quick look at opam2web and I don't see any LESS files in there either.  It doesn't seem as though the current situation of writing CSS is a regression in any way.

So in short, no-one appears to be experienced at working with LESS files which means I find it confusing when people suggest pushing changes back to that level.  If I'm wrong about this, _please do correct me_ and point me to a solution that's either being used or you're happy to start using.

### Summary

I'm normally happy to push changes 'upstream' and I've no problem adding the LESS files to the repo but given all the things I mention above, I see little value to the time spent back-porting our current changes.  There are (and likely always will be) higher priority problems.

Re-reading the above, I hope it doesn't come across as aggressive.  I'm trying to lay out my thoughts and explain my viewpoint clearly.  As always, I'm happy to hear counterpoints and corrections.

Best,
Amir

agarwal:
Your comments are perfectly reasonable. It's true that I haven't actually used LESS, and I'm just assuming that it can't possibly be too hard to compile it. But I'll refrain from commenting further till I can actually look into it. Even if we resolve this though, it seems back-porting the changes may not be worth it.

pw374:
> On what page?  The titles "Learn",... on the front page are `<h1><a>` and are still black.

On my computer, with Safari and Firefox, the `<h1><a>` titles (the 4 titles right to the 4 rounded-squares boxes, and "News") are dark grey, while "A taste of OCaml" and "Packages" (the one left to "A taste of OCaml") are black.

PR #319, opened 2013-11-21
Correct formulation

lehy:
"gives you to access" -> "gives you access to"


amirmc:
Thank you! It should be live within minutes.

PR #323, opened 2013-11-21
Update Caml Consortium

lefessan:
- Add Inria logo on the Caml Consortium page
- OCaml Consortium is not correct, the consortium is called Caml Consortium
- Membership is always 3000 Euro, and cannot be higher, as erroneously stated


amirmc:
Thanks for the PR!  I'd argue that the Consortium should change it's name.  There's already been too much confusion in the past regarding the name of the language (I still find references to `O'Caml`).  In fact the old page says "The **Caml** Consortium federates the design and development of the **Caml** language" (emphasis added), so the intent is clear.

Who should I suggest this change to?

avsm:
That change is a lot more involved since it's a legal entity, so I'd suggest merging this PR to reflect the current status quo.  I'll be at the next Consortium meeting (on Monday, in fact) and will raise this query there.

-anil

amirmc:
Sounds good.  Please do raise this next week.  Merging this now.

lefessan:
Thanks!

agarwal:
Is it actually important to have the name on this page be the legal name? This page is not a legal document. I'd argue that it's alright to say _OCaml_ Consortium pretty much everywhere, and this would avoid unnecessary confusion for most people.

PR #325, opened 2013-11-21
add link to list of packages (fix #317)

### LINKED ISSUE #317: Recent opam packages updates
Can be found at the bottom of http://ocaml.org page

1) Maybe, it will be great to have link to longer list of recent updates
2) If we will click in tab header `Recent updates`  we will be redirected to the top of the page using anchor. Now, if we will scroll down we will see that tab header is visible only when mouse is hovering it. If we move mouse out --- it is invisible.

pw374:


Chris00:
To make the link more discoverable, a line of text repeating the link should be put under the package table.

PR #359, opened 2013-12-15
Add Homebrew installation section in install.md

jdavis:
I was looking to install OCaml and couldn't find a section for installing it using Homebrew so I decided to write it because the `ocaml` package (version 4.01.0) already existed in the Homebrew formulae: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/objective-caml.rb

Let me know if any of the wording, organization, etc. should be changed! :smile:


Chris00:
Closed via 852a85d

Issue #326, opened 2013-11-21
list of recent package updates


pw374:
Just to let you know that it's normal that the list is currently irrelevant, because there's been a huge git-renaming in opam-repository.

So it's going to be either alphabetically ordered or anti-alphabetically ordered, until there's been enough actual updates.

samoht:
We have the same issue on http://opam.ocaml.org

I guess the right solution is to get the first appearance of the relevant `opam` file in the log. @AltGr any thoutghs ? 

(The relevant function in opam2web is https://github.com/ocaml/opam2web/blob/master/src/o2wUniverse.ml#L69)

AltGr:
On it already ; I am looking for the best way to get the most relevant information from git

pw374:
> We have the same issue on http://opam.ocaml.org

I noticed that ocaml.org and opam.ocaml.org had the same weird list (Z...Y...) and then I noticed that ocaml.org was using wrong version numbers... Then I realised it was because the structure of opam-repository was changed...

> I guess the right solution is to get the first appearance of the relevant opam file in the log. @AltGr any thoutghs ?

When you `git mv` a directory, it's actually all the files in that directory that are `git mv`ed, so I don't know what you mean by looking for the first appearance of some `opam` file since it has been moved as well...

However I suppose there are git-related magic tricks ("magic" here means everything that's not already understood) in opam2web... So on my side, I'll just wait until opam2web fixes this issue (or until a decision has been made not to fix that issue), and then I'll try to use opam2web to generate that table on ocaml.org. (Indeed, currently it's a small shell-script that generates that table for ocaml.org, not opam2web.)

samoht:
FWI, this is in now fixed in opam2web by https://github.com/ocaml/opam2web/commit/25b2c2c171adadf2f772a6e8e8b38321b121e523

dsheets:
Where is the code that generates `opam-update-list` referenced from https://github.com/ocaml/ocaml.org/blob/master/template/front_package.mpp? I'd like to update it because all the links are broken. Ideally, this should be incorporating the same HTML fragment used on the front page of https://opam.ocaml.org/ so these bugs can't happen.

pw374:
(The links have just been fixed.)

The code that generates that file is on the machine that generates ocaml.org.
It should be replaced by what's generating the table for opam.ocaml.org but I've hadn't had the time to do it yet. :-/

PR #369, opened 2013-12-25
Make front page package list title reflect reality

dsheets:
Also fix HTML error


dsheets:
The links are also broken but I think that should be tracked by #326.

PR #377, opened 2013-12-27
Pre- and in-order for binary trees

phimuemue:
Implemented solutions for pre- and in-order-traversal and reconstruction for binary trees. Probably does not work if there are duplicate elements in the sequences.


Chris00:
I merged it as 719cf2f.  I you have so time, there is still the reimplementation using difference lists! ;-)

PR #380, opened 2013-12-30
updated 99problems: slice solution optimized -> tail recuisive and use one additional function

zma:
updated slice solution in 99problems:

make it tail recuisive and use one additional function

### DIFF
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -387,13 +387,12 @@ SOLUTION
 
 > ```ocamltop
 > let slice list i k =
->   let rec split acc i = function
->     | [] -> (acc, []) 
->     | hd::tl as l -> if i = 0 then (acc, l)
->       else split (hd::acc) (i - 1) tl in
->   let (_, r) = split [] i list in
->   let (l, _) = split [] (k - i + 1) r in
->   List.rev l
+>   let rec aux acc i k cur = function
+>     | [] -> acc
+>     | h :: t -> if cur < i then aux acc i k (cur+1) t
+>         else if cur <= k then aux (h::acc) i k (cur+1) t
+>         else acc in
+>   List.rev (aux [] i k 0 list)
 > ```
 
 ```ocamltop
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -387,15 +387,13 @@ SOLUTION
 
 > ```ocamltop
 > let slice list i k =
->   let rec take n = function
->     | [] -> []
->     | h :: t -> if n = 0 then [] else h :: take (n-1) t
->   in
->   let rec drop n = function
->     | [] -> []
->     | h :: t as l -> if n = 0 then l else drop (n-1) t
->   in
->   take (k - i + 1) (drop i list);;
+>   let rec split acc i = function
+>     | [] -> (acc, []) 
+>     | hd::tl as l -> if i = 0 then (acc, l)
+>       else split (hd::acc) (i - 1) tl in
+>   let (_, r) = split [] i list in
+>   let (l, _) = split [] (k - i + 1) r in
+>   List.rev l
 > ```
 
 ```ocamltop
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -384,15 +384,17 @@ elements between the `i`'th and `k`'th element of the original list
 way the `List` module numbers elements).
 
 SOLUTION
-
 > ```ocamltop
 > let slice list i k =
->   let rec aux acc i k cur = function
->     | [] -> acc
->     | h :: t -> if cur < i then aux acc i k (cur+1) t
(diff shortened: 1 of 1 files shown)


Chris00:
Your solution also reconstructs the beginning of the list when we mean to discard it.  It is also less legible IMHO.  If you are concerned about the lack of tail recursion, I would just challenge the reader to write such a version for `take` (possibly with a sub-hidden solution).

zma:
@Chris00 Your point do make sense and I agree reconstructing the list is bad. My main concern was that `take` and `drop` share some common part and it will require an additional function to make `take` tail recursive.

I updated the solution in https://github.com/zma/ocaml.org/commit/2066907eaa4346a385f1e7c67e186e22e0074260 . Could you please take a look and give me some comments on this one?

Chris00:
Thanks for your update.  However, I'd like to keep the original version — because I think it is the natural solution readers would come with — and **add** a second solution **motivated** by better factorization and by being tail recursive (explaining in a couple of words why that matters). I added a couple of remarks directly on the code.

site/learn/tutorials/99problems.md
```diff
@@ -387,15 +387,12 @@ SOLUTION
...
->   let rec take n = function
->     | [] -> []
->     | h :: t -> if n = 0 then [] else h :: take (n-1) t
->   in
->   let rec drop n = function
->     | [] -> []
->     | h :: t as l -> if n = 0 then l else drop (n-1) t
->   in
->   take (k - i + 1) (drop i list);;
+>   let rec aux acc i k cur = function
```
Chris00: Please try to name local functions with something better than `aux`. I know that finding a good name is not always easy but it helps the reader.

site/learn/tutorials/99problems.md
```diff
@@ -387,15 +387,12 @@ SOLUTION
...
+>     | [] -> acc
```
Chris00: I think `List.rev` should be here.

site/learn/tutorials/99problems.md
```diff
@@ -387,15 +387,12 @@ SOLUTION
...
+>     | h :: t -> if cur < i then aux acc i k (cur+1) t
+>         else if cur <= k then aux (h::acc) i k (cur+1) t
+>         else acc in
```
Chris00: You should align `else` with `if`.

site/learn/tutorials/99problems.md
```diff
@@ -387,15 +387,12 @@ SOLUTION
...
->     | h :: t -> if n = 0 then [] else h :: take (n-1) t
->   in
->   let rec drop n = function
->     | [] -> []
->     | h :: t as l -> if n = 0 then l else drop (n-1) t
->   in
->   take (k - i + 1) (drop i list);;
+>   let rec aux acc i k cur = function
+>     | [] -> acc
+>     | h :: t -> if cur < i then aux acc i k (cur+1) t
```
Chris00: Now you do not accumulate the elements to drop them but you perform this test even when you know it will be false (after dropping the elements).  I wonder whether it is not better to write a function to abstract the common "skeleton" to both functions, passing functions as arguments to take care about their differences.

site/learn/tutorials/99problems.md
zma: This is related the non-clear `aux` function name. The `aux` here returns a reversed list, but it is immediately clear from the name. A better function name should be used as you suggested.

site/learn/tutorials/99problems.md
zma: Yes. Thanks for the comment on the format.

zma:
A new version that abstracts the `take` and `drop` by `partial_fold`:

```
let slice list i k =
  let rec partial_fold f acc n = function
    | [] -> (acc, [])
    | h :: t as l->
      if n = 0 then (acc, l) else partial_fold f (f acc h) (n-1) t in
  let (_, t) = partial_fold (fun _ _ -> []) [] i list in
  let (acc, _) = partial_fold (fun acc h -> h::acc) [] (k-i+1) t in
  List.rev acc
;;
```

@Chris00 This one also discards the accumulator after the first call to `partial_fold`. However, the accumulator is not constructed here. Another drawback of this implementation is that it seems more complex and harder to understand compared to the original one with `take` and `drop`. I am afraid this one may not be an ideal solution for the `slice` problem here, either. I think we may keep the original as the solution to the problem.

site/learn/tutorials/99problems.md
zma: Yes, that's true that there are lots comparing operations that are not needed here.

After reconsidering the `take` and `drop`, I guess they can be abstracted as an operation that "fold"s the first n elements as a list and produces a record of the accumulator and the remaining tail of the original list. I implemented on at https://github.com/ocaml/ocaml.org/pull/380#issuecomment-31457782 .

Chris00:
I have _added_ the alternative solution you propose with some explanation.  Hope it would have answered your questions if you were to stumble on that solution anew.

zma:
That is good for me. Thanks!

PR #386, opened 2014-01-12
Added an alternate solution to the replicate

brad-h:


Chris00:
What are the benefits of this alternate solution?

brad-h:
Sorry, I should have added more context. The benefits are a single inner function, List.rev is not needed, and reuse of the upvalue. The disadvantages are a larger pattern, less separation of concerns (vs the prepend of the original), and using the upvalue in that way may be considered bad style.

The alternate solution uses a single inner helper that matches on a pair. The pair counts down as it traverses the list. When the countdown is 1, the fst of the pair is the upvalue, n, and the snd is the tail.

I don't think I could make a strong argument for or against either solution, so I decided to present it as an alternate.

brad-h:
Also, the exact same approach could be applied to the next problem. Got trigger happy :)

Chris00:
On Sun, 12 Jan 2014 13:01:50 -0800, brad-h wrote:

> single inner function,

As you say, that has the disadvantage of making the code less clear.  Your function is switching between two behaviors which makes it harder to understand.

> List.rev is not needed, 

List.rev would not be needed in the original version either if we did not want `aux` to be tail recursive.  Your solution is not tail-recursive and so may blow up the stack on long lists.

> and reuse of the upvalue. 

Not sure what you mean.

brad-h:
Good call; I hadn't even noticed that. The original is a better solution. Thanks!

PR #388, opened 2014-01-21
href="about.html" -> href="/about.html"

klakplok:
The "about this site" link was relative, leading to 404 when clicked from a subdir (ex: go to docs and then click about).


pw374:
Thanks :)

PR #402, opened 2014-02-14
Update functional_programming.md

raphael-proust:
Spacing in inline infixing of multiplication operator: `( * )`


pw374:
thanks!

PR #403, opened 2014-02-14
bump modules up

speg:
While going through the tutorials (in order) the maps tutorial was rather confusing because it started off with the `module` keyword but modules had not yet been introduced and I was quickly confused.


agarwal:
Thanks!

dsheets:
http://ocaml.org/learn/ still has the old order.

amirmc:
Well spotted. That list is still hard coded so I add the relevant line.

PR #406, opened 2014-02-26
Improved French pages

lefessan:
- On mailing_lists.fr.html, fix the "en" link that didn't work on Chrome
- On support.fr.html, fix the table of consortium members, and link directly
  to mailing_lists.fr.html.


pw374:
thanks

PR #409, opened 2014-02-26
Languages: use flags images instead of letters when available

lefessan:


lefessan:
Since I was not completely sure about the correspondance between flags and other languages, I just put a few flags, the other ones should be easy to add by the corresponding people. The script checks whether the flag is available, and if not, just use the former printing.

Chris00:
Two questions:
- Aren't flags visually distracting (too many colors) ?
- Should we use the UK, US (or Australian,...) flag for English?

dbuenzli:
I also find that using flags disrupt the design and that national flags are not such a good fit for conveying the notion of language. Using ISO-639-1 like two letter codes also has the advantage of making the list sortable and thus easier to scan --- well for visitors that are used to latin scripts...

lefessan:
What I have seen on most sites is:
- A flag indicates the current language (UK for English)
- When you click on that flag, a list of other languages is displayed, where you can select your language.

However, such a system is only worth when your website is already available in many languages, which is not the current case for ocaml.org. For the current state, I think we should not worry too much about displaying a few flags in a corner. If it becomes a problem one day, it means there are enough contributors and languages to be able to switch to another design.

About the flags being visually distracting, it might actually be a good think if we want to attract people to translate pages. I don't think it would bother people trying to learn OCaml...

script/translations.ml
```diff
@@ -23,8 +23,14 @@ let translations path =
   let t = List.sort (fun (l1,_) (l2,_) -> String.compare l1 l2) !t in
   (* Convert to HTML *)
   let to_html (l, fn) =
-    if l = l0 then sprintf "<li class=\"active\">%s</li>\n" (lang_to_html l)
-    else sprintf "<li><a href=\"%s\">%s</a></li>\n" fn (lang_to_html l) in
+    let html =
+      let l = if l = "" then "en" (* default language *) else l in
+      if Sys.file_exists (Printf.sprintf "site/img/flag_%s.png" l) then
```
avsm, line 28: this needs an `alt` tag referring to the ISO code

lefessan:
I added the alt tag.

amirmc:
I'm really not keen on adding flags to represent language.  I'd find them distracting as graphics and somewhat disrespectful/imperious regarding language (i.e there are many parts of the world that speak English, Portuguese and Spanish, that are _not_ England, Portugal or Spain).

I also disagree that adding flags would encourage people to translate pages.  The list of languages in the top-right is enough to indicate that we have translations and if the aim is to encourage more, I don't think adding flags is the way to do it.  We should use another mechanism to achieve that.  From what I can see, the current system of two letter codes appears to be working well [1], so I'm inclined to keep it.

[1] http://ocaml.org/learn/tutorials/basics.html

lefessan:
> I'm really not keen on adding flags to represent language. I'd find them distracting as graphics and
> somewhat disrespectful/imperious regarding language (i.e there are many parts of the world that
> speak English, Portuguese and Spanish, that are not England, Portugal or Spain).

You mean that you are doubtful that English was invented in the UK, French in France and Portuguese in Portugal ? Maybe also saying that OCaml was invented in France is disrespectful for other people using OCaml all other the world ?

> I also disagree that adding flags would encourage people to translate pages. The list of
> languages in the top-right is enough to indicate that we have translations and if the aim is to
> encourage more, I don't think adding flags is the way to do it. We should use another
> mechanism to achieve that. From what I can see, the current system of two letter codes appears
> to be working well [1], so I'm inclined to keep it.

At least, for myself, I don't see the point of translating this website to French if there is no clear/obvious/visual way to know that there is a French translation. I don't know any other website using the current system, and actually, I know very few people who noticed these two letters on the current website and knew it was possible to have translations.

dsheets:
The site could respect accept-language header and set/read a user preference cookie to automatically select the appropriate translation for a user. If we want these features, someone should probably make a new ticket.

With these features, if non-anglophonic users visit the site and have their browsers configured to request their preferred language (maybe they don't because they don't like leaking their language/nationality), we would immediately serve the correct translation and then they would know about it because they don't even have to select a flag/ISO code.

What do you think?

avsm:
On 27 Feb 2014, at 10:04, Fabrice Le Fessant notifications@github.com wrote:

> I don't know any other website using the current system, and actually, I know very few people who noticed these two letters on the current website and knew it was possible to have translations.

I think Amir's point is that we don't distinguish between American english and British english, so which flag do you choose?  We're mostly using American spelling, so it ought to be the US flag up there.  The ISO codes are more specific.  I have no strong feelings for or against the flags, but if they are used, I'd rather see them used accurately and completely (i.e. by grabbing one of the ISO<->flag icon sets freely available online).

> Maybe also saying that OCaml was invented in France is disrespectful for other people using OCaml all other the world ?

This is a strawman and irrelevant to the current discussion.

To gather more data, I took a look at other language websites to see how they deal with this:

Python: no documentation translations at all, but there's this language information wiki: https://wiki.python.org/moin/Languages (no flags or ISO codes)

Ruby: https://www.ruby-lang.org/en/ has no flags or ISO codes, but instead has "this site in other languages" with it written in the native tongue of that language (which makes more sense than either the flag or the ISO code to me).

Haskell: no language translations that I could find.

PHP: http://www.php.net/docs.php the documentation languages are written out in English (again, no flags or ISO codes).

So this leads me to wonder if either flags or ISO codes are the right thing to do.  HTTP permits the use of 'accept-language' for the client to signify which source language it wants.  We could simply add that feature to the webserver configuration and redirect the client to the appropriate sub-tree of the website if it's available.  Since we're running OCaml.org in a VM now, this should be straightforward.

-anil

avsm:
On 27 Feb 2014, at 10:14, David Sheets notifications@github.com wrote:

> The site could respect accept-language header and set/read a user preference cookie to automatically select the appropriate translation for a user. If we want these features, someone should probably make a new ticket.
> 
> I agree with this, given that it's the recommended mechanism of serving multilingual content in RFC2616 (and clarified in RFC2295 and 2296).

-anil

lefessan:
I agree that the website should conform to "accept-language" headers. However, most websites also provide a button/selector/list to warn the user that there are other translations available. It is especially important when, for example, you are using a computer in another country, that is configured for the local language, and you need a way, without hacking the browser configuration (in the local language), to change the language displayed on the webpage.

avsm:
On 27 Feb 2014, at 10:38, Fabrice Le Fessant notifications@github.com wrote:

> I agree that the website should conform to "accept-language" headers. However, most websites also provide a button/selector/list to warn the user that there are other translations available. It is especially important when, for example, you are using a computer in another country, that is configured for the local language, and you need a way, without hacking the browser configuration (in the local language), to change the language displayed on the webpage
> 
> Yes, the selector box is still necessary (it drives me insane when I'm in a foreign country and Google forces me in the local language).  The prevailing winds in the other websites I looked at seemed to simply list the language translations in the footer, though.

-anil

lefessan:
@avsm: listing translations in the footer is good when you have "full translations" of the website. When only a few pages are translated, it is probably better to have per-page links, as currently done. My point is only to make these links more visible by using flags.

Once we have full translations, we should indeed move to a different navigation system (either different websites per language, or per-language subtrees).

By the way, I used the icons from /usr/share/locale/l10n/ on my Debian computer, which was a way, for me, to respect some kind of "standard"...

dbuenzli:
Please be very careful on how you interpret the `accept-language` header, otherwise we will end up with the horror that the debian.org site is: it redirects you to your first `accept-language` on every link you follow even though you were on a page in another language which is very annoying if e.g. you want to browse the site in english but your os is in french (funnily [their instructions](http://www.debian.org/intro/cn.en.html) to get the site in another language is to change your _browser_ preferences, right... I'm going to change my browser language to be able to browse your site in english)

Here's what I have implemented on erratique.ch and I believe is the right way of doing it, I do it cookieless but I think that having a cookie that may remember what was the language on a previous visit may be a good thing for the landing page (see below). 
1. There are always separate URIs for translated webpages, i.e. each translated webpage is addressable.  If the user comes by following a link to an URI, show that URI regardless of its cookie preference. He can still switch to the translation by himself if he wishes so.
2. On each page of the website, the links always lead to pages of the same languages, except for the links to the translation of the page of course. 
3. When a visitor gets on the landing page ocaml.org (and has no cookie set). Redirect the visitor to a translated landing page according to its accept-language (or to its cookie preference). This is the **single time** (with the error pages) that the system automatically redirects the visitor without having him make the decision. Here are the relevant `.htaccess` lines for my site:
   
   ```
   RewriteCond %{HTTP:Accept-Language} ^fr [NC]
   RewriteRule ^$ /contact [L,R=permanent]
   RewriteRule ^$ /contact.en [L,R=permanent]
   ErrorDocument 404 /e404
   ErrorDocument 403 /e403
   RewriteCond %{HTTP:Accept-Language} ^fr [NC]
   RewriteRule ^e404$ e404.fr [L]
   RewriteRule ^e404$ e404.en [L]
   RewriteCond %{HTTP:Accept-Language} ^fr [NC]
   RewriteRule ^e403$ e403.fr [L]
   RewriteRule ^e403$ e403.en [L]
   ```

Regarding the discussion with flags I will just say that what I think is the largest multilingual site in the world, wikipedia, doesn't use them. 

Usually flags are used for _commercial_ websites that want to sell you something as in that case your country of origin becomes relevant, i.e. they are used for getting localized information rather than having the site in your language. 

Living a country with four official languages we have got quite a few translated websites here and usually the language switch are either two letter codes or the language written in plain text as wikipedia does (see e.g. http://www.admin.ch). 

Making the language switch very proheminent is absurd for something you will do only once during your browsing interaction with the site and usually people that really need them know how to scan for them.

dbuenzli:
Also don't forget to put alternate hreflang `link` tags in the header of all the pages to advertise the translations to machines. See e.g. [google](https://support.google.com/webmasters/answer/189077?hl=en)'s instructions.

dsheets:
For further discussion of semi-automatic translation selection, please use #414. Sorry to derail the discussion over selector mechanism. Thank you.

Chris00:
On Thu, 27 Feb 2014 02:23:10 -0800, Anil Madhavapeddy wrote:

> On 27 Feb 2014, at 10:14, David Sheets notifications@github.com wrote:
> 
> > The site could respect accept-language header and set/read a user
> > preference cookie to automatically select the appropriate
> > translation for a user. If we want these features, someone should
> > probably make a new ticket.
> 
> I agree with this, given that it's the recommended mechanism of serving
> multilingual content in RFC2616 (and clarified in RFC2295 and 2296).

I think it is nice to have this mechanism the first time you land on
the site but then, you must also be able to change the language by
hand and that choice should be remembered the next time you visit the
site.  One must also decide whether this is put in palce only for the
main page or for all index pages.

avsm:
On 27 Feb 2014, at 17:14, Christophe Troestler notifications@github.com wrote:

> On Thu, 27 Feb 2014 02:23:10 -0800, Anil Madhavapeddy wrote:
> 
> > On 27 Feb 2014, at 10:14, David Sheets notifications@github.com wrote:
> > 
> > > The site could respect accept-language header and set/read a user
> > > preference cookie to automatically select the appropriate
> > > translation for a user. If we want these features, someone should
> > > probably make a new ticket.
> > 
> > I agree with this, given that it's the recommended mechanism of serving
> > multilingual content in RFC2616 (and clarified in RFC2295 and 2296).
> 
> I think it is nice to have this mechanism the first time you land on
> the site but then, you must also be able to change the language by
> hand and that choice should be remembered the next time you visit the
> site. One must also decide whether this is put in palce only for the
> main page or for all index pages.

Yes, I agree with this and pretty much all of @dbuenzli's points on the matter (here and in #414).

Chris00:
On Thu, 27 Feb 2014 03:23:24 -0800, Daniel Bünzli wrote:

> Also don't forget to put alternate hreflang link tags in the header of all the pages to advertise the translations to machines. See e.g. google's instructions.

The <html> tag has (automatically generated) “lang” and “xml:lang” attributes.  Is this what you mean?

dsheets:
No, the Google link describes a way to use `<link>` tags to declare alternative versions of the page in other languages.

Chris00:
@dsheets I tried to add `hreflang` links in 5e712b9.  It is not clear whether relative URLs are accepted though.  Feel free to comment.

dbuenzli:
According to the html spec the `href` attribute of the `link` tag should be an IRI _reference_, so a relative URL should be fine. I checked on the basics page and the alternate links look fine.

agarwal:
I'm interested in closing issues that don't have clear actionable items. Is there anything specific in this issue that someone is proposing to do?

Chris00:
I think the only point remaining is about  `Accept-Language` and this is tackled by #414   So IMHO this can be closed.

PR #413, opened 2014-02-26
Add a script to translate navbar.mpp

lefessan:
FWIK, there is currently no way to translate headers and footers. This pull-request provides a first solution for that:
- a script scripts/translate.ml takes the filename and a string as arguments, and returns a translation of the string in the language required by the filename extension
- this script is used around some strings in .mpp files to translate them

For every language, the script will have to be modified to add the corresponding translations. This could also be done by making the script read a translation file (TODO for next contributors).


script/translate.ml
```diff
@@ -0,0 +1,43 @@
+(* Small utility to extract the language of the filename. *)
+
+open Printf
+open Utils
+
+let language_of_ISO639 = function
```
Chris00: This function is already declared in `script/lang_of_filename.ml`.  Avoiding the duplication would be welcome.

script/translate.ml
```diff
@@ -0,0 +1,43 @@
...
+  | "ko" -> "Korean"
+  | "zh" -> "Chinese"
+  | l -> failwith(sprintf "Add the full language name for %S" l)
+
+let () =
+  let filename = Sys.argv.(1) in
+  let to_translate = Sys.argv.(2) in
+  let lang =
+    let _, lang, _ = prefix_lang_ext_of_filename (Filename.basename filename) in
+    if lang = "" then "en" else lang
```
Chris00, line 13: Since it is the second time this construction is used (it is also in `script/lang_of_filename.ml`), maybe one should factor it in `script/utils.ml`?

script/translate.ml
```diff
@@ -0,0 +1,43 @@
...
+        | "Packages" -> "Contributions"
+        | "Community" -> "Communauté"
+        | "Feedback" -> "Suggestions"
+        | "About This Site" -> "A propos de ce site"
+        | "Find Us on GitHub" -> "Dépôt GitHub"
+        | _ -> to_translate
+      end
+    | _ -> to_translate
+  in
+  Printf.printf "%s%!" translation
```
Chris00, line 36: You (rightly) opened `Printf`.  Do not put it here.

script/translate.ml
```diff
@@ -0,0 +1,43 @@
...
+  let to_translate = Sys.argv.(2) in
+  let lang =
+    let _, lang, _ = prefix_lang_ext_of_filename (Filename.basename filename) in
+    if lang = "" then "en" else lang
+  in
+  let translation =
+    match lang with
+    | "fr" ->
+      begin
+        match to_translate with
```
Chris00: IMHO this presentation will quickly become unmanageable when if the list of terms or languages grows.  Moreover, a warning for terms without translations would certainly be welcomed (one must even think about the possibility to optionally turn a missing translation into an error to help translators).

lefessan:
@Chris00 
My last commit fixes your two previous comments. 

For your last comment, I do agree it will become unmanageable on the long term, and I said it in my pull-request comment (TODO: read translations from translation files). However, I think we should use this simple system for now, and wait until enough translations are available to move to the next one (or find another volunteer to implement the other solution).

Chris00:
On Thu, 27 Feb 2014 02:11:31 -0800, Fabrice Le Fessant wrote:

> For your last comment, I do agree it will become unmanageable on the long term, and I said it in my pull-request comment (TODO: read translations from translation files). However, I think we should use this simple system for now, and wait until enough translations are available to move to the next one (or find another volunteer to implement the other solution).

The problem is that there is also the footer to translate.  This will already be less fun.  My point is that we could take a couple of days to think something that we can live with a little longer.  Reaching translators will probably involve making a bridge with tools they are familiar with, so maybe putting them in a file, say with line having the syntax

English:: French

is better?

pw374:
Just wondering: is it okay to merge this one now?

Chris00:
Yes.  I did the merge — I usually check that the site compiles and have a quick look to the result.

@lefessan Thanks for your work!

pw374:
Great :-)

PR #415, opened 2014-02-27
Fix Tutorials.Modules (fr) header overlaying breadcrumbs and lang selector

dsheets:
Apparently ---- underlining in markdown creates an H2 header which has different styling in bootstrap_mod.css than an H1 header such that the header box ends up catching most clicks on the breadcrumb or language selector links.


pw374:
thanks

PR #416, opened 2014-02-27
Fix Tutorials.Structure (ja) header overlaying breadcrumbs and lang selector

dsheets:
Same as #415


dsheets:
I've now gone through all the tutorials translations to check for this error. It is rather unfortunate that an error in the header level can disable navigation like this. There are a number of solutions from CSS to markdown checking to schematron checks on the generated XML. I don't know what the right process solution is to prevent these bugs...

Chris00:
@dsheets Feel free to suggest improvements to the CSS.  The problem comes from the fact that `h2` use [some trick](https://github.com/ocaml/ocaml.org/blob/master/site/css/bootstrap_mod.css#L16) not to fall under the header when they are referenced — for example from the TOC.  I tried to play with the `z-index` to no avail.

PR #422, opened 2014-02-28
Partial translation to French (2/3) of docs/index.md

lefessan:


site/docs/index.fr.md
```diff
@@ -0,0 +1,102 @@
+<!-- ((! set title Docs !)) ((! set documentation !)) -->
```
Chris00: I think you also want to `((! set nobreadcrumb !))`.  Landing pages are already clearly indicated by the header.

site/docs/index.md
```diff
@@ -1,5 +1,4 @@
-<!-- ((! set title Docs !)) ((! set documentation !))
-((! set nobreadcrumb !)) -->
```
Chris00, line 2: See above.

lefessan:
I added the breadcrumbs because otherwise, the translation links are hidden behind the blocks: while still being visible, they cannot be clicked. I actually noticed the same problem appears on learn/index.html and learn/index.fr.html (at least, when using Chrome). The only fix I could find without modifying the CSS was to add the breadcrumbs.

Chris00:
Ah OK.  Commit 378c09e should fix that.

lefessan:
Ok, I added the breadcrumbs back

Chris00:
Thanks.

PR #431, opened 2014-03-02
Fix home link of translated pages

lefessan:
Home link of translated pages was relative instead of being absolute

### DIFF
--- a/script/link_of_lang.ml
+++ b/script/link_of_lang.ml
@@ -6,11 +6,15 @@ let () =
   let filename = Sys.argv.(1) in
   let link = Sys.argv.(2) in
   let lang = Utils.lang_of_filename filename in
-  (* eprintf "Link = %S\n%!" link; *)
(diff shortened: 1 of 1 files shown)


Chris00:
BTW, you may be interested in putting

<pre>
(defun comment-dwim-line (&optional arg)
   "Replacement for the comment-dwim command.
If no region is selected and current line is not blank and we are
not at the end of the line, then comment current line. Replaces
default behavior of comment-dwim, when it inserts comment at the
end of the line."
   (interactive "*P")
   (comment-normalize-vars)
   (if (and (not (region-active-p)) (not (looking-at "[ \t]*$")))
       (comment-or-uncomment-region (line-beginning-position)
                    (line-end-position))
     (comment-dwim arg)))
(global-set-key "\M-;" 'comment-dwim-line)
</pre>

in your `~/.emacs` file.  It allows to (un)comment the current line with `M-;`.

PR #436, opened 2014-03-04
Fix "from source" installation method

smondet:
There should be at least a link on that page to get `ocaml-4.01.0.tar.gz`.

I removed “This is unlikely the method you want to use unless you are a developer of the OCaml compiler itself or trying to install on some esoteric platforms.” because there are many cases when one would prefer getting the original sources.

### DIFF
--- a/site/docs/install.md
+++ b/site/docs/install.md
@@ -537,7 +537,7 @@ See [here](https://github.com/hcarty/ocamlbrew).
 
 
 ## From Source
-Download the [latest development version](/releases/svn.html) using
-subversion and follow the instructions included therein. This is
-unlikely the method you want to use unless you are a developer of the
-OCaml compiler itself or trying to install on some esoteric platforms.
+
+Download the [source for your preferred OCaml release](/releases/)
+(or take the [development version](/releases/svn.html) using
+Subversion or Git) and follow the instructions included therein.


site/docs/install.md
```diff
@@ -537,7 +537,7 @@ See [here](https://github.com/hcarty/ocamlbrew).
 
 
 ## From Source
-Download the [latest development version](/releases/svn.html) using
-subversion and follow the instructions included therein. This is
-unlikely the method you want to use unless you are a developer of the
-OCaml compiler itself or trying to install on some esoteric platforms.
+
+Download the [source of the latest version](http://caml.inria.fr/ocaml/release.en.html)
```
Chris00: The link should be `/releases/4.01.0.html`.  However, one may not want to have to update it manually (one likely will forget).  So a script checking the latest version available in `/releases/` would be good.  Or else, change the text and simply point to `/releases`.

smondet: Oh I did not know there was http://ocaml.org/releases/ (there is no link from anywhere else, right?).

So this line should be

```
Download the [source of the release you want](/releases/)
```

?

should I add a commit or do you prefer a “rebase”?

Chris00: On Tue, 4 Mar 2014 15:03:35 -0800, Sebastien Mondet wrote:

> Oh I did not know there was http://ocaml.org/releases/ (there is no
> link from anywhere else, right?).

Indeed, I am afraid so.

> So this line should be
> 
> Download the [source of the release you want](/releases/)

Something like that — maybe a bit better (that does not sound too
engaging to me).

> should I add a commit or do you prefer a “rebase”?

I prefer you rebase.

smondet:
I've rebased the PR.

Chris00:
Merged as db5e4ad

PR #438, opened 2014-03-05
Make sure snippets are available before translated indexes are built

dsheets:
The first `make` used to fail on `ocaml.org/index.fr.html` because `template/front_code_snippet.html` was not built yet.

I think this new dependency layout is valid and gets rid of the special non-translated page dependencies.


Chris00:
Well spotted.  Merged as fff6265

PR #441, opened 2014-03-12
Consistently Link Box Headers

dsheets:
If boxes have 'more' links at the bottom, usually the box header will also link to that 'more' location. This PR links the "Recent Email Threads" h4 box under the "Mailing Lists" box on the Community page to the sympa mailing list archive. Unfortunately, perhaps, the style sheet does not treat an h4 link like an h1 link and the title is now regular link color instead of charcoal. I think it looks fine but maybe you don't like it.

Also, I replaced tabs with spaces and cleaned up some excessive indentation/weird line breaks in the second commit. It makes that diff ugly but I think it helps diffs in general when there aren't tabs to make things _really_ ugly. :-P

### DIFF
--- a/site/community/index.md
+++ b/site/community/index.md
@@ -42,10 +42,11 @@
                   <a href="https://sympa.inria.fr/sympa/arc/caml-list/">
                     Recent Email Threads
                   </a>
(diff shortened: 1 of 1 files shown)


Chris00:
Thanks.

PR #475, opened 2014-05-11
Oasis tutorial

Drup:
I'm working on an oasis tutorial for setting up a project quickly. The target of the tutorial is newcomers to ocaml who just want to get stuff to work. The motivation for this was "I'm repeating the same thing on #ocaml every time a beginner want help to set up a project, so I should write it down." :)

I would like to collect early opinions both on grammar/spelling and on the content. I'm also looking for other interesting oasis use cases for the example section.

Few questions :
- Should I use a more meaningful example than "foo" and "bar" for the imaginary library I'm describing ? In this case, does anyone have an idea ?
- Should I add a "Tips" sections, for very small but sometime useful informations (@jpdeplaix mentioned "how to remove the symlinks", for example) ? We could also mention some common pitfalls here (I'm waiting for propositions). 

I will rebase everything before merge, in order to make the log clean.

cc @gildor478

### DIFF
--- a/site/learn/tutorials/setting_up_with_oasis.md
+++ b/site/learn/tutorials/setting_up_with_oasis.md
@@ -10,11 +10,15 @@
 
 You first need to install OASIS. The preferred way is with opam:
 
-`opam install oasis`.
+```
+opam install oasis
+```
 
 You can also install it under debian-based system.
 
-`sudo apt-get install oasis`
+```
+sudo apt-get install oasis
+```
 
 Finally, a tarball is available on [OASIS's main page](http://oasis.forge.ocamlcore.org/).
 
@@ -66,14 +70,18 @@ Executable "foo"
 
 Now that the `_oasis` file is defined, we can generate a build system. At the root of your project, run the following command:
 
-`oasis setup -setup-update dynamic`
+```
+oasis setup -setup-update dynamic
+```
 
 This will produce the following files : `setup.ml`, `configure` and a `Makefile`.
 These files are the only one necessary for OASIS to work, and you should put only those in a version control system. OASIS will produce some other files during building that you can safely ignore. More details about the various OASIS files are given [here](#Theoasisfiles).
 
 You can now build your project by running:
 
-`make`
+```
+make
+```
 
 It will produce the `foo.byte` file than you can execute.
 
--- a/site/learn/tutorials/setting_up_with_oasis.md
+++ b/site/learn/tutorials/setting_up_with_oasis.md
@@ -1,0 +1,421 @@
+<!-- ((! set title Setting up with OASIS !)) ((! set learn !)) -->
+
+*Table of contents*
+
+# Setting up an OCaml project with OASIS
+
+[OASIS](http://oasis.forge.ocamlcore.org/) allows to give a high level description of your project and will generate a build system and packaging information for you. It allows to set up a project in a quick and simple way. The goal of this tutorial is not to explain all the details of OASIS, but to provide the necessary information to setup a project.
+
+## Getting started
+
+You first need to install OASIS. The preferred way is with opam:
+
+`opam install oasis`.
+
+You can also install it under debian-based system.
+
+`sudo apt-get install oasis`
+
+Finally, a tarball is available on [OASIS's main page](http://oasis.forge.ocamlcore.org/).
+
+### What is your project ?
+
+Your project description is going to be located in an `_oasis` file, at the root of your project. Let's start by a general description of your project!
+
+```oasis
+Name: FooBar
+Version: 0.1
+Synopsis: A library about Foo and Bar.
+Authors: Me <me@myhost.com>
+License: LGPL with OCaml linking exception
+
+Description: FooBar allows to combine foo and bar efficiently. It also allows to interface with baz.
+Homepage: http://foobar.org
+```
+
+This description might seem unnecessary at first, but will be mandatory if you want to package and distribute your library later on. It's also a nice source of information for the interested user.
+
+Only the first set of field is mandatory. You can find a list of all the available licenses [here](http://oasis.forge.ocamlcore.org/MANUAL.html#license).
+
+### A little bit of OASIS boilerplate
+
+OASIS needs some information to proceed. We will come back and explain the details [later on](#PluginsandFeatures). Just copy this part to your `_oasis` for now.
+
+```oasis
+OASISFormat: 0.4
+BuildTools: ocamlbuild
+Plugins: META (0.4), DevFiles (0.4)
+```
+
+From now on, we will assume that the `_oasis` contains everything described previously.
+
+
+## The meat of your project
+
+### An executable
+
+An `_oasis` file is organized in sections, each section defining a small part of your project. `_oasis` files are white-space sensitive, fields in a section must be properly indented.
+
+We will start by defining an executable `foo` with sources located in an `src` directory.
+
+```oasis
+Executable "foo"
+  Path: src
+  MainIs: foo.ml
+```
+
+Now that the `_oasis` file is defined, we can generate a build system. At the root of your project, run the following command:
+
+`oasis setup -setup-update dynamic`
+
+This will produce the following files : `setup.ml`, `configure` and a `Makefile`.
+These files are the only one necessary for OASIS to work, and you should put only those in a version control system. OASIS will produce some other files during building that you can safely ignore. More details about the various OASIS files are given [here](#Theoasisfiles).
+
+You can now build your project by running:
+
+`make`
+
+It will produce the `foo.byte` file than you can execute.
+
+To produce native executables and use external libraries, we add some fields to the section:
+
+```oasis
+Executable "foo"
+  Path: src
+  MainIs: foo.ml
+  CompiledObject: best
+  BuildDepends:
+    some_lib,
+    some_other_lib
+```
+
+`CompiledObject` can take the values `byte`, `native` or `best`. `best` is equal to `native` if it is available or `byte` otherwise. When the compilation is native, the produced binary is called `foo.native`.
+`BuildDepends` is a comma separated list of ocamlfind packages. [ocamlfind](http://ocaml.org/learn/tutorials/dev_tools.html#Findlibocamlfind) handles packages locally in your system and is used by opam. You can list ocamlfind packages with `ocamlfind list`.
+
+### Defining libraries
+
+To define a library, we use the `Library` section. Here we will define two libraries with some inter dependencies.
+
+```oasis
+Library "foo"
+  Path: foo_src
+  Modules: Foo
+  InternalModules: InternalFoo
+
+Library "bar"
+  Path: bar_src
+  Modules: Bar
+  InternalModules:
+    InternalBar, InternalFooBar
+  BuildDepends: foo, unix
+```
+
+As you can see, `BuildDepends` is also used for intra-project dependencies: the bar library is dependent on the foo library.
+`Modules` and `InternalModules` are comma separated list of modules. They should contain only capitalized name with no file extensions. They can include module provided by any type of file, being `.ml`, `.mli`, `.mll` or `.mly` (for [ocamllex, ocamlyacc](http://caml.inria.fr/pub/docs/manual-ocaml/lexyacc.html) and [menhir](http://cristal.inria.fr/~fpottier/menhir/)).
+`InternalModules` is used to list modules that are part of the library but should be kept hidden and not be part of the public interface of the library.
+
+`make` builds the various libraries and `make install` installs them. OASIS handles the installation using ocamlfind.
+
+### Other optional sections
+
+Executable and Library are enough to create a basic project. The following sections allow to provide more information and build extra things to distribute your project.
+
+#### Documentation
+
+Documentation is a very important part of a library. OASIS allows to build an [ocamldoc](http://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html)-powered HTML documentation for the complete project quite easily.
+
+```oasis
+AlphaFeatures: ocamlbuild_more_args
+Document "foobar_api"
+  Type: ocamlbuild (0.4)
+  BuildTools: ocamldoc
+
+  Title: API reference for FooBar
+  XOCamlbuildPath: .
+  XOCamlbuildExtraArgs:
+    "-docflags '-colorize-code -short-functors -charset utf-8'"
+  XOCamlbuildLibraries: foo, bar
+```
+
+The important part is the `XOCamlbuildLibraries` field, which is a list of ocamlfind libraries defined in the `_oasis` file.
+`XOCamlbuildExtraArgs` allows to give some ocamldoc command line flags. Those presented here are considered essential.
+Running `make doc` will create a folder `foobar_api.docdir` which contains the documentation as HTML files.
+
+#### SourceRepository and `.gitignore`
+
+As meta-data, OASIS also allows to define the address of the source repository. This can be used by packaging tools.
+
+```oasis
+SourceRepository master
+  Type: git
+  Location: https://my_git_host.org/foobar.git
+  Branch: master
+  Browser: https://my_git_host.org/foobar
+```
+
+While we are on the subject of version control system, here is an example of `.gitignore` file that will ignore just the right file for your project using OASIS.
+
+```.gitignore
+_build
+*.native
+*.byte
+*.docdir
+
+setup.data
+setup.log
+```
+
+## Going further
+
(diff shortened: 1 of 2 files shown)


site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,420 @@
+<!-- ((! set title Setting up with oasis !)) ((! set learn !)) -->
```
gildor478: I tend to use OASIS, all capitalized. This is/was an acronym... I think this is consistent with the website. 

Whenever I use "oasis", it is about the binary.

site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,420 @@
...
+
+### A little bit of oasis boilerplate
+
+Oasis needs some information to proceed. We will come back and explain the details [later on](#PluginsandFeatures). Just copy this part to your `_oasis` for now.
+
+```oasis
+OASISFormat: 0.4
+BuildTools: ocamlbuild
+Plugins: META (0.4), DevFiles (0.4)
+AlphaFeatures: compiled_setup_ml
```
gildor478: Is this feature really mandatory to use an AlphaFeature in a tutorial... This is probably too advanced.

site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,420 @@
...
+To produce native executables and use external libraries, we add some fields to the section:
+
+```oasis
+Executable "foo"
+  Path: src
+  MainIs: foo.ml
+  CompiledObject: best
+  BuildDepends:
+    some_lib,
+    some_other_lib
```
gildor478, line 97: Totally optional convention:
BuildDepends: some_lib,
                        some_other_lib

Start at first line and align with it...

site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,420 @@
...
+Library "foo"
+  Path: foo_src
+  Modules: Foo
+  InternalModules: InternalFoo
+
+Library "bar"
+  Path: bar_src
+  Modules: Bar
+  InternalModules:
+    InternalBar, InternalFooBar
```
gildor478, line 117: See the "totally optional convention" above.

site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,420 @@
...
+`make` builds the various libraries and `make install` installs them. Oasis handles the installation using ocamlfind.
+
+### Misc sections
+
+#### Documentation
+
+Documentation is a very important part of a library. Oasis allows to build an [ocamldoc](http://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html)-powered HTML documentation for the complete project quite easily.
+
+```oasis
+AlphaFeatures: ocamlbuild_more_args
```
gildor478, line 136: Same here, beware of using AlphaFeatures. I know the code will look nicer with the given flags, but maybe it should just be a Tips.

site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,420 @@
...
+  BuildDepends: foo, unix
+```
+
+As you can see, `BuildDepends` is also used for intra-project dependencies: the bar library is dependent on the foo library.
+`Modules` and `InternalModules` are comma separated list of modules. They should contain only capitalized name with no file extensions. They can include module provided by any type of file, being `.ml`, `.mli`, `.mll` or `.mly` (for [ocamllex, ocamlyacc](http://caml.inria.fr/pub/docs/manual-ocaml/lexyacc.html) and [menhir](http://cristal.inria.fr/~fpottier/menhir/)).
+`InternalModules` is used to list modules that are part of the library but should be kept hidden and not be part of the public interface of the library.
+
+`make` builds the various libraries and `make install` installs them. Oasis handles the installation using ocamlfind.
+
+### Misc sections
```
gildor478: Misc sections -> Other optional sections ?

I think past this point almost everything is optional.

Maybe also add a small sentence at the beginning:
"Executable and Library are enough to create a basic project. The following sections allow to provide more information and build extra things to distribute your project."

site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,420 @@
...
+
+```.gitignore
+_build
+*.native
+*.byte
+*.docdir
+
+setup.data
+setup.log
+setup.exe
```
gildor478: This "setup.exe" is a little bit confusing, esp for Linux user that will not understand why. This is one of the part about compiled_setup_ml that I am not super happy  with.

site/learn/tutorials/setting_up_with_oasis.md
kit-ty-kate: I don't think so. Sometimes users give up OASIS because it's too slow. At least, this feature fix this issue.

site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,420 @@
...
+lwt
+lwt.extra
+lwt.glib
+lwt.preemptive
+lwt.syntax
+lwt.syntax.log
+lwt.syntax.options
+```
+
+It is possible to define such hierarchy in Oasis using two additional fields: `FindLibName` and `FindLibParent`.
```
gildor478: There is also something called FindlibContainers, but it is maybe too advanced.

site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,420 @@
...
+Notice the `$:` which is specific for boolean fields.
+Conditional values are explained in details [here](http://oasis.forge.ocamlcore.org/MANUAL.html#conditional-value).
+
+### Tests
+
+Oasis can also handle tests, be them unit tests or not. First we want to test our `foo` executable defined earlier.
+
+```oasis
+Test "test_foo"
+  Command: $foo test/some_test_input
```
gildor478, line 246: Add
  TestTools: foo

You depend on it through $foo.

site/learn/tutorials/setting_up_with_oasis.md
kit-ty-kate, line 97: The current convention is pretty wide used.

site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,420 @@
...
+  Path: test
+  MainIs: test_bar.ml
+  Build$: flag(tests) && flag(baz)
+  CompiledObject: best
+  Install: false
+  BuildDepends: bar, baz, bar.baz, oUnit
+
+Test "test_bar_and_baz"
+  Run$: flag(tests) && flag(baz)
+  Command: $test_bar_and_baz
```
gildor478, line 266: - TestTools

site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,420 @@
...
+```oasis
+Library "foo"
+  Path: foo
+  Modules: FooMore
+  BuildDepends: foo.core, lwt, lwt.syntax
+```
+
+Oasis will deduce that camlp4 must be used.
+
+If the package's name doesn't end by `.syntax` (which is pretty unlikely), you can add this to the `_tags` file:
```
gildor478: outside of OASIS_START/STOP.

site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,420 @@
...
+### The oasis files
+
+Oasis generates all the file needed by an OCaml installation, which is quite an important amount. In this section, we will give a detailed description of each files.
+
+All these files will contain an Oasis section that should not be modified.
+
+- Oasis
+ - `setup.ml` contains the logic of the build system.
+ - `Makefile` and `configure` are convenient files produced by the plugin `DevFiles`. They call `setup.ml`.
+ - `setup.log` and `setup.data` are produced during the build process.
```
gildor478: - `setup.log` is produced by the configure step.
- `setup.data` is produced during the build process.

site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,420 @@
...
+
+- OCamlfind
+ - `META` is the library specification for ocamlfind.
+
+### Setup options
+
+`oasis setup` accepts an option `-setup-update` defined in the help as:
+
+```
+  -setup-update {none|weak|dynamic}  Define the way `setup.ml` should update
```
gildor478, line 397: You may ignore 'weak' as I intend to deprecate it sooner or later.

site/learn/tutorials/setting_up_with_oasis.md
gildor478: Are we talking about 1s or 10s here ?

I can understand people want top speed but if you only win 1s, I think this is overkill for a beginner.

kit-ty-kate: It depends on the machine, but for me it's more 1-2s (with an i5 laptop).
I think we should invite beginners to have the « good habits ».

gildor478: I would be intersted in having this data somewhere. More than 1s is really not expected and is a bug for me. I am a little bit spoiled on this topic because I use a desktop i7 + SSD. Probably not a very good choice to see speed problem.

Although, the solution should not be to use compiled_setup_ml but to optimize OASIS itself (after all what we do in there is not complex or time consuming).

site/learn/tutorials/setting_up_with_oasis.md
gildor478, line 97: The rendering of my code snippet was not good:

``` oasis
BuildDepends: some_lib,
              some_other_lib
```

But I'll be fine with whatever convention.

site/learn/tutorials/setting_up_with_oasis.md
Drup, line 397: Fine, I was not sure if it was still recommended or not.

site/learn/tutorials/setting_up_with_oasis.md
Drup: To be honest, I don't know what FindlibContainers is supposed to do.

site/learn/tutorials/setting_up_with_oasis.md
Drup, line 136: I will remove it if you give me a field "Docflags" :p

I _really_ don't want to talk about myocamlbuild customization at this point (which is the only way to pass those arguments).
And producing a non utf8 html documentation in 2014 is just nonsense.

site/learn/tutorials/setting_up_with_oasis.md
Drup: I was also somehow not sure to include it, but consider the following points: 
1. It doesn't change the semantic of the build system whatsoever.
2. As @jpdeplaix pointed, the speed difference is significant (quite more than 1s, especially on a HDD, or worse: on an unplugged laptop with battery saving turned on, which is a relatively frequent situation for me).
3. I have big hopes that it's going to be used by default in Oasis sooner or later (and then I will remove it from the tutorial).
4. At this point of the tutorial, the invocation is no more magic than  `Plugins: META (0.4), DevFiles (0.4)` for a newcomer.
5. It's stable in practice.

I don't see any downside, except the fact that the name "Alpha" is slightly scary.

site/learn/tutorials/setting_up_with_oasis.md
Drup: Of what is OASIS the acronyme ? I can't find it in the documentation.

Drup:
Thanks for all the remarks @gildor478 !
Do you feel there are interesting features that are missing here and would deserve to be presented ?

site/learn/tutorials/setting_up_with_oasis.md
kit-ty-kate: With a simple _oasis file (one library, no dependencies and one module), the result is the following (respectively with `compiled_setup_ml` and without):

```
$ time make 
./setup.exe -build
Finished, 0 targets (0 cached) in 00:00:00.
Finished, 6 targets (5 cached) in 00:00:00.
make  0.24s user 0.02s system 88% cpu 0.293 total

$ time make
ocaml setup.ml -build 
Finished, 0 targets (0 cached) in 00:00:00.
Finished, 6 targets (5 cached) in 00:00:00.
make  2.01s user 0.05s system 98% cpu 2.096 total
```

site/learn/tutorials/setting_up_with_oasis.md
gildor478: Because it is very old and probably not documented. Before OASIS was called OCamlAutobuild. And _oasis was the name of the toplevel file.
It stands for
OCaml Autobuild Source Information System 

Don't ask the real meaning, I just wanted something that will be right for a filename (_oasis) and related to the desert...

https://forge.ocamlcore.org/docman/view.php/54/94/SlidesCamlConsortium2009.pdf

During this precise consortium session, we decided that OCamlAutobuild was not a right name and change it to OASIS...

site/learn/tutorials/setting_up_with_oasis.md
gildor478: OK, so we are talking about of 1.8s overhead. I get it. It is not totally crazy and can be improved. In the context of this tutorial, I think using compiled_setup_ml should be a "tip" rather than a mandatory field. This will just help to not have to change it when we will make it set by default.

kit-ty-kate: Do you really think it can be improved ? I personally think that it's just the usual cost of using the interpreter.

gildor478: It starts to be OT wrt to this tutorial, but yes I expect that there are some way to improve it -- esp. replacing the RecDescParser by something more efficient. I suppose you use the dynamic version in both case ?

kit-ty-kate: No, I'm not using the dynamic version. All my libraries works with the normal setup (and this test don't break the rule :D).

gildor478: If you are using the big setup.ml my former remark about RecDescParser doesn't apply. But we should improve the number of lines in the generated setup.ml, so that the toplevel interpreter will have less to parse. This is also something that will have to be done at one point. This alone can give a pretty good speed up.

In the past, I never paid attention to the performance of OASIS. That something that need to be improved.

Drup: Ok, fair enough, I removed mentions of the compiled setup, except in the very last part of the tutorial about Plugins and Features.

Drup:
I think I fixed all the remarks (except the indentation one, but I think it's fine as it is).

Does it needs more stuff to be merged ?

gildor478:
That looks very good to me. Thanks for your work.

I think there is no problem with merging.

Drup:
Great! Let me rebase everything.

Drup:
It's ready. Merge whenever you want to.

site/learn/tutorials/setting_up_with_oasis.md
```diff
@@ -0,0 +1,421 @@
...
+
+# Setting up an OCaml project with OASIS
+
+[OASIS](http://oasis.forge.ocamlcore.org/) allows to give a high level description of your project and will generate a build system and packaging information for you. It allows to set up a project in a quick and simple way. The goal of this tutorial is not to explain all the details of OASIS, but to provide the necessary information to setup a project.
+
+## Getting started
+
+You first need to install OASIS. The preferred way is with opam:
+
+`opam install oasis`.
```
Chris00: Why isn't this just indented by 4 spaces?  Is there a reason not to want `<pre>`?

Drup: Hum, I thought ` on a single line would be equivalent.

Chris00: On Thu, 15 May 2014 07:32:36 -0700, Drup wrote:

> > +`opam install oasis`.
> 
> Hum, I though ` on a single line would be equivalent.

No.  That starts a new paragraph with some code.  Indenting by 4
spaces creates a <pre> block.

Drup: Fixed. I prefer to use ``` and newline than indentation, if you don't mind (I find it easier to view and modify).

samoht: Or use blocks surrounded by 4 backquotes

Chris00:
Thanks for your contribution.

PR #479, opened 2014-05-15
Add a precise deadline

garrigue:
Use UTC-11 rather than Samoan Standard Time, to avoid the confusion with Samoan Time (UTC+12)


amirmc:
Much appreciated!

PR #480, opened 2014-05-17
solution to depth-first-search

shayne-fletcher:
I've offered a depth-first-search solution on the '99 problems' page. Unfortunately, due to issues (https://github.com/ocaml/opam/issues/1410) installing opam-lib, haven't been able to see how it renders (but it should be fine).

### DIFF
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -2130,7 +2130,7 @@ module type GRAPH = sig
   type node = char
   type t
   val of_adjacency : (node * node list) list -> t
-  val dfs_fold : t -> node -> ('a -> (int * node) -> 'a) -> 'a -> 'a
+  val dfs_fold : t -> node -> ('a -> node -> 'a) -> 'a -> 'a
 end
 
 module M : GRAPH = struct
@@ -2153,36 +2153,32 @@ module M : GRAPH = struct
   }
 
   let dfs_fold g c fn acc =
-    let node u (t, state) =
-      let rec edge (t, state) v =
-        if Char_map.find v (state.color) = White then
-          dfs_visit t v {state with pred=(Char_map.add v u (state.pred))}
+    let rec dfs_visit t u {d; f; pred; color; acc} =
+      let edge (t, state) v =
+        if Char_map.find v state.color = White then
+          dfs_visit t v {state with pred=(Char_map.add v u state.pred)}
         else  (t, state)
-      and dfs_visit t u {d; f; pred; color; acc} =
-        let t, {d; f; pred; color; acc} =
-          let t = t + 1 in
-          List.fold_left edge
-            (t, {d=Char_map.add u t d; f;
-                 pred; color=Char_map.add u Gray color; acc})
-            (Char_map.find u g)
-        in
-        let t = t + 1 in
-        t , {d; f=(Char_map.add u t f); pred; 
-             color=Char_map.add u Black color; acc = fn acc (t, u)}
       in
-      if Char_map.find u (state.color) = White then dfs_visit t u state
-      else (t, state)
+      let t, {d; f; pred; color; acc} =
+        let t = t + 1 in
+        List.fold_left edge
+          (t, {d=Char_map.add u t d; f;
+               pred; color=Char_map.add u Gray color; acc})
+          (Char_map.find u g)
+      in
+      let t = t + 1 in
+      t , {d; f=(Char_map.add u t f); pred; 
+           color=Char_map.add u Black color; acc = fn acc u}
     in
     let v = List.fold_left (fun k (x, _) -> x :: k) [] (Char_map.bindings g) in
     let initial_state= 
-       {d=Char_map.empty;
-        f=Char_map.empty;
-        pred=Char_map.empty;
-        color=List.fold_right (fun x->Char_map.add x White) v Char_map.empty;
-        acc=acc}
+      {d=Char_map.empty;
+       f=Char_map.empty;
+       pred=Char_map.empty;
+       color=List.fold_right (fun x->Char_map.add x White) v Char_map.empty;
+       acc=acc}
     in
-    (snd (node c (0, initial_state))).acc
-
+    (snd (dfs_visit 0 c initial_state)).acc
 end
 
 (* Test *)
@@ -2196,7 +2192,7 @@ let g = M.of_adjacency
              'z',      ['z'] ;
             ]
 
-let l = M.dfs_fold g 'w' (fun acc (_, c) -> acc@[c]) [] 
+let l = List.rev (M.dfs_fold g 'w' (fun acc c -> c :: acc) [])
 ```
 
 #### Connected components. (*medium*)
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -2156,19 +2156,18 @@ module M : GRAPH = struct
     let rec dfs_visit t u {d; f; pred; color; acc} =
       let edge (t, state) v =
         if Char_map.find v state.color = White then
-          dfs_visit t v {state with pred=(Char_map.add v u state.pred)}
+          dfs_visit t v {state with pred=Char_map.add v u state.pred;}
         else  (t, state)
       in
       let t, {d; f; pred; color; acc} =
         let t = t + 1 in
         List.fold_left edge
           (t, {d=Char_map.add u t d; f;
-               pred; color=Char_map.add u Gray color; acc})
+               pred; color=Char_map.add u Gray color; acc = fn acc u})
           (Char_map.find u g)
       in
       let t = t + 1 in
-      t , {d; f=(Char_map.add u t f); pred; 
-           color=Char_map.add u Black color; acc = fn acc u}
+      t , {d; f=(Char_map.add u t f); pred; color=Char_map.add u Black color; acc}
     in
     let v = List.fold_left (fun k (x, _) -> x :: k) [] (Char_map.bindings g) in
     let initial_state= 


Chris00:
Did you test your code in isolation?  The build hangs when it is merged.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,146 @@ depth-first order).
...
+ with a direct call to [node] on the starting node of interest.
+
+ (2) To build the 'list of nodes reachable from the starting point',
+     prepend each node to a running list just when it's finishing time
+     has become known and it's about to be colored black.
+
+     That is...
+     {[
+        let t = t + 1 in
+            (* ... here. *)
```
Chris00: Please indent the code as you would normally.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,146 @@ depth-first order).
...
+
+module Char_set =
+  Set.Make (struct 
+    type t = char  
+    let compare=Pervasives.compare 
+  end)
+
+module Char_map =
+  Map.Make (struct 
+    type t = char
```
Chris00: Use the module `Char`.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,146 @@ depth-first order).
...
+        let t = t + 1 in
+            (* ... here. *)
+            t , {d; f=(Char_map.add u t f); 
+                  pred; color=Char_map.add u `Black color}
+     ]}
+*)
+
+module Char_set =
+  Set.Make (struct 
+    type t = char
```
Chris00: Use the module `Char`.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,146 @@ depth-first order).
...
+      d : int Char_map.t ; (*discovery time*)
+      f : int Char_map.t ; (*finishing time*)
+      pred : char Char_map.t ; (*predecessor*)
+      color : [`White|`Gray|`Black] Char_map.t ; (*vertex colors*)
+    }
+
+  let string_of_state {d; f; pred; color} =
+    let e_i k i acc = 
+      "'"^(String.make 1 k)^"'"^":"
+      ^(string_of_int i)^", "^acc  in
```
Chris00: This would be more readable with `sprintf`.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,146 @@ depth-first order).
...
+    let e_c k v acc =
+      "'"^(String.make 1 k)^"'"^":"
+      ^"'"^(String.make 1 (Char_map.find k pred))^"'"^", "^acc  in
+    "  d = {" ^ Char_map.fold e_i d "" ^ "}\n"^
+    "  f = {" ^ Char_map.fold e_i f "" ^ "}\n"^
+    "  pred ={" ^ Char_map.fold e_c pred "" ^ "}\n"
```
Chris00: Idem — use `sprintf`

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,146 @@ depth-first order).
...
+        let t = t + 1 in
+        t , {d; f=(Char_map.add u t f); 
+             pred; color=Char_map.add u `Black color}
+
+      in
+      if Char_map.find u color = `White then dfs_visit t u {d; f; pred; color}
+      else (t, {d; f; pred; color})
+
+    in 
+    let _, st = Char_set.fold node v (0, (initial_state v)) in st
```
Chris00: I think this is one of the few occasions where `snd` is better.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,146 @@ depth-first order).
...
+
+end
+
+(* Test *)
+
+let test_23_4 ()=
+  let g =
+    {
+      v = List.fold_right 
+        (Char_set.add)
```
Chris00: Please indent properly: `(Char_set.add)` is an argument of `List.fold_right`, it should be more indented.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,146 @@ depth-first order).
...
+let test_23_4 ()=
+  let g =
+    {
+      v = List.fold_right 
+        (Char_set.add) 
+        ['u'; 'v'; 'w'; 'x'; 'y'; 'z'] 
+        Char_set.empty 
+      ;
+      e = List.fold_right 
+        (fun (x, y) -> Char_map.add x y)
```
Chris00: Idem — indentattion.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,146 @@ depth-first order).
...
+module type S = sig
+  type state
+  val string_of_state : state -> string
+  val depth_first_search : graph -> state
+end
+
+module Dfs : S = struct
+
+  type state =
+    {
```
Chris00: I personally prefer the style

```
type state = {
    ...
  }
```

I do not see the point of the 2 "blank" lines.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,146 @@ depth-first order).
...
+         'y',      ['x'] ;
+         'z',      ['z'] ;
+        ]
+        Char_map.empty
+    }
+  in
+  let s = Dfs.depth_first_search g in
+  Printf.printf "%s\n" (Dfs.string_of_state s)
+
+let _ = test_23_4 ()
```
Chris00: Since you call `test_23_4` only for its side effects, use `let () = ...`.

shayne-fletcher:
Yes, the code has been tested in isolation :-/ Thanks for all the helpful comments Christophe! I'll work through them. I guess the best thing to do is make the changes and when done, issue another pull request?

shayne-fletcher:
Issued a second pull request containing improvements suggested in the previous revision's review comments

shayne-fletcher:
Hate to bother... Any word Christophe? Good to merge? Anything more needs doing here? Thanks!

Chris00:
On Tue, 20 May 2014 06:54:23 -0700, Shayne Fletcher wrote:

> Hate to bother... Any word Christophe? Good to merge? Anything more needs doing here? Thanks!

Sorry, I am very busy right now.  I'll try to take a look tomorrow.

shayne-fletcher:
Of course. Thank-you and please forgive my pushiness!

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,128 @@ depth-first order).
...
+  t , {d; f=(Char_map.add u t f); pred; color=Char_map.add u `Black color}
+  ]}
+
+*)
+module Char_set = Set.Make (Char)
+module Char_map = Map.Make (Char)
+
+type vertices = Char_set.t
+type adjacents = (char list) Char_map.t
+type graph = { v:vertices; e:adjacents }
```
Chris00: I wonder: why is there a `Char_set.t` to enumerate the vertices?  Aren't the vertices implicit in the `Char_map.t`, i.e. the list of vertices can be found using, say, `fst(Char_map.bindings adjacents)`?

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,128 @@ depth-first order).
...
+    pred : char Char_map.t ; (*predecessor*)
+    color : [`White|`Gray|`Black] Char_map.t ; (*vertex colors*)
+  }
+
+  let string_of_state {d; f; pred; color} =
+    let e_i k i acc = 
+     Printf.sprintf "'%c':%d, %s" k i acc in
+    let e_c k v acc =
+     Printf.sprintf "'%c':%c, %s" k (Char_map.find k pred) acc in
+    "  d = {" ^ Char_map.fold e_i d "" ^ "}\n"^
```
Chris00: Constructing a string by repeating the operation `^` (or `sprintf "... %s" ... acc`) is really inefficient.  Construct a list of individual strings and concatenate them at the end.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,128 @@ depth-first order).
...
+  val depth_first_search : graph -> state
+end
+
+module Dfs : S = struct
+
+  type state = {
+    d : int Char_map.t ; (*discovery time*)
+    f : int Char_map.t ; (*finishing time*)
+    pred : char Char_map.t ; (*predecessor*)
+    color : [`White|`Gray|`Black] Char_map.t ; (*vertex colors*)
```
Chris00: Why no simply use a variant?

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,128 @@ depth-first order).
...
+        let t = t + 1 in
+        t , {d; f=(Char_map.add u t f); 
+             pred; color=Char_map.add u `Black color}
+
+      in
+      if Char_map.find u color = `White then dfs_visit t u {d; f; pred; color}
+      else (t, {d; f; pred; color})
+
+    in 
+    snd (Char_set.fold node v (0, (initial_state v)))
```
Chris00: `(initial_state v)` → `initial_state v`

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,128 @@ depth-first order).
...
+    let e_i k i acc = 
+     Printf.sprintf "'%c':%d, %s" k i acc in
+    let e_c k v acc =
+     Printf.sprintf "'%c':%c, %s" k (Char_map.find k pred) acc in
+    "  d = {" ^ Char_map.fold e_i d "" ^ "}\n"^
+    "  f = {" ^ Char_map.fold e_i f "" ^ "}\n"^
+    "  pred ={" ^ Char_map.fold e_c pred "" ^ "}\n"
+
+  let depth_first_search {v; e} =
+    let initial_state v =
```
Chris00: No sure you want to make it a function.  After all, you apply it once on the `v` of `depth_first_search`.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,128 @@ depth-first order).
...
+      Char_set.fold (
+        fun n {d;f;pred;color} -> 
+          { d; f; pred; color = Char_map.add n `White color }
+      ) v {d=Char_map.empty; f=Char_map.empty; 
+           pred=Char_map.empty; color=Char_map.empty} 
+    in
```
Chris00: If you use the fact that the vertices are implicit in the map, isn't this equivalent to

``` ocaml
let v = { d = Char_map.empty;  f = Char_map.empty;  pred = Char_map.empty;
          color= Char_map.map (fun _ -> White) e}
```

site/learn/tutorials/99problems.md
shayne-fletcher: That representation doesn't admit the possibility of isolated nodes.

site/learn/tutorials/99problems.md
shayne-fletcher: You mean something like the following?

``` ocaml

type color_t = | White | Gray | Black
```

I reasoned that using a polymorphic variant here respects the notion that the enumeration as stated is not exhaustive of all possible colors. That said, if you argue that it induces a negative effect in terms of complexity or efficiency, I'm OK with changing it.

site/learn/tutorials/99problems.md
shayne-fletcher: Noted. How's this?

``` ocaml

let string_of_state {d; f; pred; color} =
  let cat = String.concat ", " in
  " d = {"^(cat (List.map (fun (x, y)->Printf.sprintf "'%c':'%d'" x y) (Char_map.bindings d)))^"}\n"^
  " f = {"^(cat (List.map (fun (x, y)->Printf.sprintf "'%c':'%d'" x y) (Char_map.bindings f)))^"}\n"^
  " pred = {"^(cat (List.map (fun (x, y)->Printf.sprintf "'%c':'%c'" x y) (Char_map.bindings pred)))^"}\n"

```

site/learn/tutorials/99problems.md
shayne-fletcher: Yes, as mentioned above modifying the representation such that we can rely on the vertices being implicit in the edge map excludes graphs that contain isolated nodes. That said, the spirit of what you said can be applied to nice effect (see comment above about inlining initialization):

``` ocaml

{ d=Char_map.empty; f=Char_map.empty; pred=Char_map.empty; color=Char_set.fold (fun x -> Char_map.add x `White) v Char_map.empty}

```

site/learn/tutorials/99problems.md
shayne-fletcher: You are right. Insight provided by the comment that follows means that initialization can be inlined without loss of readability.

site/learn/tutorials/99problems.md
shayne-fletcher: Will be inlining the production of the initial state as discussed above, redundant parentheses will be removed.

shayne-fletcher:
Have responded to your comments inline, made   improvements and issued a new pull request. Thanks again for your thoughtful reviews Christophe!

site/learn/tutorials/99problems.md
Chris00: What does `Char_map.add adjacent x []` represent then?

site/learn/tutorials/99problems.md
Chris00: In the end the type for `color` is closed anyway, so I do not see the point.  (There may also be a small performance cost in using polymorphic variants but this is less important than “doing the right thing”. ☺)

site/learn/tutorials/99problems.md
Chris00: On Wed, 21 May 2014 12:49:49 -0700, Shayne Fletcher wrote:

> In site/learn/tutorials/99problems.md:
> 
> > -    let e_i k i acc = 
> > -     Printf.sprintf "'%c':%d, %s" k i acc in
> > -    let e_c k v acc =
> > -     Printf.sprintf "'%c':%c, %s" k (Char_map.find k pred) acc in
> > -    "  d = {" ^ Char_map.fold e_i d "" ^ "}\n"^
> > -    "  f = {" ^ Char_map.fold e_i f "" ^ "}\n"^
> > -    "  pred ={" ^ Char_map.fold e_c pred "" ^ "}\n"
> >   +
> >   I see that function can be 'inlined'. I argue that the result of doing so makes the resulting program segment quite a lot harder to read and, quoting the 'Zen of Python', "readability counts" :)

It is not a matter of inlining.  Say you want to concatenate s₁,..., sₙ where (for simplicity) all these strings have length k.  If you do as above (((s₁ ^ s₂) ^ s₃) ^ ⋯ ^ sₙ), then you construct a string of length 2k, then 3k,... so in total you allocate 2k + 3k + ⋯ + nk = k · O(n²).  If you use String.concat, then only the final string of length k · n will be allocated.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,113 @@ depth-first order).
...
+    f : int Char_map.t ; (*finishing time*)
+    pred : char Char_map.t ; (*predecessor*)
+    color : [`White|`Gray|`Black] Char_map.t ; (*vertex colors*)
+  }
+
+  let string_of_state {d; f; pred; color} =
+    let cat = String.concat ", " in
+    " d = {"^(cat (List.map (fun (x, y)->Printf.sprintf "'%c':'%d'" x y) (Char_map.bindings d)))^"}\n"^
+    " f = {"^(cat (List.map (fun (x, y)->Printf.sprintf "'%c':'%d'" x y) (Char_map.bindings f)))^"}\n"^
+    " pred = {"^(cat (List.map (fun (x, y)->Printf.sprintf "'%c':'%c'" x y) (Char_map.bindings pred)))^"}\n"
```
Chris00: This is more efficient but still hard to read.  You see a lot of duplication in this code.  Also lines should be at most 80 chars long (easier to read, display fine on any device).  What about something like:

``` ocaml
let string_of_state {d; f; pred; color} =
  let open Printf in
  let bindings m fmt =
    let b = Char_map.bindings m in
    String.concat ", " (List.map (fun (x,y) -> sprintf fmt x y) b) in
  sprintf " d = {%s}\n f = {%s}\n pred = {%s}\n"
          (bindings d "'%c':'%d'") (bindings f "'%c':'%d'")
          (bindings pred "'%c':'%c'")
```

site/learn/tutorials/99problems.md
shayne-fletcher: Argh! You are right of course... I don't know what I was thinking yesterday :)

site/learn/tutorials/99problems.md
shayne-fletcher: Ok, that settles it. Will fix.

site/learn/tutorials/99problems.md
shayne-fletcher: Very nice

shayne-fletcher:
Changed the representation to eliminate the redundant vertex set and incorporated the improved `string_of_state` implementation. New pull request issued.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,122 @@ depth-first order).
...
+  let t = t + 1 in
+  (* ... here. *)
+  t , {d; f=(Char_map.add u t f); pred; color=Char_map.add u `Black color}
+  ]}
+
+*)
+
+module Char_map = Map.Make (Char)
+
+type graph = { e:(char list) Char_map.t }
```
Chris00: I am not sure the record should be kept.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,122 @@ depth-first order).
...
+      in
+      if Char_map.find u color = White then dfs_visit t u {d; f; pred; color}
+      else (t, {d; f; pred; color})
+    in
+    let v = List.fold_left (fun acc (x, _) -> x::acc) [] (Char_map.bindings e)
+    in
+    snd (
+      List.fold_right node v
+        (0,
+         {d=Char_map.empty;
```
Chris00: I think it would be clearer if you bound the record and put `snd (List.fold_right node v (0, init_state))` on a single line.

site/learn/tutorials/99problems.md
shayne-fletcher: Thought you might say that :)

site/learn/tutorials/99problems.md
shayne-fletcher: Yep. You got it. Give me 5 minutes.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,122 @@ depth-first order).
 <!-- SOLUTION -->
```
Chris00, line 2126: Note this commented "solution button".  You should distinguish what you hide and the test you show.  Printing the state is rather part of the solution — to help the reader to figure out how it works.  Since the question asks for a  transversal, a `fold` function visiting the nodes in depth-first order may be used to design various examples.

Chris00:
There are still some parts of the code I'd like to read more carefully.  I'll do that as soon as I can.
Thanks for your efforts.

shayne-fletcher:
Of course Christophe and no need to thank, in fact it is I that should like to thank you. This is proving to be a most beneficial experience for me :) New pull request made

site/learn/tutorials/99problems.md
shayne-fletcher, line 2126: Sorry I might be being a bit slow :( Could I ask you to be specific about precisely what you'd like me to do here?

shayne-fletcher:
I appreciate you're busy Christophe but this is felling stalled now which is a little discouraging. Is there anything I can do to advance this project?

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,120 @@ depth-first order).
...
+  snd (List.fold_right node v initial_state)
+  ]}
+  with a direct call to [node] on the 'source' vertex of interest ([s]
+  say)
+  {[
+  snd (node s initial_state)
+  ]}.
+
+  (2) To build the 'list of nodes reachable from the starting point',
+  prepend each node to a running list just when it's finishing time
```
avsm: grammar: "it's" -> "its"

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,120 @@ depth-first order).
...
+           'v',      ['y'] ;
+           'w', ['z'; 'y'] ;
+           'x',      ['v'] ;
+           'y',      ['x'] ;
+           'z',      ['z'] ;
+          ]
+          Char_map.empty
+  in
+  let s = Dfs.depth_first_search g in
+  Printf.printf "%s\n" (Dfs.string_of_state s)
```
avsm: Minor: This can just be `print_endline (Dfs.string_of_state s)` to avoid all the Printf machinery

Chris00:
On Thu, 5 Jun 2014 04:43:39 -0700, Shayne Fletcher wrote:

> I appreciate you're busy Christophe but this is felling stalled now which is a little discouraging. Is there anything I can do to advance this project?

Sorry, I'll try to have a look this WE.  From memory, what I recall is that it is not clear what to show the reader in order to illustrate what the function should do.  One cannot show the internal state because we want people to be able to implement the function their own way.  One may however say that the tree must be specified by its adjacency list and I was thinking a “fold” function would be OK.

shayne-fletcher:
Appreciate your comments Anil and thanks for your time Christophe. I'll stand by - looking forward to hearing your thoughts post the weekend.

shayne-fletcher:
Are you thinking along the lines of something like this Christophe?

``` ocaml

module Char_map = Map.Make (Char)
type graph = (char list) Char_map.t

module type S = sig
  type 'a state
  val fold : ('a -> char -> 'a) -> 'a -> graph -> 'a
end

module Dfs : S = struct

  type colors = White|Gray|Black

  type 'a state = {
    d : int Char_map.t ; (*discovery time*)
    f : int Char_map.t ; (*finishing time*)
    pred : char Char_map.t ; (*predecessor*)
    color : colors Char_map.t ; (*vertex colors*)
    acc : 'a ; (*user specified type used by 'fold'*)
  }

  let fold fn acc g =
    let node u (t, {d; f; pred; color; acc}) =
      let rec dfs_visit t u {d; f; pred; color; acc} =
        let edge (t, {d; f; pred; color; acc}) v =
          if Char_map.find v color = White then
            dfs_visit t v {d; f; pred=(Char_map.add v u pred); color; acc}
          else  (t, {d; f; pred; color; acc})
        in
        let t, {d; f; pred; color; acc} =
          let t = t + 1 in
          List.fold_left edge
            (t, {d=Char_map.add u t d; f;
                 pred; color=Char_map.add u Gray color; acc})
            (Char_map.find u g)
        in
        let t = t + 1 in
        t , {d; f=(Char_map.add u t f); pred; color=Char_map.add u Black color; acc = fn acc u}
      in
      if Char_map.find u color = White then dfs_visit t u {d; f; pred; color; acc}
      else (t, {d; f; pred; color; acc})
    in
    let v = List.fold_left (fun k (x, _) -> x :: k) [] (Char_map.bindings g) in
    let initial_state= 
       {d=Char_map.empty;
        f=Char_map.empty;
        pred=Char_map.empty;
        color=List.fold_right (fun x->Char_map.add x White) v Char_map.empty;
        acc=acc}
    in
    (snd (List.fold_right node v (0, initial_state))).acc

end

(* Test *)

let () =
  let string_of_list 
      (f:'a -> string) 
      (l:'a list) : string = 
    "[" ^ String.concat ";" (List.map f l) ^ "]"  
  in

  let g =
       List.fold_right
          (fun (x, y) -> Char_map.add x y)
          ['u', ['v'; 'x'] ;
           'v',      ['y'] ;
           'w', ['z'; 'y'] ;
           'x',      ['v'] ;
           'y',      ['x'] ;
           'z',      ['z'] ;
          ]
          Char_map.empty
  in

  let l = Dfs.fold (fun acc c -> c :: acc) [] g in
  print_endline (string_of_list (fun c -> String.make 1 c) l)
```

(This program prints "[w;z;u;v;y;x]" - verified to be a correct for the example).

shayne-fletcher:
If indeed this is consistent with what you were thinking, it seems to me that this generalizes quite nicely:

``` ocaml

module Dfs = struct
  module type S = sig
    type t
    module Node_map : Map.S with type key = t
    val fold : ('a -> t -> 'a) -> 'a -> (t list) Node_map.t -> 'a
  end

  module Make (M : Map.OrderedType) = struct
    (*...*)
  end
end

module Graph : Dfs.S with type t = char = Dfs.Make (Char)

let () =
 (* ... *)
  let l = Graph.fold (fun acc c -> c :: acc) [] g in
  print_endline (string_of_list (fun c -> String.make 1 c) l)
```

Chris00:
On Sat, 7 Jun 2014 06:36:15 -0700, Shayne Fletcher wrote:

> Are you thinking along the lines of something like this Christophe?

Absolutely.

>   let fold fn acc g =
>     let node u (t, {d; f; pred; color; acc}) =

The arguments of “node” are not used in the inner functions.  Maybe
you want to declare those before.

“{d; f; pred; color; acc}” → “state”

> ```
>   let rec dfs_visit t u {d; f; pred; color; acc} =
> ```

Maybe putting “edge” before “dfs_visit” in a mutual recursive binding
would make the code more readable?

> ```
>     let edge (t, {d; f; pred; color; acc}) v =
>       if Char_map.find v color = White then
>         dfs_visit t v {d; f; pred=(Char_map.add v u pred); color; acc}
>       else  (t, {d; f; pred; color; acc})
> ```

I think it would be clearer to write (because it clearly shows the
fields you are changing):

```
  let edge (t, state) v =
    if Char_map.find v color = White then
      dfs_visit t v {state with pred = Char_map.add v u pred}
    else  (t, state)
```

> ```
>     t , {d; f=(Char_map.add u t f); pred;
>          color=Char_map.add u Black color; acc = fn acc u}
> ```

Since you keep track of “t”, why not pass it to “fn”?  The statement
should also be modified accordingly.

> ```
>   in
>   if Char_map.find u color = White then dfs_visit t u {d; f; pred; color; acc}
>   else (t, {d; f; pred; color; acc})
> 
> in
> 
> let v = List.fold_left (fun k (x, _) -> x::k) [] (Char_map.bindings g) in
> ```
> 
> [...](snd %28List.fold_right node v %280, initial_state%29%29).acc

You did not specify a starting point.  Moreover you fold on all nodes
so that even isolated nodes will be processed.  This is not what the
statement asks.

> let _ =

You should prefer “let () =” which makes sure the code that follows
returns a unit type (it will catch if you forgets an argument for
example).

>   let string_of_list 
>       (f:'a -> string) 
>       (l:'a list) : string = 
>     "[" ^ String.concat ";" (List.map f l) ^ "]"  
>   in

This is not needed if one lets the toplevel print the result.

>   let g =
>        List.fold_right
>           (fun (x, y) -> Char_map.add x y)
>           ['u', ['v'; 'x'] ;
>            'v',      ['y'] ;
>            'w', ['z'; 'y'] ;
>            'x',      ['v'] ;
>            'y',      ['x'] ;
>            'z',      ['z'] ;
>           ]
>           Char_map.empty
>   in

This is what will be shown as an example to the user.  It would be
better that the representation of the graph does not “leak out” in
this way.  May I suggest that the signature S be somethinh like

module type GRAPH = sig
  type node = char
  type t
  val of_adjacency : (node \* node list) list -> t
  val dfs_fold : t -> node -> ('a -> node -> 'a) -> 'a -> 'a
end

(For “dfs_fold”, I used the “t must come first” principle:
https://blogs.janestreet.com/core-principles-uniformity-of-interface/
and added the starting node)

>   let l = Dfs.fold (fun acc c -> c :: acc) [] g in
>   Printf.printf "%s\n" (string_of_list (fun c -> String.make 1 c) l)
> 
> (This program prints "[w;z;u;v;y;x]" - verified to be a correct for the
> example).

You do not need to print explicitly, the toplevel will show the value
(I can do it after merging).

Also, your code on

  ['u', ['v'; 'x'] ;
   'v',      ['y'] ;
   'w', ['z'; 'y'] ;
   'x',      ['v'] ;
   'y',      ['x'] ;
   'z',      ['z'] ;
  ]

returns “[a; b; d; c]”, not what one would expect.

shayne-fletcher:
Thanks very much Christophe, I'm really pleased with the effect on the program on implementing your insights! Thanks also for spotting that silly bug!!

With the revised implementation, taking `'w'` as the source vertex, the top-level is now giving output:

``` ocaml
module type GRAPH =
  sig
    type node = char
    type t
    val of_adjacency : (node * node list) list -> t
    val dfs_fold : t -> node -> ('a -> int * node -> 'a) -> 'a -> 'a
  end
module M : GRAPH
val g : M.t = <abstr>
val l : M.node list = ['z'; 'v'; 'x'; 'y'; 'w']
```

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,79 @@ depth-first order).
...
+let g = M.of_adjacency
+            ['u', ['v'; 'x'] ;
+             'v',      ['y'] ;
+             'w', ['z'; 'y'] ;
+             'x',      ['v'] ;
+             'y',      ['x'] ;
+             'z',      ['z'] ;
+            ]
+
+let l = M.dfs_fold g 'w' (fun acc (_, c) -> acc@[c]) []
```
Chris00: One should accumulate normally `c :: acc` and reverse the list `l` after — what you do has quadratic cost.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,79 @@ depth-first order).
...
+    f : int Char_map.t ; (*finishing time*)
+    pred : char Char_map.t ; (*predecessor*)
+    color : colors Char_map.t ; (*vertex colors*)
+    acc : 'a ; (*user specified type used by 'fold'*)
+  }
+
+  let dfs_fold g c fn acc =
+    let node u (t, state) =
+      let rec edge (t, state) v =
+        if Char_map.find v (state.color) = White then
```
Chris00: (state.color) → state.color

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,79 @@ depth-first order).
...
+          dfs_visit t v {state with pred=(Char_map.add v u (state.pred))}
```
Chris00: (state.pred) → state.pred

Chris00: Is the `u` here correct (it is coming from `node`)?

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,79 @@ depth-first order).
...
+          List.fold_left edge
+            (t, {d=Char_map.add u t d; f;
+                 pred; color=Char_map.add u Gray color; acc})
+            (Char_map.find u g)
+        in
+        let t = t + 1 in
+        t , {d; f=(Char_map.add u t f); pred; 
+             color=Char_map.add u Black color; acc = fn acc (t, u)}
+      in
+      if Char_map.find u (state.color) = White then dfs_visit t u state
```
Chris00: Now that `node` is only used once on `c`, the state will obviously be `White`.

site/learn/tutorials/99problems.md
```diff
@@ -2126,10 +2126,79 @@ depth-first order).
...
+        let t, {d; f; pred; color; acc} =
+          let t = t + 1 in
+          List.fold_left edge
+            (t, {d=Char_map.add u t d; f;
+                 pred; color=Char_map.add u Gray color; acc})
+            (Char_map.find u g)
+        in
+        let t = t + 1 in
+        t , {d; f=(Char_map.add u t f); pred; 
+             color=Char_map.add u Black color; acc = fn acc (t, u)}
```
shayne-fletcher: On Sun, Jun 8, 2014 at 6:42 AM, Christophe Troestler <
notifications@github.com> wrote:

> I appreciate you pass t but I am wondering whether it is the right value
> of t to pass. I think it would be better that t is the distance to the
> initial node (something that is easy to describe in the statement).

​Or perhaps, go back to not passing it for simplicity in the problem
statement? I agree that the finishing time would not be helpful on its own.

## 

Shayne Fletcher

Chris00: Yep, I think this is better (not passing `t`).

Chris00:
On Sat, 7 Jun 2014 16:08:42 -0700, Shayne Fletcher wrote:

> With the revised implementation, taking 'w' as the source vertex, the top-level
> val l : M.node list = ['z'; 'v'; 'x'; 'y'; 'w']

Shouldn't the source vertex be first?  Seems to me the calls are done in reverse (one should call `f` on the nodes closer to 'w' first).

shayne-fletcher:
On Sun, Jun 8, 2014 at 7:07 AM, Christophe Troestler <
notifications@github.com> wrote:

> Shouldn't the source vertex be first? Seems to me the calls are done in
> reverse (one should call `f` on the nodes closer to 'w' first).

​Should they? My understanding has been that the call should occur not at
discovery but at the point the node has been fully explored, i.e.
'finished'. This accounts for the ordering observed. What do you think -
should this behavior be changed?

## 

Shayne Fletcher

Chris00:
On Sun, 8 Jun 2014 04:10:49 -0700, Shayne Fletcher wrote:

> On Sun, Jun 8, 2014 at 7:07 AM, Christophe Troestler <
> notifications@github.com> wrote:
> 
> > Shouldn't the source vertex be first? Seems to me the calls are done in
> > reverse (one should call `f` on the nodes closer to 'w' first).
> 
> ​Should they? My understanding has been that the call should occur not at
> discovery but at the point the node has been fully explored, i.e.
> 'finished'. This accounts for the ordering observed. What do you think -
> should this behavior be changed?

If I refer to https://en.wikipedia.org/wiki/Depth-first_search#Example (I intend to point the statement to that page), this is how I interpret « a depth-first search starting at A [...] will visit the nodes in the following order: A, B, D, F, E, C, G ».  Does the book you refer to says otherwise?

shayne-fletcher:
On Sun, Jun 8, 2014 at 7:15 AM, Christophe Troestler <
notifications@github.com> wrote:

> Does the book you refer to says otherwise?

I don't think it's explicitly stated in my reference (Cormen et. al). I
think I'm with you in thinking that the ordering should start with the root
node of the tree. I'll adjust the code.

## 

Shayne Fletcher

shayne-fletcher:
Invoking `fn` on discovery (instead of finishing as before) results in output `['w'; 'z'; 'y'; 'x'; 'v']`.

Chris00:
OK, one final question: are you ready to write some more text (possibly later) at the end explaining/showing how to use the fields of the state and the gray color for useful purposes?  I would indeed like that the reader to understand the usefulness of keeping track of all these elements.

shayne-fletcher:
On Sun, Jun 8, 2014 at 7:59 AM, Christophe Troestler <
notifications@github.com> wrote:

> OK, one final question: are you ready to write some more text (possibly
> later) at the end explaining/showing how to use the fields of the state and
> the gray color for useful purposes? I would indeed like that the reader to
> understand the usefulness of keeping track of all these elements.

​Not a problem. I'll enjoy doing that Christophe​!

## 

Shayne Fletcher

shayne-fletcher:
Do we proceed as is and work on the narrative independently or, are we waiting on that text before going forward?

Chris00:
Please submit another pull request for the narrative.

shayne-fletcher:
On Sun, Jun 8, 2014 at 11:26 AM, Christophe Troestler <
notifications@github.com> wrote:

> Please submit another pull request for the narrative.

​Yes! Hey, Christophe? Thanks for your patience. This has been a very
productive learning experience for me and I thank you for it. Perhaps I
will see you later this year at ICFP (I'll buy you a beer ;) )? In any
case, stay tuned - I'll be back shortly with the first draft of the
narrative.​

## 

Shayne Fletcher

Chris00:
On Sun, 8 Jun 2014 08:49:50 -0700, Shayne Fletcher wrote:

> Perhaps I will see you later this year at ICFP (I'll buy you a beer ;) )?

I'd like to go.  Not sure yet I will be able to.

avsm:
> Perhaps I will see you later this year at ICFP (I'll buy you a beer ;) )

I do a passable impression of @chris00 if it'll get me a free beer :-)

shayne-fletcher:
On Sun, Jun 8, 2014 at 2:20 PM, Anil Madhavapeddy notifications@github.com
wrote:

> I do a passable impression of @chris00 https://github.com/chris00 if
> it'll get me a free beer :-)
> 
> ​Anil old boy, you'll get one from me anyway :)

## 

Shayne Fletcher

Chris00:
Yes, I think that could be useful for the readers.  Please submit a PR.

PR #512, opened 2014-08-20
opam.ocaml.org integration

dsheets:
This should prepare ocaml.org to generate a template for opam.ocaml.org and host the necessary assets. No user-visible changes should occur after application.

Please review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,12 @@ syncotherfiles:
 	  --exclude '*.cmi' --exclude '*.cmo' --exclude '*.annot' \
 	  -rltHprogv site/* ocaml.org/
 
+# This target is PHONY (see end) because we don't want to enumerate
+# all the dependents that invalidate the resulting
+# opam.ocaml.org/template.xhtml for fear of that enumeration being
+# incomplete or falling out of date. That is, we treat ocaml.org's
+# page generation process as a black box here and suffer no noticeable
+# performance degradation.
 opam.ocaml.org_template:
 	$(MAKE) -f Makefile.opam.ocaml.org
 
--- a/Makefile
+++ b/Makefile
@@ -19,21 +19,8 @@ syncotherfiles:
 	  --exclude '*.cmi' --exclude '*.cmo' --exclude '*.annot' \
 	  -rltHprogv site/* ocaml.org/
 
-OPAM_OCAML_ORG_TITLE=<t:insert xmlns:t='http://ocaml.org/xmlns/template\#' name='title'/>
-opam.ocaml.org-template: script/absolute_urls
-	$(MPP) -set "notranslations" \
-	       -set "title=$(OPAM_OCAML_ORG_TITLE)" \
-               -set "filename=opam.ocaml.org/packages" \
-	       -set "custom-head=opam.ocaml.org/head.xhtml" \
-	       -set "custom-edit=opam.ocaml.org/edit.xhtml" \
-	       -set "custom-search=opam.ocaml.org/search.xhtml" \
-	       -set "custom-contact=opam.ocaml.org/contact.xhtml" \
-	       -set "nobreadcrumb" \
-	       -set "packages" \
-	       -set "page=opam.ocaml.org/page.xhtml" \
-	       template/main.mpp \
-	       -o opam.ocaml.org/template.xhtml
-	script/absolute_urls --path opam.ocaml.org opam.ocaml.org/template.xhtml
+opam.ocaml.org_template:
+	$(MAKE) -f Makefile.opam.ocaml.org
 
 # List of files to build, sorted according to their source format
 MD_FILES = $(patsubst site/%, ocaml.org/%, $(patsubst %.md, %.html, \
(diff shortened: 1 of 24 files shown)


template/main.mpp
```diff
@@ -29,9 +31,10 @@
     <link rel="alternate" type="application/rss+xml" title="RSS"
 	  href="http://planet.ocaml.org/rss20.xml" />
     !}}
+    {{!c ifdef custom-head {{!i input ((! get custom-head !)) i!}} c!}}
     <script>
-      if (document.implementation.hasFeature("http://www.w3.org/TR/SVG11/feature#Image", "1.1")) {
-      document.documentElement.className = "svg";
+      if (document.implementation.hasFeature('http://www.w3.org/TR/SVG11/feature#Image', '1.1')) {
```
Chris00, line 36: Both type of quotes are possible in JS, no?

dsheets, line 36: Yes, they are, but opam2web uses cow uses xmlm which assumes it's talking to an XML parser which would treat the content of `<script>` as PCDATA rather than CDATA which means that character entities are not parsed and all kinds of nasty junk can be included verbatim in the body of a `<script>`. The HTML `<script>` parsing rules are gnarly and the New Web Order (WHATWG) has decided in their glue-sniffing wisdom to **not** provide a way to treat the content of `<script>` like the content of any other tag if you want to ship `<script>` with character entities embedded...

This is all a long way to say that xmlm outputs `&quot;` in PCDATA (everything) so that it doesn't have to distinguish the quoting context (it uses `"` for attribute quoting and so does not want to inject new `"` characters).

If we just avoid `<`, `>`, `&`, and `"` in embedded `<script>` and `<style>` for now, we'll be ok...

Chris00, line 36: OK.

Chris00, line 36: Thanks for the explanation!

Makefile
```diff
@@ -19,6 +19,9 @@ syncotherfiles:
 	  --exclude '*.cmi' --exclude '*.cmo' --exclude '*.annot' \
 	  -rltHprogv site/* ocaml.org/
 
+opam.ocaml.org_template:
```
Chris00, line 28: May you tell why you have a phony target if you actually want to generate the file `opam.ocaml.org/template.xhtml`.

dsheets, line 28: I didn't want to list all the dependencies/didn't look up how they are tracked through mpp files. The generation happens so quickly, I don't think it's a problem to re-run it even when everything is up-to-date.

Chris00, line 28: On Wed, 20 Aug 2014 15:39:29 -0700, David Sheets wrote:

> I didn't want to list all the dependencies/didn't look up how they are tracked through mpp files. The generation happens so quickly, I don't think it's a problem to re-run it even when everything is up-to-date.

OK.  I am not too concerned about the speed but rather your motivation to do it this way.  Maybe a short comment could help other readers?

dsheets, line 28: 6f763eb18a8c16c96069f22aa34ab6f377481382

Chris00, line 28: On Wed, 20 Aug 2014 16:26:06 -0700, David Sheets wrote:

> In Makefile:
> 
> > @@ -19,6 +19,9 @@ syncotherfiles:
> >         --exclude '_.cmi' --exclude '_.cmo' --exclude '_.annot' \
> >         -rltHprogv site/_ ocaml.org/
> > 
> > +opam.ocaml.org_template:
> 
> 6f763eb

Thanks.

opam.ocaml.org/contact.xhtml
```diff
@@ -0,0 +1,13 @@
+<div class="column">
+  <div class="entry">
+    <h1>Contact</h1>
```
Chris00, line 3: I am a bit concerned that people browsing the site will expect the same links to have the same effect (not noticing the URL difference, just relying on the visual layout).  Maybe changing the title may improve that?

opam.ocaml.org/edit.xhtml
```diff
@@ -0,0 +1 @@
+<t:seq xmlns:t="http://ocaml.org/xmlns/template#" />
```
Chris00, line 1: Since you already use a XML template engine for opam.ocaml.org, I am wondering whether — for the medium term — we should not make it a separate packaga and use it for both sites.

opam.ocaml.org/contact.xhtml
amirmc, line 3: David mentioned this to me too. I think we should open a new issue to track this particular point (and deal with it later). 

Best wishes,
Amir

## 

sent via mobile

Chris00:
The changes look good to me.  If nobody else comment, I'll merge them at the end of the day (10PM UTC+2).

opam.ocaml.org/contact.xhtml
dsheets, line 3: This fragment is currently unused as `custom-contact` is not passed in `Makefile.opam.ocaml.org`.

opam.ocaml.org/edit.xhtml
dsheets, line 1: I absolutely agree but I was overridden for the short term to deliver features rather than decouple `opam2web`. There are at least 3 (more) packages that I want to pull out of `opam2web` and I think we could get quite a bit of benefit from reuse by ocaml.org.

dsheets:
OK to merge?

Chris00:
Sorry, I am back to my normal duties which means that I will regularly have to prioritize things at the expense of my OCaml work...

PR #517, opened 2014-08-24
add abstracts to program

garrigue:


Chris00:
We are not so keen to put PDF files in this repository.  Could you use https://github.com/ocaml/ocaml.org-media for those?  Thanks.

garrigue:
On 2014/08/24 21:01, Christophe Troestler wrote:

> We are not so keen to put PDF files in this repository. Could you use https://github.com/ocaml/ocaml.org-media for those? Thanks.
> 
> —
> Reply to this email directly or view it on GitHub.

Fine, what should I do ?
ocaml.org-media looks empty.
Where should I put my files ?
Maybe it's simpler if I host them elsewhere…

```
Jacques
```

garrigue:
By the way, for last year's meeting, the pdf's are in the main repository.

```
Jacques
```

Chris00:
On Sun, 24 Aug 2014 05:19:48 -0700, garrigue wrote:

> By the way, for last year's meeting, the pdf's are in the main repository.

Yes, I know, but we would prefer not to repeat it!

Chris00:
On Sun, 24 Aug 2014 05:16:58 -0700, garrigue wrote:

> On 2014/08/24 21:01, Christophe Troestler wrote:
> 
> > We are not so keen to put PDF files in this repository. Could you use https:/
> > /github.com/ocaml/ocaml.org-media for those? Thanks.
> > 
> > Fine, what should I do ?
> > ocaml.org-media looks empty.
> > Where should I put my files ?

It seems to me that the simpler is to create a directory OUD2014 and put your files inside.

garrigue:
On 2014/08/24 21:35, Christophe Troestler wrote:

> On Sun, 24 Aug 2014 05:16:58 -0700, garrigue wrote:
> 
> > On 2014/08/24 21:01, Christophe Troestler wrote:
> > 
> > > We are not so keen to put PDF files in this repository. Could you use https:/
> > > /github.com/ocaml/ocaml.org-media for those? Thanks.
> > > 
> > > Fine, what should I do ?
> > > ocaml.org-media looks empty.
> > > Where should I put my files ?
> 
> It seems to me that the simpler is to create a directory OUD2014 and put your files inside.

But then, how do I refer to the files I put there?
And how can I test that the links are correct, before you merge the changes?

```
Jacques
```

garrigue:
On 2014/08/24 21:44, Jacques Garrigue wrote:

> On 2014/08/24 21:35, Christophe Troestler wrote:
> 
> > On Sun, 24 Aug 2014 05:16:58 -0700, garrigue wrote:
> > 
> > > On 2014/08/24 21:01, Christophe Troestler wrote:
> > > 
> > > > We are not so keen to put PDF files in this repository. Could you use https:/
> > > > /github.com/ocaml/ocaml.org-media for those? Thanks.
> > > > 
> > > > Fine, what should I do ?
> > > > ocaml.org-media looks empty.
> > > > Where should I put my files ?
> > 
> > It seems to me that the simpler is to create a directory OUD2014 and put your files inside.
> 
> But then, how do I refer to the files I put there?
> And how can I test that the links are correct, before you merge the changes?

OK, if you really want me to use ocaml.org-media, I think the right way to do it is
to have the Makefile of ocaml.org handle it.
I.e., ocaml.org-media, should have exactly the same directory structure as ocaml.org,
and it would be automatically copied to the target.

This could be done by changing the target syncotherfiles, in ocaml.org/Makefile:

syncotherfiles:
    rsync --exclude '_~' --exclude '_.md' --exclude '_.html' \
      --exclude '_.cmi' --exclude '_.cmo' --exclude '_.annot' \
      -rltHprogv site/\* ocaml.org/
    if test -d ../ocaml.org-media; then \
     rsync --exclude '_~' --exclude '_.md' --exclude '_.html' \
      --exclude '_.cmi' --exclude '_.cmo' --exclude '_.annot' \
      -rltHprogv ../ocaml.org-media/\* ocaml.org/ ; fi

This would allow me to test the site before doing a pull-request.

```
Jacques
```

garrigue:
I moved the pdf files to ocaml.org-media.
Note that now the target syncotherfiles also rsyncs files from ../ocaml.org-media.
This supposes that the repository for ocaml.org-media is checked out in a directory
next to ocaml.org.
Also, you need to merge my pull request on ocaml.org-media simultaneously.

Is it ok this time?

OCaml 2014 is 12 days from now, and I'll probably have to update files repeatedly...

Chris00:
1. You need to rework your commits so that the PDF files are not (never) checked in your repo. (deleting them is not enough, maybe smashing your commits will do).
2. I think you can serve the PDF files directly from github (e.g. https://github.com/ocaml/ocaml.org-media/blob/master/meetings/ocaml/2014/ocaml2014_1.pdf?raw=true).  Another solution may be desirable for the longer term but we would prefer that contributors need not to check out ocaml.org-media — which may become bulky — to compile the site.

garrigue:
On 2014/08/25 00:00, Christophe Troestler wrote:

>   • You need to rework your commits so that the PDF files are not (never) checked in your repo. (deleting them is not enough, maybe smashing your commits will do).

Sorry, I don't know how to do that.

>   • I think you can serve the PDF files directly from github (e.g. https://github.com/ocaml/ocaml.org-media/blob/master/meetings/ocaml/2014/ocaml2014_1.pdf?raw=true). Another solution may be desirable for the longer term but we would prefer that contributors need not to check out ocaml.org-media — which may become bulky — to compile the site.

They don't need to: this is the reason I added an if.
If they don't checkout the media, they will have some local broken links,
but this is already the case since local files do not behave exactly like served ones anyway.
For me, this is essential: otherwise I cannot check locally before I create a push request.

```
Jacques
```

garrigue:
It seems that I could squash the commits by roaming the web.
Is it ok this time?
Otherwise I'll move my files to somewhere else, because using ocaml.org-media without the overlay would be too much of a pain.

agarwal:
The solution you came to is correct. The ocaml.org-media repo is meant to be copied as-is to the root of the site (except perhaps the README.md, which helps to explain the purpose of the repo). Thanks for the patch, so we can actually start using it.

agarwal:
@garrigue Did you intend to have two copies of the program, in program.md and index.md? Or should we now delete program.md?

lefessan:
I don't know how the site is synchronized, but currently, all the links appear but the files themselves are not available (404 Not Found).

agarwal:
Ah yes, of course. We have to update the server to sync the ocaml.org-media repo. I'll try to figure that out.

agarwal:
I did a manual copy, so the links should work now. Issue #520 requests this to be fixed correctly.

garrigue:
Thank you, everything works fine now.

For the duplicated program, at first I was thinking of removing the separate page, but then I remembered the basic principle of the web: stability of links. I'll come up with a better solution later...

PR #525, opened 2014-08-28
Fix links to CUFP slides (Verlaguet)

smondet:
Right now from there: http://ocaml.org/learn/ we are lead to a 404 for Facebook's CUFP slides.


agarwal:
Thanks!

PR #536, opened 2014-09-23
Improve the install page, fix some typos.

dmbaturin:
Improved the install page a little, fixed some typos and odd (as of me) wording.
The XCode plugin section that says there's no native compiler for Intel makes me think it needs complete rewrite, but I'm not using either OS X or said plugin so I can't really say anything here.

### DIFF
--- a/site/docs/install.md
+++ b/site/docs/install.md
@@ -11,11 +11,9 @@ ways. Broadly, the options are:
   Mac OS X, ...)
 * install from source code.
 
-These different options are detailed below and explain how to install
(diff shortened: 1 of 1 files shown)


agarwal:
Thanks. This page needs a lot of work. This helps.

dmbaturin:
I do have OS X, even though I don't use it other than for interoperability testing, so I will try the OS X instructions and XCode plugin later and update those sections if needed.

agarwal:
That would be great. Thanks!

PR #540, opened 2014-09-28
opam.ocaml.org: 57/99 problem code

RjiukYagami:
i've done cone to 57. problem from 99 problem list.

https://github.com/emil3566/99-ocamls-problem/blob/master/src/50-59/57.Layout%20a%20binary%20tree%20%283%29.ml

It's commented and i allow creators of procjet to change it on their needs. I'm just putting it here becuase it might be helpfull, cuase there is no code for this task. It uses Core.Std so it should be said somewhere :) . If u find it usefull, feel free to use it, or copy.


Chris00:
Please rebase your changes on `master` (the `syndic` branch is not the branch to work against for this kind of changes).  You pull request must **only** contain your changes (which can't be seen above BTW).

PR #541, opened 2014-09-28
opam.ocaml.org: 57/99 problem code

RjiukYagami:
I've made code to 57. problem of 99 ocaml's problems. Hope you like it :) . It uses Core.Std it's main idea is to build tree without looking at position and then if we have colision we stretch subtrees of this nodes in their lca ( lowest common ancestor ) :yellow_heart:

### DIFF
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -1763,7 +1763,7 @@ SOLUTION
 >      | Node(j, x, y) ->
 >         let (val1, val2) = List.Assoc.find_exn rev_done_tree j in
 >         N(j, (val1+add_to_all), val2 , (build_answer x), (build_answer y));;
->...
+>```
 
 ```ocamltop
 build_answer example_layout_tree;;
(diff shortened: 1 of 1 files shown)


site/learn/tutorials/99problems.md
```diff
@@ -1636,10 +1636,140 @@ nodes. How tight can you pack together two subtrees to construct the
 combined binary tree? This is a difficult problem. Don't give up too
 early!
 
-<!-- SOLUTION -->
+SOLUTION
 
-```ocaml
-  (* solution pending *)
+>```ocamltop
+>   open Core.Std;;
```
Chris00, line 1642: Please do not use `Core`, we prefer the solutions to require only the sandard distribution with no additional package.  In your case, that only amounts to swap some arguments and/or remove labels.

Chris00:
Simpler solution in https://github.com/ocaml/ocaml.org/commit/26a189c4ef424c1e1510cb453a923f02860317ab

PR #547, opened 2014-10-09
Improve organization in install.html

lefessan:
- put all Linux distributions under a first-level item
- put all package managers under a first-level item
- move FreeBSD at the end, as items should be sorted by popularity

Conflicts:
    site/docs/install.md

### DIFF
--- a/site/docs/install.md
+++ b/site/docs/install.md
@@ -15,7 +15,12 @@ The following sections explain how to use each of these methods.
 For additional information about finding and installing OCaml libraries,
 see the [libraries](/learn/libraries.html) page.
 
-## OPAM
+## Multi-Platform Package Managers
+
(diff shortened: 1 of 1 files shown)


avsm:
I have no particular objection about the ordering, but it's worth noting that FreeBSD is almost certainly more popular than some of the more obscure Linux distributions that we list in this page, and definitely more widely deployed.

lefessan:
I completely agree with that, FreeBSD is also more popular than most of the package managers in the first section. My point was more to have Linux < Windows < Mac OS X < FreeBSD.

avsm:
That sounds good.  @madroach has been importing OpenBSD/OCaml ports at a high rate recently, so we can probably break out FreeBSD/OpenBSD/NetBSD into their own section in the future.

PR #559, opened 2014-10-24
Clean ocamlbuild pages

dsheets:
Fix the formatting of the _Making plugins_ page, add titles to pages, remove a couple of junk pages. Please review.


pw374:
About 
site/learn/tutorials/ocamlbuild/Making_plugins.md
I think the things revealing a bug in OMD would be better in backquotes, instead of backslash-escaping some square brackets.

Philippe Wang
(mail sent from a phone)

dsheets:
Good idea. Updated.

pw374:
If it builds as expected, I think it should be merged now. :)

(I'm away from my laptop, so I can't build this now.)

Philippe Wang
(mail sent from a phone)

PR #577, opened 2014-11-14
Add McGill University to the list

chenglou:
(Editor auto stripped some trailing whitespaces.)

Using a bit.ly link because the `~` in the URL gets transformed into `%7E` and the mcgill website fails to parse that, lol.

I can't modify the google maps for some reason?

### DIFF
--- a/site/learn/teaching-ocaml.md
+++ b/site/learn/teaching-ocaml.md
@@ -31,7 +31,7 @@ OCaml a slightly easier task.
   [create an issue](https://github.com/ocaml/ocaml.org/issues/new).
 
 ### Mailing list
-We set-up a [mailing list](http://lists.ocaml.org/listinfo/teaching) for professors who are teaching OCaml in schools and universities to discuss issues relating to using OCaml in an educational context. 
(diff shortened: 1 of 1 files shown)


pw374:
The [Markdown parser we use](https://github.com/ocaml/omd) normally won't transform the `~` into a `%7E`, so I think you should use the original link instead. Also, it seems you may make the link shorter by removing the trailing `index.php`.

The webmaster of http://www.cs.mcgill.ca/ should be told that `%7E` is wrongly interpreted... :-/

chenglou:
Updated.
(How do I update the google maps pin?)

pw374:
Thanks!

About the Google Maps pin, I've no idea. But @yansh or @amirmc might know?

yansh:
On 14/11/14 13:43, Philippe Wang wrote:

> About the Google Maps pin, I've no idea. But @yansh 
> https://github.com/yansh or @amirmc https://github.com/amirmc 
> might know?
> 
> @chenglou Yes, you can definitely edit the google map, follow the link

https://www.google.com/maps/d/edit?mid=zk8_K4G_usic.kkzYvEvqV44Q

There is a link on the page.

chenglou:
Maybe it's the permission setup? I don't see the link
screenshot 2014-11-14 09 38 57

yansh:
@chenglou yep, you're right. should work now. Sorry and thanks pointing it out.

PR #585, opened 2014-11-15
Update compiling_with_gnu_make.md

gasche:
It looks like Markus Mottl has moved away from bitbucket. The [old link](https://bitbucket.org/mmottl/ocaml-makefile) goes to a redirection page that suggest [this page](http://mmottl.github.io/ocaml-makefile/) I used in this pull-request. One might prefer to directly link to the source repository, but this page's documentation is interesting and seems fairly complete -- one may wonder whether it could even replace the current `compiling_with_gnu_make` page.

### DIFF
--- a/site/learn/tutorials/compiling_with_gnu_make.md
+++ b/site/learn/tutorials/compiling_with_gnu_make.md
@@ -5,7 +5,7 @@
 *Table of contents*
 
 ## Using GNU make with OCamlMakefile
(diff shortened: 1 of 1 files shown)


pw374:
It seems to me that it's fine to link to http://mmottl.github.io/ocaml-makefile/, especially since there's a very visible link to the repo on github!

pw374:
Thanks!

PR #595, opened 2014-11-27
Update teaching-ocaml.md

grayswandyr:
Added course at ISAE/Supaéro.


amirmc:
Thank you!  Is this also on the [map](https://www.google.com/maps/d/edit?mid=zk8_K4G_usic.kkzYvEvqV44Q)?

amirmc:
I'll add it to the map now.

grayswandyr:
Le 27/11/2014 14:01, Amir Chaudhry a écrit :

> I'll add it to the map now.

Thanks, I would have needed a Google account I guess. Too bad 
OpenStreetMap wasn't used.

best

david

amirmc:
I'm sure we'd be perfectly happy to switch to OSM maps but some of us are not as familiar with it.  It might even be easier to automate with the appropriate libraries.  

I've opened a new issue, #596, to track this. Would be great to have more contributors and this seems like a useful, self contained item.

PR #597, opened 2014-11-28
updated profiling on os x

nilsbecker:
Shark is long gone. Instruments works but does not give any useful function names, so profiling on OS X is essentially broken.


dbuenzli:
I wrote that a long time ago. I don't think it's worth speaking about the past here, it's just noise at that point. I'd kill the shark mentions.

nilsbecker:
ok.

agarwal:
I meant to merge, then close. Done now.

PR #603, opened 2014-12-29
Fixed path to the HTML tarball

Teemperor:
In the book ( https://ocaml.org/learn/books.html#RealWorldOCaml ) section, the link to the HTML tarball of "The OCaml System: Documentation and User's Manual" has a typo which results in 404. This fixes the issue.


pw374:
Thanks!

PR #604, opened 2014-12-30
Fixed dead links.

Teemperor:
I crawled the page for dead links and tried to fix as many of the broken links as possible. I used the new location for moved content and archive.org for content that totally disappeared from the internet.

Notes:
- The diode_graph.gif was already replaced in the english version, so i did the same to the other languages who still linked to that missing image.
- The 3.12 release has no public MSVC version, so i completly removed the paragraph about that version from the release page.
- The svn checkout example also resulted to a 404 error due to the '/dir' suffix, so i removed that.


Teemperor:
These are btw the remaining dead links on the site: http://pastebin.com/BrTnx1y8

pw374:
Thanks!!

agarwal:
I've opened #607.

PR #606, opened 2014-12-31
Fixed more dead links.

Teemperor:
Notes (from top to bottom of the unified diff):

rss2html.ml: Removed space at the end of the link (it works, but it produces HTML warnings with that space).

planet/index.md: Planet has moved into the "planet/"-subdirectory, so all links to images need to one directory higher to be valid.

docs/index.[fr.]md: https://opam.ocaml.org/doc/Quick_Install.html => https://opam.ocaml.org/doc/Install.html

docs/install.md: mmottl moved to github


agarwal:
Thanks for all these fixes!

PR #610, opened 2015-01-18
Fix a bug in p65 of 99 problems

shaform:
Hi,
I think there is an error in this solution. For example, it cannot handle this tree correctly:

``` ocaml
let htree =
    let leaf x = Node (x,Empty,Empty) in
    Node('n', Empty,
         Node('u', Node('p', Empty, leaf 'q'), Empty));;
```

The correct answer should be:

``` ocaml
N (n, 1, 1, E, N (u, 5, 2, N (p, 3, 3, E, N (q, 4, 4, E, E)), E))
```

But the original solution gives:

``` ocaml
N (n, 5, 1, E, N (u, 9, 2, N (p, 7, 3, E, N (q, 8, 4, E, E)), E))
```


Chris00:
Fixed — 24cbb22 (made the modification to `translate_dst` as it seemed more appropriate).

PR #612, opened 2015-01-20
Fix 'explicitely' typo.

michaelsproul:
I was reading the pointer guide and noticed "explicitely" in a few places instead of "explicitly", here's a patch that removes all occurrences of the former :smile:


Chris00:
Thanks.

PR #614, opened 2015-01-20
Removed - "Install the binary package from Inria;"

yansh:
Inria doesn't provide MacOS binary packages anymore. See http://caml.inria.fr/download.en.html

### DIFF
--- a/site/docs/install.md
+++ b/site/docs/install.md
@@ -295,7 +295,7 @@ been added to OCaml for better support of Windows.
 
 ## Mac OS X
 On Mac OS X, there are, at least for the base package, a few different
-ways to go: Use OPAM; Install the binary package from Inria; Install via Fink;
+ways to go: Use OPAM; Install via Fink;
 Install via MacPorts; or build it manually from
 sources. For each of these approaches, you will need to have at least
 the OS X developer tools installed — any other requirements will be


site/docs/install.md
```diff
@@ -295,7 +295,7 @@ been added to OCaml for better support of Windows.
 
 ## Mac OS X
 On Mac OS X, there are, at least for the base package, a few different
-ways to go: Use OPAM; Install the binary package from Inria; Install via Fink;
```
Chris00, line 298: If it is removed, then the next section does not make sense anymore.  Since the Inria package was the natural way to go, some recommendations would be welcome for newcomers — do all installation method work equally well and are they all up to date?

yansh, line 298: The PR just removes the recommendation to use Inria package.The rest is still valid. Any MacOS experts can confirm that?

agarwal, line 298: Homebrew is the most well maintained binary package on Mac OS X at this time.

yansh, line 298: Yes, that’s also what worked for me.

Chris00, line 298: May I suggest that this paragraph be rewritten taking into account these experiences and possibly that the subsequent sections be reordered accordingly?

yansh, line 298: > On Jan 20, 2015, at 12:45 PM, Christophe Troestler notifications@github.com wrote:
> 
> In site/docs/install.md https://github.com/ocaml/ocaml.org/pull/614#discussion-diff-23239855:
> 
> > @@ -295,7 +295,7 @@ been added to OCaml for better support of Windows.
> > 
> >  ## Mac OS X
> >  On Mac OS X, there are, at least for the base package, a few different
> > -ways to go: Use OPAM; Install the binary package from Inria; Install via Fink;
> > May I suggest that this paragraph be rewritten taking into account these experiences and possibly that the subsequent sections be reordered accordingly?

Not sure what else there is to add. I would definitely put homebrew first though.

@amirmc any thoughts?

amirmc, line 298: I'd completely agree that homebrew is the way to do things on a Mac, especially since it also helps with managing the external dependencies that some opam packages need.

@yansh since you've recently gone through this on a new machine perhaps you can write up notes, re-order and add to the PR?

yansh, line 298: Done. Let me know what you guys think.

Chris00:
Superseded by #620

PR #615, opened 2015-01-20
Removed the paragraph about using Inria packages

yansh:
Inria doesn't provide MacOS binary packages anymore. See http://caml.inria.fr/download.en.html

As a result, there is no image for OCaml 4.02.X in 
https://ocaml.org/releases/


Chris00:
Superseded by #620

PR #616, opened 2015-01-22
Add an opam file for pinning. Fixes #609.

### LINKED ISSUE #609: create local opam file
It's been a long time since I've actually been able to build the site locally. Why don't we create a local opam file to force us to state the dependencies clearly. We can then use that for Travis as requested in #241.

In the mean time, it will be great if anyone can state what is required. Is there a specific version of omd, ocamlrss, opamfu, etc. needed? I'm trying to work through it, but still stuck.

dsheets:
Depends on ocaml/opam-repository#3439.

### DIFF
--- a/opam
+++ b/opam
@@ -1,0 +1,24 @@
+opam-version: "1.2"
+name: "ocamldotorg"
+version: "2.1.0"
+maintainer: "The OCaml.org team"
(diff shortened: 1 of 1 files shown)


dsheets:
opam2web 1.4.0 is now published. Please test and review.

PR #628, opened 2015-02-23
Gtk tutorial updates

dmbaturin:
Example program used to segfault because of missing GtkMain.Main.init call.
Removed references to obsolete Gtk 1.2. Examples seem to work well in Gtk2.
Replaced the old lablgtk link with an OCaml Forge link, the original website is obsolete it seems.


Chris00:
Thanks.

PR #646, opened 2015-04-11
Chinese tutorial on half way

HuStmpHrrr:
translated almost 50% of them. it's more than i think. will do the rest later on.


Chris00:
Thanks.

PR #652, opened 2015-05-11
Fix a typo regarding "transparant"

pinkfloyda:


pw374:
Should we use "opaque" instead? Or perhaps "seamless"?

Chris00:
If "transparent" is not clear enough (the word "transparent" is meant, not its negation), maybe we should reformulate instead.  Maybe something like "but the programmer doesn't have to bother with that..."?  The help of a native English speaker would be nice. @agarwal @avsm ?

pw374:
"invisible"?

dbuenzli:
hidden

pw374:
Wow, I've just read the entire page, it is super Frenchy...

pinkfloyda:
I personally like 'hidden', 
we can reformulate it to be
"but it is hidden from the programmer since pointer
 handling is done by the compiler"

agarwal:
I would say "this is not visible to the programmer", but I think "transparent" is fine too.

pw374:
Oxford dictionary says:

> **transparent**
> adjective
> 1. (of a material or article) allowing light to pass through so that objects behind can be distinctly seen: _transparent blue water | fine transparent fabrics._
> 2. easy to perceive or detect: _the residents will see through any transparent attempt to buy their votes | the meaning of the poem is by no means transparent._
> • having thoughts or feelings that are easily perceived; open: _you'd be no good at poker—you're too transparent._
> • (of an organization or its activities) open to public scrutiny: _if you had transparent government procurement, corruption would go away._
> 3. (Computing) (of a process or interface) functioning without the user being aware of its presence.
> 4. (Physics) transmitting heat or other radiation without distortion. _CFCs and water vapour are virtually transparent to incoming short-wave solar radiation._

The use of "transparent" is correct here. However for who's not aware that there's a definition specific to computing, it is quite confusing: just try to apply the other definitions!

pw374:
I'm closing this PR. Don't hesitate to open new one if need be.

dbuenzli:
I really think "hidden" should have been used. It seems consistent to me with our usage of hiding in the context of the OCaml module system --- hiding type representations.

pw374:
I agree that "hidden" is a good candidate.

Could you make a PR please?
(This one won't be merged anyway, since it's wrong.)

Issue #554, opened 2014-10-22
Add More OCaml to Books page


amirmc:
http://ocaml-book.com/more-ocaml-algorithms-methods-diversions/

agarwal:
Related to #524.

dbuenzli:
I'm currently going through the book and (somehow as expected) it's an excellent one. It's a shame it's still not mentioned on the website. @johnwhitington I don't think anyone will blame you if you take action to integrate it yourself.

johnwhitington:
I've had a go at getting ocaml.org to build on my machine so I can begin to make contributions, but no luck yet (stuck on getting ocamlnet to build for now).

Perhaps the root of the problem is that something in "make deps" seems to rely on OCaml being exactly version 4.01.0, so everything I'm doing may be out of date. Can that requirement really still be needed?

```
=-=- Synchronising pinned packages =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=  🐫 
[ocamlorg] /Users/john/repos/ocaml.org/ already up-to-date
[ERROR] ocamlorg is not available because your system doesn't comply with
        ocaml-version = "4.01.0".
```

dbuenzli:
@johnwhitington You can keep this simple by simply using the pencil on the top right corner visible on any page of the website. This will lead you to github where you can simply edit the `.md` file by clicking again the pencil. I guess in this case you will maybe need to make a separate commit for adding the image file to the `img` directory.

johnwhitington:
See Pull Request #657

dsheets:
Fixed by #657.

Issue #524, opened 2014-08-27
Add *More OCaml* book to front docs page?


dsheets:
Also, I think there was another issue somewhere about some recent OCaml book in French...

Chris00:
Yes, there is space to add 2 more.  

For the record: http://www.amazon.com/More-OCaml-Algorithms-Methods-Diversions/dp/0957671113/
and http://ocaml-book.com/more-ocaml-algorithms-methods-diversions/

Chris00:
French book: https://github.com/ocaml/ocaml.org/issues/445

johnwhitington:
I have prepared a cover image for "More OCaml" at 300x376, which is the same size as the one you're using for "OCaml from the Very Beginning" on the books & documentation page:

http://www.coherentpdf.com/more-ocaml-300-376.png

PR #657, opened 2015-05-23
Adds "More Ocaml" to learn/books

johnwhitington:
[I have refrained from adding it to docs/index.html. There seems to be room for two more books, at least on my browser. I suggest "More OCaml" and "Apprendre a programmer avec OCaml". I would be happy to prepare another pull request for that page, if people agree that there is space, and to provide a pull request to add "Apprendre a programmer" to learn/books too]

This pull request would close bug #554 and bug #524 if applied. "Apprendre a programmer...." is bug #534.


pw374:
thanks!

Chris00:
> I would be happy to prepare another pull request for that page,

Yes, please!  :+1:

PR #677, opened 2015-07-13
Updated links to Cornell and Illinois Universities

rizo:
See: https://github.com/rizo/awesome-ocaml/pull/47


amirmc:
Thank you!

PR #688, opened 2015-08-17
Add Ocsigen blog to the planet

Drup:


Drup:
Is there a delay ? It's still not shown [here](http://ocaml.org/community/planet/)

pw374:
There is a delay. But I don't think it's because of the delay.

(I'm not sure, but it seems to me that the oscigen feed is downloaded, but then, for some reason that isn't obvious to me, it gets dumped...)

pw374:
It could/might be because the feed is bad: https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Focsigen.github.io%2Ffeed.xml !!

Chris00:
Do not have the time to look at it now.  Please open an issue so we do not forget it.

Drup:
@pw374 Oups. It should be a well behaved feed now.

pw374:
I confirm that fixing the feed changed something: now its contents don't get dumped by `rss2html`. So, the posts should be included next time the feed page is regenerated... which could be about now+the time it takes to (re)build the site...

Edit: I confirm that the posts are now included. We just have to wait until the end of the current build process, which is imminent.

Drup:
It's in now! thanks.

Chris00:
@Drup May you post the old feed somewhere to have a testbed to improve error reporting?  Please post it [here](https://github.com/ocaml/ocaml.org/issues/689).

Drup:
The ocsigen blog feed is gone again! According to [this](https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Focsigen.github.io%2Ffeed.xml) the feed is good.

pw374:
Are you sure?

The post `Summer Ocsigen releases — Ocsigen blog (Gabriel Radanne), Aug 17, 2015`
does appear on http://ocaml.org/community/planet/ right now (or did, a few seconds ago)

Drup:
But not on http://planet.ocaml.org

pw374:
I wonder, isn't http://planet.ocaml.org supposed to be deprecated?

Drup:
well, it got the main url and I would guess most people that were using it before still do ...

pw374:
http://planet.ocaml.org is the "new" url of http://planet.ocamlcore.org, which was maintained by OCamlCore, which went out of business a few years ago...

I think we should redirect http://planet.ocaml.org to http://ocaml.org/community/planet/ instead!  --> ping @avsm @amirmc @Chris00 @gildor478

gildor478:
@pw374 even out of business, I still maintain some stuff on *.ocamlcore.org. ;-)

But it doesn't make sense to keep two resources and we should do the redirect.

Right now the only noticeable difference (for me) is that the OCamlCore Forge Project should point to: 
http://forge.ocamlcore.org/export/rss_sfprojects.php
(the other one is staled)

If someone, with edit right, can check that both lists are the same and tell me that everything is fine, I'll add a redirect from planet.{ocaml,ocamlcore}.org to http://ocaml.org/community/planet tonight.

amirmc:
Since this involves things under the domain name, we should mention/discuss it on the infrastructure list. 

There may well be people subscribed to the feed from planet.ocaml.org. Inserting a final post into just that feed to describe the change might be a good thing to do. 

Best wishes,
Amir

## 

Sent via mobile

gildor478:
My plan is to anyway do a post from my blog post or the admin account of forge.ocamlcore.org and it will should be displayed on planet.ocamlcore.org, planet.ocaml.org and ocaml.org/community/planet.

But I agree, that we should do the following:
1. Agree on a date for the migration (e.g. 2015-10-01)
2. I do a blog post that should appear on planet.o.o
3. do the migration on the given date

Feel free to start discussion on infrastructure ML and update governance document accordingly, I think this is a good move. Ping me on this bug when we will have reach an agreement on the date, so that I can publish a blog post.

Chris00:
http://ocaml.org/community/planet/ is indeed supposed to be the newer version of http://planet.ocaml.org/ — using only OCaml code.  An Atom feed is generated too: http://ocaml.org/feed.xml

PR #694, opened 2015-08-24
add some basic explanations about using perf

c-cube:
some useful bits of informations I gathered on #ocaml about how to use perf, and how to make it aware of the call graph (thanks to def-lkb). It seems like the right place to save this bit of information.


Drup:
Maybe add how to enable frame pointers ?

Also, it's not on older version of ocaml, it's on older version of _perf_.

gasche:
Ping? This pull-request looks very useful and mergeable.

agarwal:
Thanks for the ping. Merged.

PR #706, opened 2015-10-01
tutorial/Basics: advertise utop more prominently

gasche:
This proposed  change comes from the discussion in ocaml/ocaml#241 , where @jwatzman points out that `utop` does not have enough visibility in the "Basics" tutorial -- it is mentioned, but maybe it should be emphasized more.

It is hard to be informative yet concise enough: I realize this Basics tutorial is the first approach of OCaml content, so it should be to the point and should not drown the reader in pointless distinctions. Maybe the utop can be made more visible (than the previous mention) but more concise (than this proposed change).

We may need to differentiate better between a page that would be "a tour of the language" (let me show you why you should care, without necessarily having you run code) and "basic setup if you know you want to use". Currently the Basics tutorial tries to do both and it is non-optimal at them both.

For reference to the Install page, I used full URLs instead of relative navigation. Is there a preferred way?

### DIFF
--- a/site/learn/tutorials/basics.md
+++ b/site/learn/tutorials/basics.md
@@ -8,18 +8,31 @@
 
 The easiest way to get started is to run an interactive session in your browser, at <http://try.ocamlpro.com/>.
 
-To test things on your own computer, run the `ocaml` command and enter statements in the standard REPL (Read–Eval–Print Loop) provided.
+To install OCaml on your computer, see the [Install](/docs/install.html) documentation.
+
+To quickly try small OCaml expressions, you can use a toplevel, or REPL (Read–Eval–Print Loop). The `ocaml` command provides a very basic toplevel (you should install `rlwrap` through your system package manager and run `rlwrap ocaml` to get history navigation). If you can install it through [OPAM](/docs/install.html#OPAM) or your system package manger, we recommend the use of the [utop](https://github.com/diml/utop) toplevel instead, which has the same basic interface but is much more convenient to use (history navigation, auto-completion, etc.).
+
 Use `;;` to indicate that you've finished entering each statement:
 
-```shell
+```console
 $ ocaml
-        OCaml version 4.01.0
+        OCaml version 4.02.3
 
 # 1+1;;
 - : int = 2
 ```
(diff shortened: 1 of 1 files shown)


agarwal:
Thanks for the contribution.

> I used full URLs instead of relative navigation. Is there a preferred way?

Yes, you should use absolute or relative paths, depending on what makes more sense. Use absolute if the relative position of the document being linked to isn't meaningful. Specifically for your link, use an absolute path; change `https://ocaml.org/docs/install.html` to `/docs/install.html`. If you were referring to another tutorial, it would make more sense to use a relative path because all the tutorials are expected to be in the same folder.

gasche:
Thanks, I updated the PR to use absolute paths.

Chris00:
On Thu, 1 Oct 2015 05:42:41 -0700, Ashish Agarwal wrote:

> Thanks for the contribution.
> 
> | I used full URLs instead of relative navigation. Is there a preferred way?
> 
> Yes, you should use absolute or relative paths, depending on what makes more sense. Use absolute if the relative position of the document being linked to isn't meaningful. Specifically for your link, use an absolute path; change https://ocaml.org/docs/install.html to /docs/install.html. If you were referring to another tutorial, it would make more sense to use a relative path because all the tutorials are expected to be in the same folder.

Also, the use of absolute paths disallow local testing.

agarwal:
> Also, the use of absolute paths disallow local testing.

I guess you mean the path with the domain. By "absolute", I meant starting with a slash, but still excluding the domain.

Chris00:
On Thu, 1 Oct 2015 07:16:31 -0700, Ashish Agarwal wrote:

> | Also, the use of absolute paths disallow local testing.
> 
> I guess you mean the path with the domain. By "absolute", I meant starting with a slash, but still excluding the domain.

When you test locally (not running a web server), the scheme is file:// so using a path starting with / will look for the file from the root of your filesystem.  Not what you want.

agarwal:
Thanks for the clarification. IIUC, my suggestion is however still correct since our scripts convert absolute paths to relative paths during local testing.

PR #712, opened 2015-10-05
Add IRILL in the Giving Support area

lefessan:
As I said in a previous issue, there are some other non-commercial contributors to OCaml, that would deserve by their major contributions to be in the "Giving Support" section


agarwal:
Thanks! Even I learned more from this pull request.

PR #715, opened 2015-10-15
reference newer ocamlbuild manual

c-cube:
The new ocamlbuild manual by @gasche is very nice, and it should be mentioned on ocaml.org.

### DIFF
--- a/site/learn/tutorials/ocamlbuild/index.md
+++ b/site/learn/tutorials/ocamlbuild/index.md
@@ -6,6 +6,7 @@
 ## Documentation
 
 + [User manual](http://gallium.inria.fr/~pouillar/ocamlbuild/ocamlbuild-user-guide.html)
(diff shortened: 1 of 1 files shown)


gasche:
Thanks for the nice words. In my mind the current home of the new manual is temporary, and there are some bits in it that will only make sense after the next release (sometime this year). I'm fine with the pull-request as it is, but then I'll update the text after the manual has moved to a more definitive place (at least for the user-exposed version; the send-me-pull-requests place may remain the same).

c-cube:
In fact, it would be reasonable to move the new ocamlbuild manual into ocaml.org itself, I think, once you are reasonably satisfied with it. I still think it can be referenced now, leaving the URL to be updated when convenient.

agarwal:
I've merged, but I'd have been happy referring to @gasche's manual as _the_ manual, and link to the original as the _older_ one.

PR #717, opened 2015-10-23
Fix error in section on mutual recursion

mclehman:
The code illustrating mutual recursion to determine parity returned incorrect values. As given, `even 5` returned true and `odd 5` returned false.


dsheets:
Thanks!

PR #720, opened 2015-10-25
Fix typo in C/C++ comparison

mclehman:
Example C expression lacked trailing semicolon

### DIFF
--- a/site/learn/tutorials/hashtbl.md
+++ b/site/learn/tutorials/hashtbl.md
@@ -84,7 +84,7 @@ Hashtbl.find my_hash "t"
 
 
 To find out whether there is an
-entry in the `my_hash` for a letter we would do:
+entry in `my_hash` for a letter we would do:
 
 ```ocamltop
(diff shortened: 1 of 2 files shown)


mclehman:
Meant to add second commit as a PR from a second branch. Let me know if you'd prefer I redo it.

site/learn/tutorials/hashtbl.md
```diff
@@ -45,14 +45,16 @@ Hashtbl.add my_hash "w" "world";
...
 
 ```ocamltop 
 Hashtbl.find my_hash "h"
 ```
+Notice how it returns just one element? That element
+was the last one entered in with the value of `"h"`.
+
 What we probably want is all the elements that start with `"h"`. To do
-this we want to find all of them. What better name for this then
+this we want to *find all* of them. What better name for this then
```
dsheets, line 57: ... for this _than_...

mclehman, line 57: Oops. Good eye.

PR #722, opened 2015-10-25
Fix typos, change style

mclehman:
- Removes superfluous "just"
- "semantic" -> "semantics"
- Stylistic change. Splits sentence to avoid referring to pointers with "their" and the developer with "it". That combination is confusing in my opinion.
- Adds missing comma
- "in" -> "into"
- "points" -> "point"


Chris00:
Thanks.

PR #727, opened 2015-10-26
Discourage use of 4.02.0

gasche:
Hugo Herbelin ( @herbelin ) reports that he mistakenly tried 4.02.0 and observed massive Coq compilation times. I'm trying to update the web-facing information to avoid users getting caught like this.


gasche:
An even more future-proof to avoid this issue would be to be able to mark an OPAM switch as deprecated: I discuss this in ocaml/opam#2046 .

PR #745, opened 2015-12-10
Wrong LGPL clause number referenced in exception

camillem:
Hi,
The clause number referenced in the exception is apparently wrong
See
https://github.com/ocaml/ocaml/commit/2d26308ad4d34ea0c00e44db62c4c24c7031c78c#diff-9879d6db96fd29134fc802214163b95a
(same problem on Inria's website : http://caml.inria.fr/ocaml/license.en.html )
Have a nice day,
Camille


agarwal:
Instead of this change, we should avoid duplicating the license on ocaml.org altogether. Let's instead link directly to the license file on the ocaml repo.

PR #747, opened 2015-12-26
added /usr/lib/ocaml

haytham-med:
clarify location of the standard library

### DIFF
--- a/site/learn/tutorials/modules.md
+++ b/site/learn/tutorials/modules.md
@@ -315,7 +315,7 @@ or by specifying this in the .mli file:
 module F (X : X_type) : Y_type
 ```
 Overall, the syntax of functors is hard to grasp. The best may be to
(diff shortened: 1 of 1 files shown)


pw374:
But it's not always in that place...

agarwal:
Maybe you'd like to link to the files on github, e.g. [set.ml](https://github.com/ocaml/ocaml/blob/trunk/stdlib/set.ml).

haytham-med:
pw374, isn't it always in this location in linux?

pw374:
Nope. It's pretty trivial to make a Linux distribution that installs it elsewhere.

Philippe Wang
(mail sent from a phone)

pw374:
Good idea, in my opinion.

Philippe Wang
(mail sent from a phone)

Chris00:
Fixed following @agarwal 's suggestion.

PR #749, opened 2016-01-08
Update install.md

msprotz:
The descriptions for the various windows flavors was a little bit misleading and out-of-date. I've hopefully fixed that.

### DIFF
--- a/site/docs/install.md
+++ b/site/docs/install.md
@@ -240,11 +240,7 @@ urpmq --list | grep ocaml
 
 ## Windows
 
-Under Windows, two solutions are available to use OCaml: the official
-OCaml distribution, and OCamlPro's provides OCPWin.
-
-
-Three ports of OCaml for Microsoft Windows are currently available. For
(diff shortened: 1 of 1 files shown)


pw374:
This patch looks good to me. I don't know why this hasn't been merged yet... but now it has conflicts... :-/

pw374:
Thanks @msprotz

lefessan:
Since your patch is touching the part on OCPWin, could you add back the link, since the problem for which it has been removed has now been known to be fixed for the last four months ?

pw374:
Apologies, I thought I was careful enough for that link to stay! But I guess in all the rebase/not-rebase attempts for merging this, I missed something. Thanks for pointing this mistake.

Since we're at it, I wondered if we should put OCPWin before the other ways, because OCPWin seems to be the easiest way.

pw374:
Oh! But I was careful enough not to let that link disappear! Don't rely on the diff of this PR, because the merge was not done on Github, so the diff does not show at all the result of the merge!

PR #751, opened 2016-01-15
Add back link to ocpwin

lefessan:
It seems that the link to ocpwin was removed some time ago, for a bug that has now been fixed. It's nice to know that all other packaging platforms linked on this page are free of bugs.


agarwal:
@lefessan I agree that a warning could have been added explaining the issue, instead of removing the link. Thanks for adding it back.

PR #752, opened 2016-01-27
Korean translation update

seungjin:
Two letter English(Roman Alphabet) abbreviation to its native language(Korean).

### DIFF
--- a/script/translations.ml
+++ b/script/translations.ml
@@ -7,6 +7,7 @@ let lang_to_html l =
   if l = "" then "en" (* default language *)
   else if l = "zh" then "中文"
   else if l = "ja" then "日本語"
+  else if l = "ko" then "한국어"
   else l
 
 let translations path =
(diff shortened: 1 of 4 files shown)


agarwal:
I'm accepting this patch though I don't know Korean. @seungjin Thanks for your contributions. Perhaps you can explain a couple of points:
- In several places you replaced 기본기 with 기본사항. Can you explain why?
- You added large pieces of English in some places. Are you planning to incrementally convert that to Korean?

Thanks.

agarwal:
Sorry, it was of course already merged, so I just meant to ask the questions.

seungjin:
- 기본기(previous one) and 기본사항(newer one) can be the same meaning. But the previous one is more like "basic skill" and the newer one is "basic information". For example, we mostly use 기본기(previous one) for meaning the "basic" skill of martial art. So I changed to more general term that sound about right for the context. 
- Yes, I am planing and doing translation those to English when I get free time.

agarwal:
@seungjin Thanks for the answers!

PR #754, opened 2016-02-01
Add Fabrice and Louis in about.html

lefessan:
Names have been sorted alphabetically, since the former order was not
specified.


avsm:
Woops, I shouldn't be in there -- any chance you could excise me in this PR as well? :-)

amirmc:
Anil is meant to be in that list as he maintains the actual machine that the site is served from. That's why he's part of the 'WWW project' (see additional info on the infrastructure [wiki page](https://github.com/ocaml/infrastructure/wiki/www.OCaml.org)). I think we need to add David Sheets too.

I'll submit a PR from a branch with those edits.

PR #758, opened 2016-02-20
Add Rudenoise OCaml Posts feed to plant_feeds

rudenoise:
Hi,

just adding my feed details as requested by Amir.

Cheers

Joel


amirmc:
Thanks for adding the feed!  I've submitted a minor PR to your fork.

The CI error here is quite bizarre. I'm not quite sure what's going on but I don't think it's a result of this PR.

rudenoise:
My commit doesn't seem to be the cause of the error.

From what I can see, opam is not able to load a config file for OCaml 4.01.0, the error is broadcast here: https://github.com/ocaml/opam/blob/master/src/state/opamState.ml#L1406

Will close and reopen to trigger fresh build, in the hope that it's a network error during refresh.

amirmc:
I'm not sure about the errors in Travis but we should track those separately. I'll merge this PR since I don't think it's related to the problem.

PR #763, opened 2016-02-29
Add Gina Maini's blog "Gina Codes" to the planet_feeds.txt file

wiredsister:
- this feed is filtered by the tags & categories for OCaml [http://www.gina.codes/feed.ocaml.xml]
- if you have any questions, @wiredsis on twitter or email Gina at gina@beancode.io


amirmc:
Thanks for this! I notice in the feed that there isn't any excerpt or body content. Just the title. Is this intended?

wiredsister:
@amirmc Yeah... that's not intended. I also noticed I'm duplicating tags and categories. I'll update and reopen.

PR #772, opened 2016-03-25
reorganize teaching page with regions

yminsky:
Three regions listed: North America, Europe and Asia.  Also, sorted within regions.


amirmc:
Wonderful! Thank you very much :)

Issue #713, opened 2015-10-08
add instructions for contributing


agarwal:
As requested by @gasche in this [email](http://lists.ocaml.org/pipermail/infrastructure/2015-October/000555.html), we should explain how contributions can be made to the site. We should:
- [ ] Write an explanation of how to contribute, especially how to contribute content. How to contribute to the site's build is probably less important but some basic information couldn't hurt. These instructions could be potentially go in `contributors.md`, making the page more about "contributing".
- [ ] Link to above from:
  - the Contact section in footer
  - the About page
  - the repo's README

PR #777, opened 2016-04-08
tutorial/basics add syntax for fun and function

mstrhh:
see title. My first try to make a contribution.

### DIFF
--- a/site/learn/tutorials/basics.md
+++ b/site/learn/tutorials/basics.md
@@ -470,7 +470,7 @@ important but often ignored runtime warnings, as in Perl).
 
 ## Function definition syntax
 
-There are several ways to define functions, which might be confusing at first.
+There are several ways to write a function definition, which might be confusing at first.
 
 Let's take the previous example and rewrite it:
 
@@ -487,7 +487,8 @@ a function that takes exactly 1 argument.
 
 All three average functions will compile to the same code.
 
-The `function` syntax variant is what is done internally, and you can see how 
-the code is structured exactly like the inferred type.
+The `function` syntax variant is what is done internally, and you can see 
+that this code corresponds more closely to what the toplevel prints 
+when showing the function type.
 
 [utop]: https://github.com/diml/utop
--- a/site/learn/tutorials/basics.md
+++ b/site/learn/tutorials/basics.md
@@ -468,5 +468,26 @@ removes a whole class of errors which cause segfaults,
 `NullPointerException`s and `ClassCastException`s in other languages (or
 important but often ignored runtime warnings, as in Perl).
 
+## Function definition syntax
+
+There are several ways to define functions, which might be confusing at first.
+
+Let's take the previous example and rewrite it:
+
+```ocamltop
+let average1 = 
+  fun a b -> (a +. b) /. 2.0
+
+let average2 = 
+  function a -> function b -> (a +. b) /. 2.0
+
+```
+The keyword `fun` defines a function that takes 1..n parameters. The keyword `function` defines
+a function that takes exactly 1 argument. 
+
+All three average functions will compile to the same code.
+
+The `function` syntax variant is what is done internally, and you can see how 
+the code is structured exactly like the inferred type.
 
 [utop]: https://github.com/diml/utop


dbuenzli:
I'm not aware of the surrounding text, but I don't think that the snippet is a very good explanation. Especially "The `function` syntax variant is what is done internally," is more confusing than enlighting in my opinion and leaves the impression that all this syntax was introduced only to annoy the programmer.

One should start by introducing `function`, an use `let f = function x ->...`. Then show the shortcut `fun` for multiple arguments with the idea that `let f = fun x y -> ...` is equivalent to `let f = function x -> function y -> ...`  and finally introduce the syntactic sugar `let f x y = ...` that stands for `let f  = fun x y -> ...`. The text should makes it clear that we have two separate ideas here 1) a notion of _function as value_ and ways to declare them and 2) a way to _name_ these values which is the `let` which has syntactic support for functional values since those are so widespread.

It should also be noted that `function` itself is argument-wise definitionally more expressive since it allows to pattern match on the argument, something both `fun` and the `let` sugar do not allow you to do.

mstrhh:
@dbuenzli in the surrounding text this function definition syntax is not mentioned at all, so this addition makes sense. The fact that more can be said to it is no argument against starting with showing fun and function usage.

dbuenzli:
> The fact that more can be said to it is no argument against starting with showing fun and function usage.

It's not that more can be said: I don't think that what is being said by this addition is well said and enlightening to a newcomer.

mstrhh:
It is that more can be said - I am reading the paragraphs _One should start by ..._ and _It should also be noted ..._ this way.

I disagree that it's not enlightening to the newcomer. Experience tells me that there is not one single way to explain something to beginners.

Apart from this, if my contribution is not plain wrong the ocaml.org team could decide to let it evolve online. This is how online communities are working. Then I could add it to other languages.

pw374:
@dbuenzli I strongly believe it's generally a bad idea to follow the order "let f = function x / let f = fun x / let f x =" when it's about introducing the language as a "real programming language".

The order where we start by the extreme-core language and then syntactic sugar is relevant only when you deal with a semantics course. Not a programming one.

Start with the simplest things, and then when the simple things are well-understood, it's okay to go deeper in the details. 

In the same vein, you don't start with explaining that `1 + 1` is syntactic sugar for `(+) 1 1`.

pw374:
@mstrhh Thank you for your contribution. I'm not 100% convinced that this "function syntax" stuff should be in this page, in the **Basics**. However one thing that would make this a little more legitimate would be to explain anonymous functions first! When one writes `function x ->`, one is defining an anonymous function (regardless of if it's given a name some way or another). So, instead of relying on the compiler to explain and justify syntactic sugar, perhaps we should focus on basic semantics (please do avoid at all costs denotational semantics, or, even worse, operational semantics, in this page, because even if the topics are very interesting, they don't have their place on this page!)

That being said... Let's merge, and move forward by asking for more contributions!

dbuenzli:
Le dimanche, 10 avril 2016 à 12:01, Philippe Wang a écrit :

> @dbuenzli (https://github.com/dbuenzli) I strongly believe it's generally a bad idea to follow the order "let f = function x / let f = fun x / let f x =" when it's about introducing the language as a "real programming language".

Right, let's try not to give sense at all to all these things and let's not put emphasis on the notion of value. After all that's what "real programming" is about: random people trying to bang imperative statements until they are satisfied by some kind of random output.

If I were a newcomer and reading this section my only reaction would why the heck do they have so many different syntaxes for the same thing, that doesn't look nice. Things are more acceptable once you can explain them.  

Daniel

pw374:
Ok but then, start with the notion of value! What you're saying seems to be completely out of context.
Have a look at the whole page http://ocaml.org/learn/tutorials/basics.html first perhaps?!

dbuenzli:
> What you're saying seems to be completely out of context.

It's not out of context, I think it is a bad idea to introduce this syntax here as introduced by this PR. I suggested how I think it could be meaningfully be introduced so that it doesn't make newcomer wonder why these OCaml people have such a mess with their syntax. If it doesn't fit because the tutorial is broken then either rewrite the tutorial or avoid mentioning these things here.

pw374:
> If it doesn't fit because the tutorial is broken then either rewrite the tutorial or avoid mentioning these things here.

Or give a chance to people to contribute and try make it better? 

I think this tutorial is bad. No, actually, no, I think it's very very bad. I don't know who wrote it. (But I do hope I'm not hurting any feelings by saying it's very very bad.) 

I think it's not worse with this PR merged than not merged.

So, let's say that we're at 1—1.1 (1 for @dbuenzli, 1 for @mstrhh, and 0.1 for myself), so I feel it's okay to have this merged. Now, @dbuenzli, if you find another person to add a vote to undo the changes of this PR, we'll hit undo. ;-)

dbuenzli:
> Or give a chance to people to contribute and try make it better?

The point is that this makes it even worse. But anyways I don't care, we do not have the same quality expectations it seems.

pw374:
> The point is that this makes it even worse. 

By how much? 

> But anyways I don't care, we do not have the same quality expectations it seems.

I hope we'll reach the quality you want, but it's not by blocking the few attempts we have that we'll go anywhere.

dbuenzli:
> By how much?

Too much.

pw374:
ok... Let's rewrite this tutorial...

mstrhh:
@pw374 Thank you, I was getting quite exasperated about the first reaction to my microcontribution to what looks like one of the more coherent paragraphs of the tutorials.

I mean for example compared to [the mega crash course](http://ocaml.org/learn/tutorials/get_up_and_running.html).

Rewrite it all? With this friction? And the style questions - could different people write in the style of [the null pointers part?](http://ocaml.org/learn/tutorials/null_pointers_asserts_and_warnings.html). Is this recommended style?

Ideas:
- starting [with some guidelines](https://github.com/ocaml/ocaml.org/issues/713). 
- and then preparing a tutorial chapter structure, 2 levels or so
- maybe a paragraph having a master author allowing him to write in his style, until the paragraph is replaced by a better one (for not saying until it is downvoted)

/Str.

dbuenzli:
> @pw374 Thank you, I was getting quite exasperated about the first reaction to my microcontribution to what looks like one of the more coherent paragraphs of the tutorials.

@mstrhh What you don't get here is that this is not about style, it's about _content_, it's about the first conceptual presentation of the language to newcomers. Adding more content doesn't necessarily imply improvement.

mstrhh:
@dbuenzli about style: your answer sounds like you are issuing a "style <> content" warning without reading the [null pointers linked text](http://ocaml.org/learn/tutorials/null_pointers_asserts_and_warnings.html): read about a website and a database administrator and stubborn users and tell me that style does not matter. (No offense to the writer of this section.)

about "the first conceptual presentation": is it that? Tutorials on the net are usually rather: first steps, see some syntax, write some code, "how I stopped worrying about what is it in it's essence and started writing happily Some OCaml". Okay, get some concepts. A conceptual presentation is a CS introductory course, what Dan Grossman did (I like it).

Is this discussion leading somewhere? De la discussion jaillit la lumière?

dbuenzli:
> "how I stopped worrying about what is it in it's essence and started writing happily Some OCaml".

That's precisely what I'd like to avoid, if the only thing you are interested is presenting the syntax so that one can write code like one is used to write it in another language then there is no point in learning OCaml. 

Learning a new language should be an opportunity to have another conceptual take on computation --- especially in this case insisting on values and expressions vs statements. These days everybody pretends to be doing "functional programming", sadly rarely is this the case in practice, people are just writing, imperative, sequential spaghettis in a typed language. If you want to change this you need to talk about essence, from the onset.

mstrhh:
@dbuenzli okay, that is a clear mission statement. 

But the world is not 0 or 1, rather many 0's and 1's.  Youtube is full of presentations of languages and their concepts, but many of them are just "warmed up air" because they fail to give the attendents an experience of what this in practice means.

I'll shut up now, this is not my job-site.

PR #778, opened 2016-04-18
Remove the link to the interactive session at OCamlPro's website

armish:
FWIW: as a newbie, I found it much harder to work with that interactive demo and learn OCaml as it is both out-dated ([28 issues](https://github.com/OCamlPro/tryocaml/issues) and [5 PRs](https://github.com/OCamlPro/tryocaml/pulls) still open waiting to be fixed/merged) and not being actively maintained (`Latest commit ddd533d  on Feb 24, 2015`).

### DIFF
--- a/site/learn/tutorials/basics.md
+++ b/site/learn/tutorials/basics.md
@@ -6,8 +6,6 @@
 
 ## Running OCaml code
 
-The easiest way to get started is to run an interactive session in your browser, at <http://try.ocamlpro.com/>.
-
 To install OCaml on your computer, see the [Install](/docs/install.html) documentation.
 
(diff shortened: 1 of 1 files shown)


dbuenzli:
I think it's still nice for people to be able to test snippets in the browser. Maybe link to http://ocsigen.org/js_of_ocaml/2.7/files/toplevel/index.html or better have a hosted version of that on ocaml.org ?

armish:
@dbuenzli: agree with you on that point, but I don't think it is a priority. 

That link sitting at the top of the documentation makes it look like it is one thing that everybody should definitely check out; but, in fact, it is a relatively confusing resource for beginners.

klakplok:
Indeed, we do not maintain this version of TryOCaml anymore.
We are working on a replacement, based on the exercise platform we built for the OCaml MOOC.
A demo of the MOOC platform as it was at the end of the MOOC (without our recent developments) is available at [https://try.ocamlpro.com/fun-demo/].

klakplok:
I'm not sure this replacement was the right move.
TryOCaml has been designed as a learning tool for beginners, the js_of_ocaml toplevel is more of a technical demo or a tool for people who already know some OCaml.
So if the unfixed typos and small bugs are really a problem, I suggest that you just drop the link instead of replacing it with the link to the toplevel, until a new beginner-friendly environment is available.

Chris00:
The goal is to have a toplevel to try the code in the tutorials, it does not need to support a learning environment itself.

Note that the link on the front page purposely still points to tryocaml.  It would be good to have an updated version soon.

PR #779, opened 2016-04-28
basic tutorial on error handling

c-cube:
explains quickly exceptions and the `result` type, with a few possible ways of representing errors with `result`.

### DIFF
--- a/site/learn/tutorials/error_handling.md
+++ b/site/learn/tutorials/error_handling.md
@@ -6,23 +6,140 @@
 
 ## Exceptions
 
-TODO: explain briefly
-TODO: lack of information in types
+The traditional way of handling errors in OCaml is exceptions. The
+stdlib relies heavily upon them.
 
-TODO: Printexc.register_printer/to_string
-TODO: backtraces
+Exceptions belong to the type `exn` (an extensible sum type):
+
+```ocaml
+
+exception Foo of string
+
+let i_will_fail () =
+  raise (Foo "ohnoes!")
+
+```
+
+Here, we add a variant `Foo` to the type `exn`, and create a function
+that will raise this exception. Now, how do we handle exceptions?
+The construct is `try ... with ...`:
+
+```ocaml
+let safe_inverse n =
+  try Some (1/n)
+  with Division_by_zero -> None
+
+let safe_list_find p l =
+  try Some (List.find p l)
+  with Not_found -> None
+```
+
+We can try those functions:
+
+```ocaml
+# 1/0;;
+Exception: Division_by_zero.
+# safe_inverse 2;;
+- : int option = Some 0
+# safe_inverse 0;;
+- : int option = None
+
+# List.find (fun x->x mod 2 =0)  [1;3;5]
+Exception: Not_found.
+# safe_find (fun x->x mod 2 =0)  [1;3;4;5]
+- : int option = Some 4
+# safe_find (fun x->x mod 2 =0)  [1;3;5]
+- : int option = None
+```
+
+The biggest issue with exceptions is that they do not appear in types.
+One has to read the documentation to see that, indeed, `Map.S.find`
+or `List.hd` are not total functions, and that they might fail.
+
+### Printing
+
+To print an exception, the module `Printexc` comes in handy. For instance,
+the following function `notify_user : (unit -> 'a) -> 'a` can be used
+to call a function and, if it fails, print the exception on `stderr`.
+
+```ocaml
+let notify_user f =
+  try f()
+  with e ->
+    let msg = Printexc.to_string e in
+    Printf.eprintf ("there was an error: %s\n" msg);
+    raise e
+```
+
+OCaml knows how to print its builtin exception, but you can also tell it
+how to print your own exceptions:
+
+```ocaml
+exception Foo of int
+
+let () =
+  Printexc.register_printer
+    (function
+      | Foo i -> Some (Printf.sprintf "Foo(%d)" i)
+      | _ -> None (* for other exceptions *)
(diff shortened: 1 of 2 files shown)


site/learn/tutorials/index.md
```diff
@@ -32,6 +32,7 @@ tracker there to request or offer new tutorials. Thanks!
 * [Formatting and Wrapping Text](format.html)
 * [99 Problems (solved) in OCaml](99problems.html)
 * [OCamlbuild](ocamlbuild/)
+* [Error handlibg](error_handling.html)
```
trefis: Typo.

site/learn/tutorials/error_handling.md
```diff
@@ -0,0 +1,145 @@
+<!-- ((! set title Error Handling!)) ((! set learn !)) -->
+
+*Table of contents*
+
+# Error Handling
+
+## Exceptions
+
+The traditional way of handling errors in OCaml is exceptions. The
```
dbuenzli: Somehow I personally think we should move away from this model. So I'd rather say the `old-fashioned way`.

site/learn/tutorials/error_handling.md
```diff
@@ -0,0 +1,145 @@
...
+Exception: Not_found.
+# safe_find (fun x->x mod 2 =0)  [1;3;4;5]
+- : int option = Some 4
+# safe_find (fun x->x mod 2 =0)  [1;3;5]
+- : int option = None
+```
+
+The biggest issue with exceptions is that they do not appear in types.
+One has to read the documentation to see that, indeed, `Map.S.find`
+or `List.hd` are not total functions, and that they might fail.
```
dbuenzli, line 57: I think it would be nice to have a sentence around here that explains that using exceptions for non-exceptional error conditions is bad (e.g. parsing user input, trying to find an element in a data structure where it may not be).

site/learn/tutorials/index.md
dbuenzli: Typo ?

site/learn/tutorials/error_handling.md
```diff
@@ -0,0 +1,145 @@
...
+to call a function and, if it fails, print the exception on `stderr`.
+
+```ocaml
+let notify_user f =
+  try f()
+  with e ->
+    let msg = Printexc.to_string e in
+    Printf.eprintf ("there was an error: %s\n" msg);
+    raise e
+```
```
dbuenzli, line 110: Why not show how to print stack traces directly ? These kind of bit of codes get c&p around and let's have good error reports through good c&p.

site/learn/tutorials/error_handling.md
c-cube: that's your opinion, I think exceptions are very nice in many cases. I don't want this page to be biased.

site/learn/tutorials/error_handling.md
c-cube, line 110: I added a `Printexc.get_backtrace()`, indeed.

site/learn/tutorials/error_handling.md
dbuenzli: > that's your opinion, I think exceptions are very nice in many cases. I don't want this page to be biased.

Hum, this is equally biased... if you like to shoot yourself in the foot, it's your problem.

It seems there's a consensus, IIRC even among the OCaml dev-team, that using `Not_found` and `Failure` exceptions in the way they are used in the stdlib was a design error. It would be nice if that error could not be perpetuated to newcomers.

c-cube: `Failure` is certainly a mistake, because it is too wide in scope and people rely on the error message (hence the new warning). I am not aware of a consensus that `Not_found` is bad, if you also provide a `option`-based version of the function. There are also projects that use exceptions and `Printexc` to manage errors globally (when they have a large number of unrecoverable errors), rather than  threading a result type in every function. Exceptions are not deprecated, even if you don't use them.

dbuenzli: > I am not aware of a consensus that Not_found is bad, if you also provide a option-based version of the function.

Maybe you didn't debug enough programs that simply crash with `Not_found` on stdout. The only sane way of dealing with this is to have `find` that returns an option and `get` that raises `Invalid_argument` (i.e programming error) if not found only to be used for code path were you know by construction that the thing should be in there (of course if we had a good `Option` module in the stdlib the latter would not even be needed). 

> Exceptions are not deprecated, even if you don't use them.

We are talking about material for the newcomer, put them on good tracks.

avsm: Why not just state that "One way of handling errors in OCaml is exceptions" and remain objective?  The subjective portion comes later on when you say "It is considered good practice nowadays, ..." at the point you describe `Result.t`.

c-cube: well that's why I suggest the `foo`/`foo_exn` construct. I indeed had to debug a lot of `Not_found` (well, backtraces to the rescue) and so will any newcomers. Exceptions won't disappear.

dbuenzli: > well that's why I suggest the foo/foo_exn construct. 

Ah hungarian notation... Well every one has their own taste.

> I indeed had to debug a lot of Not_found (well, backtraces to the rescue) and so will any newcomers. Exceptions won't disappear.

But you can minimize them. Sorry to see you don't understand the social aspects of this.

c-cube:
can we move on on this issue?

pw374:
Thank you @c-cube for your contribution!

@dbuenzli don't forget you may open PRs with changes to this page to enhance it. ;-)

PR #785, opened 2016-05-10
Update basics.md

arglucas:
Fixed an error in the range function that results in it always returning []. (Added brackets around recursive call to range).


arglucas:
Looking at the Travis error this looks to be something about the build rather than the change although happy to dive in and take a more detailed look.

Error: Unbound module CalendarLib (Line 382)

dbuenzli:
Your problem must be somewhere else, you dont need these parens (the code works as expected).

arglucas:
Thanks.

PR #787, opened 2016-05-15
releases: reformulate 'latest' handling

gasche:
I found having an explicit list item "latest", followed by another item "4.03.0", was a bit awkward. What do you think?


pw374:
In my opinion it's better with your PR, so I've merged.
Thanks.

agarwal:
Looks good. Thanks.

PR #788, opened 2016-05-16
FAQ: update for label/constructor disambiguation in 4.02

whitequark:
Also, clarify some other passages that are confusing unless the reader
keeps Caml Light in mind.

### DIFF
--- a/site/learn/faq.md
+++ b/site/learn/faq.md
@@ -193,7 +193,7 @@ type point_2d = {x : float; y : float};;
 {x = 10.; y = 20.; z = 30.};;
 ```
 
-Since OCaml 4.02, field are automatically disambiguated when types are
+Since OCaml 4.02, fields are automatically disambiguated when types are
 known. For example, in `let u:point_3d = ... in u.x`, `u.x` refers to
 the field of `point_3d` even though it is shadowed. However, field
 disambiguation does not work when type information is not available
@@ -249,11 +249,13 @@ and you can even coerce a `point_3d` to a `point_2d`.
 
 #### How to define two sum types that share constructor names?
 
-Since OCaml 4.02, constructors are automatically disambiguated based on
-arity. For example, in `type a = A;; type b = A of int;; A(1)`, `A`
-has the type `a` even though its constructor is shadowed. However, this
-disambiguation may produce extremely surprising results (think of
-`type x = true of bool;; true true;;`, so one may consider avoiding
+Since OCaml 4.02, constructors are automatically disambiguated when types
+are known. For example, in `type a = A;; type b = A of int;; let x:a = A`,
+`A` is recognized as belonging to the type `a` even though its constructor
+is shadowed. However, constructor disambiguation does not work when type
+information is not available (e.g. in `let get_n x = match x with A -> 1`
+where the type of `get_n` is not otherwise constrained), and may produce
+confusing results when types are omitted, so one may consider avoiding 
 the problem entirely.
 
 Generally speaking, sharing names between two constructors is not
--- a/site/learn/faq.md
+++ b/site/learn/faq.md
@@ -248,7 +248,15 @@ and you can even coerce a `point_3d` to a `point_2d`.
 
 #### How to define two sum types that share constructor names?
 
-Generally speaking you cannot. As for all other names, you must use
+Since OCaml 4.02, constructors are automatically disambiguated based on
+arity. For example, in `type a = A;; type b = A of int;; A(1)`, `A`
+has the type `a` even though its constructor is shadowed. However, this
+disambiguation may produce extremely surprising results (think of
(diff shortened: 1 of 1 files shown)


site/learn/faq.md
```diff
@@ -192,14 +192,23 @@ type point_3d = {x : float; y : float; z : float};;
 type point_2d = {x : float; y : float};;
 {x = 10.; y = 20.; z = 30.};;
 ```
-The simplest way to overcome this problem is simply ... to use different
+
+Since OCaml 4.02, labels are automatically disambiguated when types are
+known. For example, in `let u:point_3d = ... in u.x`, `u.x` refers to
+the field of `point_3d` even though it is shadowed. However, this
+disambiguation does not work in all cases, and may produce confusing
```
Chris00: > this disambiguation does not work in all cases

In which case is it not working?

whitequark: When the type is not known at the time when it is performed.

Chris00: On 2016-05-16, whitequark wrote:

> In site/learn/faq.md:
> 
> > @@ -192,14 +192,23 @@ type point_3d = {x : float; y : float; z : 
> > float};;
> >  type point_2d = {x : float; y : float};;
> >  {x = 10.; y = 20.; z = 30.};;
> > 
> > ```
> > -The simplest way to overcome this problem is simply ... to use 
> > different
> > +
> > +Since OCaml 4.02, labels are automatically disambiguated when types 
> > are
> > +known. For example, in `let u:point_3d = ... in u.x`, `u.x` refers 
> > to
> > +the field of `point_3d` even though it is shadowed. However, this
> > +disambiguation does not work in all cases, and may produce 
> > confusing
> > ```
> 
> When the type is not known at the time when it is performed.

Because "this disambiguation" follows an example where the type is 
explicitly given, one could think it refers to that.  I think the 
sentence should be reworded to be less... ambiguous.

whitequark: Wish granted.

site/learn/faq.md
```diff
@@ -239,7 +248,15 @@ and you can even coerce a `point_3d` to a `point_2d`.
 
 #### How to define two sum types that share constructor names?
 
-Generally speaking you cannot. As for all other names, you must use
+Since OCaml 4.02, constructors are automatically disambiguated based on
+arity. For example, in `type a = A;; type b = A of int;; A(1)`, `A`
```
Chris00: I do not think that disambiguation is based on arity.  In OCaml 4.03.0:

``` ocaml
# type a = A of int * int;;
type a = A of int * int
# type b = A of int;;
type b = A of int
# A(1,1);;
Characters 1-6:
  A(1,1);;
   ^^^^^
Error: This expression has type 'a * 'b
       but an expression was expected of type int
```

It is _type directed disambiguation_ that works:

``` ocaml
# (A(1,1) : a);;
- : a = A (1, 1)
```

site/learn/faq.md
```diff
@@ -239,7 +248,15 @@ and you can even coerce a `point_3d` to a `point_2d`.
...
+has the type `a` even though its constructor is shadowed. However, this
+disambiguation may produce extremely surprising results (think of
+`type x = true of bool;; true true;;`, so one may consider avoiding
+the problem entirely.
```
Chris00, line 259: Thanks for your contribution.  However, this amounts to say that OCaml has a feature that is actually not so good...  I think this paragraph and the next one should be written to teach the user how to use this feature, possibly showing associated difficulties (e.g. teaching the user how to figure out more complicated error messages).

whitequark, line 259: > However, this amounts to say that OCaml has a feature that is actually not so good...

That's more or less true, if you ask me.

Chris00, line 259: > That's more or less true, if you ask me.

I disagree.  I don't know what you are expecting would be a solution in a type inferred language.  Moreover, this is backward compatible (you get the old behavior without type annotations) and I am not convicted about more complicated error messages.  In any case, the goal of a FAQ is not to give opinions about the languages but to help the user to understand what is available in a way he can have his own informed opinion.

whitequark, line 259: Yes. And that the feature is fragile is a fact, not an opinion.

Chris00, line 259: > Yes. And that the feature is fragile is a fact, not an opinion.

Then present compelling facts to enlighten the reader.

lefessan:
I agree with Chris00 that the paragraph on type disambiguation by arity is wrong, types are disambiguated by propagating types, not by checking their arity (`true true` works because the type checker infers that a boolean is expected as argument, and thus the second `true` is infered by just checking it belongs to bool).

Could you remove that paragraph, so that we can merge the first one ?

whitequark:
Removed all incorrect info.

whitequark:
Or more specifically, updated it to be correct.

PR #789, opened 2016-05-16
FAQ: clarify usage of tuple accepting functions

whitequark:
This seems likely to trip up programmers new to OCaml, who are almost
certainly used to delimiting arguments with commas.

### DIFF
--- a/site/learn/faq.md
+++ b/site/learn/faq.md
@@ -375,6 +375,12 @@ let sum' (x, y) = x + y;;
 sum' (1, 2);;
 ```
 
+By convention, OCaml code generally uses curried functions rather
+than functions accepting a tuple as an argument. Of course, this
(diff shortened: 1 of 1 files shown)


site/learn/faq.md
```diff
@@ -375,6 +375,12 @@ let sum' (x, y) = x + y;;
 sum' (1, 2);;
 ```
 
+By convention, OCaml code generally uses curried functions rather
+than functions accepting a tuple as an argument. Of course, this
+does not apply to cases where the tuple is denoting a data structure
+on its own (e.g. `(float, float, float)` being used to represent
```
lefessan: Maybe use `float * float` (type notation) instead of `(float,float,float)` which is a mix of types and values notations

whitequark:
Fixed

PR #790, opened 2016-05-16
FAQ: clarify section on mutually recursive functors

whitequark:
It referred to the entire chapter 7 as "subject to change at any time", which is
just meaningless FUD, and did not give a good example.

Oh, and the link was broken, because the anchor order was changed and
this wasn't even pointing at the recursive modules section (which prompted
me to inline the example).

### DIFF
--- a/site/learn/faq.md
+++ b/site/learn/faq.md
@@ -594,12 +594,31 @@ printing functions by the high level printing functions provided by
 
 ## Module Language
 
-#### Can I have two mutually recursive compilation units / structures / signatures / functors?
+#### Can I have two mutually recursive structures, signatures, functors?
 
-Currently not in the stable langage. However there exists an [OCaml
-extension](http://caml.inria.fr/pub/docs/manual-ocaml/extn.html#sec220)
-(which is subject to change or removal at any time) which addresses some
-of these problems.
+Yes, but they cannot be toplevel, and structures always have to have 
+an explcit signature. Recursive structures may be defined as follows:
+
+```ocamltop
+module rec A : sig
+  type a = { x: int }
+end = struct
+  type a = { x: int }
+  let b : B.b = { y = 1.0 }
+end and B : sig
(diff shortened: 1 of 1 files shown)


site/learn/faq.md
```diff
@@ -594,12 +594,31 @@ printing functions by the high level printing functions provided by
 
 ## Module Language
 
-#### Can I have two mutually recursive compilation units / structures / signatures / functors?
+#### Can I have two mutually recursive structures, signatures, functors?
+
+Yes, but they cannot be toplevel, and structures always have to have
```
lefessan: I think `cannot be toplevel` is a bit misleading: toplevel values are usually values that are defined in the first level of a module, whereas here, you probably mean `compilation units`. So, I would actually divide the reply in two sections: one for compilation units, one for non-compilation units, in which case you don't need to specify it anymore.

site/learn/faq.md
```diff
@@ -594,12 +594,31 @@ printing functions by the high level printing functions provided by
...
+
+In a similar way, mutually recursive signatures and functors can also
+be defined.
 
-Currently not in the stable langage. However there exists an [OCaml
-extension](http://caml.inria.fr/pub/docs/manual-ocaml/extn.html#sec220)
-(which is subject to change or removal at any time) which addresses some
-of these problems.
+Note that recursive compilation units are not possible: the toplevel
+structures and signatures corresponding to `.ml`/`.mli` files cannot
```
lefessan: That's not completely true: two compilation units can be recursive on types, as long as there is an order in which it is possible to compile them sequentially with all dependencies defined. It is typically the case when you export abstract versions of the types in the interfaces, and the manifest versions in the implementations can then refer to the types, leading to recursive manifest types, but with abstract interfaces.

whitequark: I rewrote it in the way you suggested, and added an example.

PR #794, opened 2016-05-22
Word missing in sentence

verbosemode:
"You should modules when the data structures..." should read "You should use modules when the data structures..."


pw374:
Thanks.

PR #798, opened 2016-06-19
fix spelling error in package manager

sblondon:
It's a minor patch, only fixing a spelling error:
manger -> manager


pw374:
thanks!

Issue #699, opened 2015-09-10
Google Archive of Caml Mailing List not Up to Date?


lindig:
The website points to Google Groups for the mailing list: [fa.caml](http://groups.google.com/groups?group=fa.caml). From my point of view, the latest posts there are from July 2015, suggesting the the archive is not kept up to date. If so, I think it either should be fixed, mentioned as incomplete, or the link be removed.

agarwal:
This is currently being discussed in a thread on the OCaml List.

amirmc:
Just for info, the thread starts at: https://sympa.inria.fr/sympa/arc/caml-list/2015-09/msg00024.html

I’d be in favour of mirroring the list to lists.ocaml.org somehow.  I believe this has been discussed before but no-one has had time to look into it.

lindig:
I think a more useful archive and interface than Google Groups is at [GMANE](http://dir.gmane.org/gmane.comp.lang.caml.inria). It provides a good web interface, as well as RSS and NNTP interfaces. It also seems to be up to date. I'd suggest to replace the link.

Chris00:
> The website points to Google Groups for the mailing list: fa.caml.

In my opinion, none of these should be on the website — the threads are not complete.  The interface of the mailing list at Inria may not be the easier one to deal with but it is the authoritative archive.

gasche:
@Chris00 invited me to share my pain points with the current Sympa interface here. I see three basic needs for mailing-list archives:
- Need 1: User is not currently subscribed to the list but want to have a look at how the OCaml community is doing lately. They need a reasonable way to have a look at recent discussions and follow a particular conversation they are interested in. (I do that myself with, say, llvm-dev or lkml). Most mailing-list archive browsing software will do the job there, and I think the current Sympa instance, if we removed the irritating paging and just listed the topics of the month, would do fine (we have to check that it's easy to follow a thread across two different months, that may be an extra pain point)
- Need 2: User remembers a particular message or thread that they want to link for reference. They need a way to search this message back in the archive. The Sympa search feature is completely broken for this purpose (I guess Google archives is better, but a well-indexed non-Google archive could do the job fine; I just sent a message to caml-list to ask to change robots.txt to allow indexing). As a second choice, suppose the user found the thread back in their mailbox, it should be very easy to find a thread from its precise title or its date (something the Sympa paging currently makes much more bothersome than it needs to be).
[...]

PR #802, opened 2016-07-16
Better mailing-lists page

gasche:
This is a first step towards improving our "places where community discussions happen" information on ocaml.org, taking ideas from #699 and the recent [caml-list discussions](https://sympa.inria.fr/sympa/arc/caml-list/2016-07/msg00018.html).

The changes currently proposed only do the minimal thing, which is to add more information to the page https://ocaml.org/community/mailing_lists.html, reachable from the "Community" page by cliking the "see more lists" link. I think that the community page itself should be changed to de-emphasize Mailing-lists and instead present "Discussion places" (including mailing-list, IRC, and maybe in the future some shiny 4.0 discussion with beginner questions and virtual reality pokemons or something).

I'm frustrated by the choice of presenting only a subset of the options on the Community page; I'm tempted to remove caml-annouce and lists.ocaml.org and replace them by the IRC channel and, for example, Caml Weekly News -- because subjectively they feel more important for the frontpage space. But I wonder if we couldn't just list all communication channels on the Community page, instead of this weird two-page split that makes many options harder to reach than they should be.

### DIFF
--- a/site/community/mailing_lists.md
+++ b/site/community/mailing_lists.md
@@ -25,9 +25,14 @@ the subscribers of the list. Messages are generally in English but
 sometimes also in French. In 2010, the list has about 1500 subscribers,
 who exchange about 300 messages per month.
 
-[Subscribe](https://sympa.inria.fr/sympa/subscribe/caml-list) | [Inria
-Archives](https://sympa.inria.fr/sympa/arc/caml-list) | [Google Groups
-Archives](http://groups.google.com/groups?group=fa.caml)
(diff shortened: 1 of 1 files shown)


dbuenzli:
Le samedi, 16 juillet 2016 à 01:12, Gabriel Scherer a écrit :

> I'm tempted to remove caml-annouce and lists.ocaml.org (http://lists.ocaml.org) and replace them by the IRC channel and, for example, Caml Weekly News

FWIW I thought exactly the same the other day when I had a look at this page.  

Daniel

PR #817, opened 2016-08-04
OCaml 2016: add more preprint links

gasche:


Chris00:
Merged as commit c45eb03.

PR #822, opened 2016-08-17
Added Haxe to success stories

fulopm:


pw374:
thanks

fulopm:
@pw374, there is a typo in my commit. The URL format is not correct in the beginning of the text. Should I make an another PR correcting this?

pw374:
@fulopm thanks, I've just fixed it. (https://github.com/ocaml/ocaml.org/commit/e4151ab0cc7c8dba922e586a1e0c4de22a033829)

PR #823, opened 2016-08-18
update for gentoo ebuild

antranigv:
update gentoo instructions to use complete ebuild name to bypass conflict with app-xemacs/ocaml.


pw374:
thanks

PR #825, opened 2016-08-30
Add Bash on Windows solution for Windows.

dannywillems:
I use it since a month and everything is ok. No issue for the moment (only with IPV6 with ocsigenserver but a solution was found).


pw374:
thanks

PR #828, opened 2016-09-03
Change the license

kryptine:
ocaml/ocaml Apparently INRIA has recently dumped the Q Public License for OCaml and switch to LGPL 2.1 for everything, including the compiler and libraries

### DIFF
--- a/site/docs/license.fr.md
+++ b/site/docs/license.fr.md
@@ -3,60 +3,12 @@
 # La licence d'OCaml
 
 <p>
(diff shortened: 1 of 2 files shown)


avsm:
Is it worth mentioning on this page that the LGPLv2 holds since OCaml 4.03 and above? It might also be worth linking to the [Consortium License](https://ocaml.org/docs/consortium-license.html) from here to inform people that a BSD-license alternative is also available as a membership perk.

dsheets:
Fixes #830.

dbuenzli:
Is there any reason why this wasn't merged ?

PR #829, opened 2016-09-13
Make the install page reasonable (closes #256).

### LINKED ISSUE #256: Download/Install page design feedback
from @edwintorok at https://github.com/ocamllabs/sandbox-ocaml.org/issues/20

---

In general I like the landing page redesign, but there are some aspects that I liked better on the old site, the install page is one of them.

Clicking on "Download OCaml" should take me to http://ocaml-redesign.github.io/docs/install directly, instead of jumping through the platform. TBH I like the old http://ocaml.org/releases/ page better than the new design.

On the install page I'd prefer to have the information from http://ocaml.org/releases/,
perhaps structured with menus like on http://ocaml.org/install.html.
The default tab could be 'Source', and it should show how to download and install the latest _stable_ release from source (4.01.0 now).
The release notes should be visible in the common area, and binary packages accessible from menus or links.
It is fine if instructions on accessing the SVN (or github mirror) is hidden behind a menu / link, I think it is more important to advertise the stable source tarball.

Also it'd be nice if a checksum of the source tarball would be available (even if not digitally signed, but of course a digital signature would be even better).

dbuenzli:
The only thing I didn't clean are the Windows instructions since I don't know enough about what are resonable instructions.

I didn't try to build the website. A basic rendering of the new page can be seen here: 
https://github.com/dbuenzli/ocaml.org/blob/master/site/docs/install.md

### DIFF
--- a/site/docs/install.md
+++ b/site/docs/install.md
@@ -8,576 +8,197 @@ The latest version of OCaml is **{{! get LATEST_OCAML_VERSION !}}**.
 For more information about this release, see the
 [{{! get LATEST_OCAML_VERSION !}} page](/releases/{{! get LATEST_OCAML_VERSION !}}.html).
 
-The OCaml compiler and libraries can be installed in several
-ways. Broadly, the options are:
+The OCaml compiler and libraries can be installed in several ways:
 
-* use OPAM, a package manager specific to OCaml
-* use a package manager supported by your platform (Windows, Linux,
-  Mac OS X, ...)
-* install from source code.
+* With [OPAM](#OPAM), the OCaml package manager (recommended).
+* With a system package manager supported by your platform
+  ([FreeBSD](#FreeBSD), [Linux](#Linux), [macOS](#macOS),
+   [OpenBSD](#OpenBSD), [Windows](#Window))
+* In your [browser](#Browser).
+* By compiling [the source](#FromSource).
 
-The following sections explain how to use each of these methods.
-For additional information about finding and installing OCaml libraries,
-see the [libraries](/learn/libraries.html) page. Information about the different versions of OCaml is available on the [releases](/releases/) page.
+For finding and installing OCaml libraries, see the
+[libraries page](/learn/libraries.html). Information about the
+different versions of OCaml is available on the [releases](/releases/)
+page.
 
-## Multi-Platform Package Managers
+## OPAM
 
-Some dedicated package managers are available for OCaml, that can be used
-on multiple platforms:
+[OPAM](https://opam.ocaml.org/) is the package manager for OCaml. It is
+the recommended way to install the OCaml compiler and OCaml
+packages. Simply follow the
+[OPAM install instructions](https://opam.ocaml.org/doc/Install.html).
 
-### OPAM
+## [FreeBSD](https://www.freebsd.org/)
 
-[OPAM](https://opam.ocaml.org/) is a package manager for OCaml. It is the recommended method to install the OCaml compiler and OCaml
-packages. The default OPAM
-[repository](https://opam.ocaml.org/packages/) is actively maintained
-and serves as the de facto master set of OCaml packages. OPAM also
-provides features that would be unavailable with the other methods,
-such as the ability to have multiple versions of OCaml installed at
-the same time, or have multiple universes of packages for different
-projects with conflicting package dependencies.
-
-Note that there is a bootstrapping issue, since OPAM is itself
-implemented in OCaml. To address this, binary OPAM packages are made
-available on many platforms. See
-[here](http://software.opensuse.org/download.html?project=home%3Aocaml&package=opam)
-and below, in the platform specific sections, for installation
-instructions. Alternatively, you can compile from source by grabbing
-one of the recent
-[releases](https://github.com/ocaml/opam/releases). More information
-is available on the
-[Install pages](http://opam.ocaml.org/doc/Install.html).
-
-Once you have OPAM installed, the `switch` and `install` commands are
-the ones you're likely to need first. For example:
-
-```
-opam switch 4.02.1
-eval `opam config env`
+```bash
+pkg_add -r ocaml-nox11 # If you don't want X11 support 
+pkg_add -r ocaml
 ```
 
-will compile OCaml 4.02.1 and install it in a location managed by
-OPAM. The second line assures you are now using this version of the
-compiler. You can skip this step if you're happy with the version of
-OCaml available on your system , but this lags substantially on many
-platforms.
+## Linux
 
-Next you can install packages, e.g.
+Most Linux distributions allow OCaml and/or OPAM to be installed
+directly through the system package manager.
 
-```
-opam install batteries core
-```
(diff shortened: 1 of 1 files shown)


site/docs/install.md
```diff
@@ -8,255 +8,144 @@ The latest version of OCaml is **{{! get LATEST_OCAML_VERSION !}}**.
...
-* `libpcre-ocaml-dev` which is a binding to PCRE.
-
-Check the [packages available in
-Debian](http://packages.debian.org/search?keywords=ocaml&searchon=all&suite=testing&section=all).
-
-### Ubuntu
+Other Debian packages related to OCaml are
+[listed here](http://packages.debian.org/search?keywords=ocaml&searchon=all&suite=testing&section=all).
+They are made available through
+[The OCaml debian packaging policy](http://pkg-ocaml-maint.alioth.debian.org/ocaml_packaging_policy.html/index.html).
```
avsm: It's the "Debian OCaml packaging policy" I think

site/docs/install.md
```diff
@@ -8,255 +8,144 @@ The latest version of OCaml is **{{! get LATEST_OCAML_VERSION !}}**.
...
 
-[Ubuntu](http://www.ubuntu.com/) is a Debian derivative and uses the same
-package manager, so Debian instructions are applicable to it too.
+### [Ubuntu](http://www.ubuntu.com/)
 
-You can view the list of OCaml packages in Ubuntu repositories
-[here](http://packages.ubuntu.com/search?keywords=ocaml).
+[Ubuntu](http://www.ubuntu.com/) is a Debian derivative so the
+[Debian instructions](#Debian) are applicable. Other Ubuntu packages related
```
avsm: How are the Debian instructions applicable?  The packages don't directly work (or at least, it's not recommended to mix binary packages between Ubuntu and Debian).  It would be shorter to recommend source installation via OPAM perhaps.

FWIW, I do have Debian Jessie backport Dockerfiles here: https://github.com/avsm/debian-jessie-backports-opam that generate Debs of OPAM.  I could look into compiling regular releases of these for Ubuntu as well.

site/docs/install.md
```diff
@@ -8,255 +8,144 @@ The latest version of OCaml is **{{! get LATEST_OCAML_VERSION !}}**.
...
 
-Under Windows, two native solutions are available to use OCaml: the official
-OCaml distribution, and OCamlPro's OCPWin.
-Another solution is to use « Bash on Windows » and install OCaml as on Ubuntu. Main OCaml versions and OPAM work. You can use OCaml as you are on Ubuntu. The disadvantage of this method is you don't build native Windows binaries but Linux binaries.
+Under Windows, two native solutions are available to use OCaml: the
+official OCaml distribution, and OCamlPro's OCPWin.  Another solution
+is to use « Bash on Windows » and install OCaml as on Ubuntu. Main
+OCaml versions and OPAM work. You can use OCaml as you are on
+Ubuntu. The disadvantage of this method is you don't build native
+Windows binaries but Linux binaries.
```
avsm: There is also http://protz.github.io/ocaml-installer/ which is 4.02.3 (unlike OCPWin, which still seems to be 4.01.0)

site/docs/install.md
```diff
@@ -305,279 +194,23 @@ installers and zip archives are provided for OCaml 4.01.0 and 4.02.1. Some addit
...
+The following pages allow to directly try snippets of
+OCaml in your browser:
 
-[Inria stopped supporting MacOS binaries for OCaml 4.02.X](http://caml.inria.fr/download.en.html).
-For earlier versions of OCaml  Intel disk image images follow the [download](../releases/) link.
+* [Try OCaml](https://try.ocamlpro.com/) by OCamlPRO.
+* An
+  [OCaml toplevel](http://ocsigen.org/js_of_ocaml/dev/files/toplevel/index.html)
+  compiled by the [`js_of_ocaml`](http://ocsigen.org/js_of_ocaml/)
+  project.
```
avsm, line 192: We could also link to iocamljs here: https://andrewray.github.io/iocamljs/

site/docs/install.md
```diff
@@ -8,255 +8,144 @@ The latest version of OCaml is **{{! get LATEST_OCAML_VERSION !}}**.
...
 
-### ODB
-See [here](https://github.com/thelema/odb).
-
-### ocamlbrew
-See [here](https://github.com/hcarty/ocamlbrew).
+```bash
+pkg_add -r ocaml-nox11 # If you don't want X11 support 
+pkg_add -r ocaml
+```
```
avsm, line 38: OpenBSD has also had support for ages.  The instructions are:

## [OpenBSD](https://www.openbsd.org)

``` bash
pkg_add ocaml opam
```

avsm (approved):
Thanks very much for taking the time to refresh this page!

site/docs/install.md
dbuenzli: Will add this.

site/docs/install.md
dbuenzli: > How are the Debian instructions applicable? The packages don't directly work (or at least, it's not recommended to mix binary packages between Ubuntu and Debian).

This is what was written on the previous page. But WDYM, why wouldn't `apt-get install ocaml` work ?

dbuenzli: In any case I'll simply repeat the instructions since the person who wants to install ocaml doeesn't give a shit that Ubuntu is a Debian derivative.

site/docs/install.md
dbuenzli: This is the one that is linked under "Official OCaml distribution in fact".

dbuenzli:
Amended the commit with @avsm comments. 

Finally went through the windows section aswell: Removed a lot of things since the details mentioned can be obtained at the link endpoints. I also added a link to @fdopen's experimental opam port since this is what worked best for me at some point. I removed the link to `wodi` since the page says that it's no longer supported. 

I'm sure the windows stuff could still be improved and trimmed. I somehow doubt about the "Use « Bash on Windows » and install OCaml as on Ubuntu.". I left it since I don't know what it means, but I highly suspect that bash on windows doesn't give you `apt-get`...

site/docs/install.md
avsm: Yes indeed, the background information on Ubuntu is pretty irrelevant here

site/docs/install.md
avsm: It's not really clear to me what "official" means in this capacity, since it's not generated by the INRIA release manager (@damiendoligez).  It's also a major release out of date -- @msprotz, any thoughts on how we should refer to your distribution in the Windows install section?

dbuenzli:
Is there still something that needs to be done for this to be merged ? The current install page makes me physically ill.

pw374:
Thanks.

site/docs/install.md
msprotz: Nope... maybe "trusted", or "partially maintained"? I'm in kind of a best effort situation right now... we talk about it at the ML workshop. I'm around Nara.

PR #837, opened 2016-09-25
Add a tail recursive solution to question 22

zzhjerry:
The title of the problem is: Create a list containing all integers within a given range.

A tail recursive `range` function can generate a larger range. i.e Using the new implementation can generate result from `1` to `9999999` but using the given implementation to generate such a range will cause `Stack overflow during evaluation (looping recursion?).`

### DIFF
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -502,6 +502,16 @@ SOLUTION
 >     if a > b then [] else a :: aux (a+1) b  in
 >   if a > b then List.rev (aux b a) else aux a b;;
 > ```
+>
+> A tail recursive implementation:
+>
+> ```ocamltop
+> let range a b =
+>   let rec aux acc high low =
+>     if high >= low then
+>       aux (high::acc) (high - 1) low else acc
+>   in if a < b then aux [] b a else List.rev (aux [] a b)
+> ```
 
 ```ocamltop
 range 4 9;;
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -509,7 +509,8 @@ SOLUTION
 > let range a b =
 >   let rec aux acc high low =
 >     if high >= low then
->       aux (high::acc) (high - 1) low else acc
+>       aux (high::acc) (high - 1) low
+>     else acc
 >   in if a < b then aux [] b a else List.rev (aux [] a b)
 > ```
 


site/learn/tutorials/99problems.md
```diff
@@ -502,6 +502,16 @@ SOLUTION
...
 >   if a > b then List.rev (aux b a) else aux a b;;
 > ```
+>
+> A tail recursive implementation:
+>
+> ```ocamltop
+> let range a b =
+>   let rec aux acc high low =
+>     if high >= low then
+>       aux (high::acc) (high - 1) low else acc
```
pw374: if there's a newline after `then`, there should be at least a newline before `else`, otherwise visually it feels like there's no `else`branch.

pw374:
Thanks.

I'm merging because your additional implementation gives the same results as the existing one.

But I suspect either the implementation is wrong or there's a typo in

```
 If first argument is smaller than second, produce a list in decreasing order.
```

I mean, shouldn't it be 

```
 If first argument is *greater* than second, produce a list in decreasing order.
```

??

site/learn/tutorials/99problems.md
```diff
@@ -502,6 +502,17 @@ SOLUTION
...
+>
+> A tail recursive implementation:
+>
+> ```ocamltop
+> let range a b =
+>   let rec aux acc high low =
+>     if high >= low then
+>       aux (high::acc) (high - 1) low
+>     else acc
+>   in if a < b then aux [] b a else List.rev (aux [] a b)
```
Chris00, line 514: This is not the most efficient when a >= b.

Chris00:
@pw374 Yes, this is a typo.  May you please fix it?

site/learn/tutorials/99problems.md
pw374: It's not but making it efficient means duplicating some code or adding quite some complexity in the style of the code. Also, that List.rev was already used in the other implementation.

pw374:
> @pw374 Yes, this is a typo. May you please fix it?

done :)

Chris00:
On 2016-10-02, Philippe Wang wrote:

> ```
> @pw374 Yes, this is a typo. May you please fix it?
> ```
> 
> done :)

Thanks.

Chris00:
On 2016-10-02, Philippe Wang wrote:

> @pw374 commented on this pull request.
> 
> ---
> 
> In site/learn/tutorials/99problems.md:
> 
> > @@ -502,6 +502,17 @@ SOLUTION
> >     if a > b then [] else a :: aux (a+1) b  in
> >   if a > b then List.rev (aux b a) else aux a b;;
> > 
> > ``````
> > +>
> > +> A tail recursive implementation:
> > +>
> > +> ```ocamltop
> > +> let range a b =
> > +>   let rec aux acc high low =
> > +>     if high >= low then
> > +>       aux (high::acc) (high - 1) low
> > +>     else acc
> > +>   in if a < b then aux [] b a else List.rev (aux [] a b)
> > ``````
> 
> It's not but making it efficient means duplicating some code or adding 
> quite
> some complexity in the style of the code. Also, that List.rev was 
> already used
> in the other implementation.

Right.  Maybe one should write a comment about this so the reader does 
not think it is the best way.  That said, passing "succ" or "pred" as an 
additional argument avoids the duplication.

PR #848, opened 2016-10-12
Added  English number words and Sudoku solution

Theomat:


site/learn/tutorials/99problems.md
```diff
@@ -2604,10 +2625,139 @@ between 1 and 9. The problem is to fill the missing spots with digits in
...
+ | Num of int (* Values given *)
+ | Var of int;; (* Values we try *)
+type grid = cell array array;;
+```
 
-```ocaml
-(* example pending *);;
+SOLUTION
+> The idea is to start filling with available values in each case.
+> When there is no available valeus, it means we made a mistacke so we go back to where we had to make a choice between different available values and choose the next one.
```
pw374: mistacke  -> mistake

Theomat: Thanks for pointing that out, i'll correct the different typos.

Chris00:
"English number words" solution in bc57e20dc212b7d57c639acbac2a73c7605d11a0.

Chris00:
Integrated in 075b5587c58e722ec4d0561e43f91e214dfb7578 and then modified.

PR #851, opened 2016-10-21
Add Arch Linux installation instructions

juergenhoetzel:
I maintain the OCaml packages für Arch Linux. It would be nice to mention OCaml  is available in our official repository.


pw374:
thanks!

PR #855, opened 2016-11-07
Fix french translation of "Tutorials"

redianthus:
Hi,

This PR fix the french translation of `Tutorials` which should be `Tutoriels` instead of `Tutoriaux` - see [tutoriels](https://fr.wiktionary.org/wiki/tutoriels) and [tutoriaux](https://fr.wiktionary.org/wiki/tutoriaux) from wiktionary.


agarwal:
I might be misunderstanding the wiktionary definitions, but it seems both are valid. I'll let a French speaker merge this request.

Chris00:
Both are correct but _tutoriaux_ is seen as an outdated version of _tutoriels_.  I'm fine with the change.

PR #873, opened 2017-02-02
Add anchors to FAQ page so they are externally linkable

tekknolagi:
This makes it easy for people to link to specific official explanations of topics like arbitrary precision arithmetic.

### DIFF
--- a/site/learn/faq.md
+++ b/site/learn/faq.md
@@ -7,7 +7,7 @@ which is widely used by the OCaml community.
 *Table of contents*
 
 ## General Questions
-#### What is OCaml?
+#### <a name="what-is-ocaml"></a> What is OCaml?
 
 OCaml is a programming language. It is a functional language, since the
 basic units of programs are functions. It is a strongly-typed language;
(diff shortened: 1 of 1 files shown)


Chris00:
The HTML generated from Markdown already define anchors from the title, for example for [arbitrary precision arithmetic](https://ocaml.org/learn/faq.html#Isitpossibletodocomputationswitharbritraryprecisionarithmetics).  Maybe it is too much work to retrieve the URL and a small "copy this link' icon would be welcome?

tekknolagi:
Oh! I did not realize that. Wow, I really should have checked. I think that is a great idea. Where is the template that I should modify to add that?

Chris00:
I'm afraid modifying a template won't be enough.  It can possibly made part of [`omd`](https://github.com/ocaml/omd) through an option (@pw374, any thoughts?) or achieved by post-processing.

pwbs:
So, we want to generate a button with a "copy this link" function? It shouldn't be hard to do...

Chris00:
@pwbs Any development on this?

XVilka:
@tekknolagi ping?

avsm:
I'll go ahead and close this as it's not progressing, and there is a new version of omd out (#1321), so this could be addressed in a future iteration of the site.

PR #901, opened 2017-05-31
Update /meetings/index.fr.md

pwbs:


Chris00:
When you merge, would you please not create merge commits — they are useless here (for a single commit) and clutter history.

pw374:
My apologies.
I'll try to keep that in mind.
Every place has its own policy, it can be hard to keep track.

Chris00:
On 2017-05-31, Philippe Wang wrote:
> 
> My apologies.

No need to apologize.  IMHO, merging a separate branch is useful when 
there is some development along it so the changes are grouped.  I do not 
see the point for a single commit.

PR #903, opened 2017-05-31
Travis: Pin opam2web for the compilation to succeed

Chris00:


Chris00:
Argh, I meant to see whether this was fixing the builds...

Chris00:
The docker cloud is useless to me as I cannot see the error (login required).  @avsm ?

pw374:
I thought it'd be useful to merge it to see if it's indeed fixing the builds, since if it was broken, it won't do much harm and worst case scenario we can revert...

Chris00:
On 2017-05-31, Philippe Wang wrote:
>
> I thought it'd be useful to merge it to see if it's indeed fixing the 
> builds, since if it was broken, it won't do much harm and worst case 
> scenario we can revert...

I restarted the Travis build and it was indeed the culprit.  So it was 
fine! ☺

PR #906, opened 2017-06-07
Delete a sentence that has already been translated

fylimas:
An English sentence which has already been translated into Chinese is deleted.


Chris00:
Thanks.

PR #918, opened 2017-07-03
add solution to "English Number Words" challenge

chrismamo1:
this might not be an ideal solution to the problem, but it's pretty good and seems to work

### DIFF
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -2548,11 +2548,41 @@ in full words. Example: 175 must be written as one-seven-five. Write a
 function `full_words` to print (non-negative) integer numbers in full
 words.
 
-<!-- SOLUTION -->
+SOLUTION
 
-```ocaml
-(* example pending *);;
-```
+> A recursive solution that should work in `O(log(n))` space and time (or `O(n)` if you take `n` to be the number of digits base 10 of the input).
+> ```ocamltop
+> let full_words n =
+>   let rec aux n =
+>     let d = n mod 10 in
+>     let d' = match d with
+>       | 0 -> "zero"
+>       | 1 -> "one"
+>       | 2 -> "two"
+>       | 3 -> "three"
+>       | 4 -> "four"
+>       | 5 -> "five"
+>       | 6 -> "six"
+>       | 7 -> "seven"
+>       | 8 -> "eight"
+>       (* d < 10, so the catchall case is just here for the compiler *)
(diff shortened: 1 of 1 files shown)


site/learn/tutorials/99problems.md
```diff
@@ -2548,11 +2548,41 @@ in full words. Example: 175 must be written as one-seven-five. Write a
...
-<!-- SOLUTION -->
+SOLUTION
 
-```ocaml
-(* example pending *);;
-```
+> A recursive solution that should work in `O(log(n))` space and time (or `O(n)` if you take `n` to be the number of digits base 10 of the input).
+> ```ocamltop
+> let full_words n =
+>   let rec aux n =
```
Chris00, line 2556: It would be better to use a more descriptive name than `aux`.

site/learn/tutorials/99problems.md
```diff
@@ -2548,11 +2548,41 @@ in full words. Example: 175 must be written as one-seven-five. Write a
...
+>       | 6 -> "six"
+>       | 7 -> "seven"
+>       | 8 -> "eight"
+>       (* d < 10, so the catchall case is just here for the compiler *)
+>       | 9 | _ -> "nine"
+>    in
+>    let n' = n / 10 in
+>    if n' = 0
+>      then begin
+>        let buf = Buffer.create 64 in
```
Chris00, line 2574: I think it would be clearer to create the buffer outside of the recursive function and have this function fill the buffer.

site/learn/tutorials/99problems.md
```diff
@@ -2548,11 +2548,41 @@ in full words. Example: 175 must be written as one-seven-five. Write a
...
+>        Buffer.add_string buf d';
+>        buf
+>     end
+>     else begin
+>       let buf = aux n' in
```
Chris00, line 2579: Not so nice that it is not terminal recursive.  How about creating a list of the strings (which will naturally be in the right order) and then concatenating them (`String.concat`) with `"-"` in between?

Chris00 (changes requested):
Thanks for your contribution.  May you please address the following remarks?

Chris00:
I wrote a solution in bc57e20dc212b7d57c639acbac2a73c7605d11a0.

chrismamo1:
Thanks for the feedback! Sorry I didn't see it soon enough to address the issues (and it looks like a solution already existed in another PR, I should've checked there first).

PR #919, opened 2017-07-08
Tutorial, structure of ocaml programs: remove the emphasis on `;;`

Octachron:
This PR proposes to remove the unfortunate emphasis on the difficulty to use `;;` in [this tutorial](https://ocaml.org/learn/tutorials/structure_of_ocaml_programs.html#Usingandomittingand).
Instead, this PR presents first the sequencing operator ";" and only after discuss the question of ";;".
The whole section has been rewritten to focus on the fact that `;;` is mostly used in the toplevel and in tutorials, and that there are entire code bases which never use `;;`.

### DIFF
--- a/site/learn/tutorials/structure_of_ocaml_programs.md
+++ b/site/learn/tutorials/structure_of_ocaml_programs.md
@@ -363,7 +363,7 @@ Note especially the last one - I'm using `;` as an operator to "join"
 two statements. All functions in OCaml can be expressed as:
 
 ```ocaml
- let name [parameters] = expression
+ let name [parameters] = expression ;;
 ```
 OCaml's definition of what is an expression is just a little wider
 than C's. In fact, C has the concept of "statements"- but all of C's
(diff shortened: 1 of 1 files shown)


site/learn/tutorials/structure_of_ocaml_programs.md
```diff
@@ -335,145 +335,119 @@ Actually this is really useful when you want to import a nested module
...
+ let f x b y = if b then x+y else x+0
+ let f x b y = x + (if b then y else 0)
+ let f x b y = x + (match b with true -> y | false -> 0)
+ let f x b y = x + (let g z = function true -> z | false -> 0 in g y b)
+ let f x b y = x + (let _ = y + 3 in (); if b then y else 0)
+ let f = fun x -> fun b -> fun y -> if b then x+y else x+0
+```
 
-Have a look at a section from the second graphics example above:
+Note especially the last one - I'm using `;` as an operator to "join"
```
hcarty: It's not the last but second to last here. The last example doesn't use `;`

site/learn/tutorials/structure_of_ocaml_programs.md
```diff
@@ -335,145 +335,119 @@ Actually this is really useful when you want to import a nested module
...
+## The disappearance of `;;`
 
-Rule #2 is that *sometimes* you can elide the `;;`. As a
-beginner you shouldn't worry about this — you should always put in the
-`;;` as directed by Rule #1. But since you'll also be reading a lot of
-other peoples' code you'll need to know that sometimes we can elide
-`;;`. The particular places where this is allowed are:
+Now we're going to look at a very important issue. All examples above
+ended with a double semi-colon `;;`. However, if you look at OCaml code
+outside of tutorials, you will find whole code base that does not
```
hcarty: `s/code base/code bases/`

site/learn/tutorials/structure_of_ocaml_programs.md
```diff
@@ -335,145 +335,119 @@ Actually this is really useful when you want to import a nested module
...
 
-* Before the keyword `let`.
-* Before the keyword `open`.
-* Before the keyword `type`.
-* At the very end of the file.
-* A few other (very rare) places where OCaml can "guess" that the next
- thing is the start of a new statement and not the continuation of
- the current statement.
+The truth is that `;;` is mostly used in the toplevel and tutorials to
+mark the end of an Ocaml phrase and send this phrase to the toplevel
```
hcarty: `s/Ocaml/OCaml/`

site/learn/tutorials/structure_of_ocaml_programs.md
```diff
@@ -335,145 +335,119 @@ Actually this is really useful when you want to import a nested module
...
-> `+` just like a function. For instance, I can define a `sum_list`
-> function, to sum a list of ints, like:
-> 
-> ```ocamltop
-> let sum_list = List.fold_left ( + ) 0
-> ```
+or if the result of the expression is `unit`
+
+```ocaml
+let () = expression ()
```
hcarty: Maybe better to introduce this first, since `_` can silently eat coding errors like unintentional partial application.

site/learn/tutorials/structure_of_ocaml_programs.md
```diff
@@ -335,145 +335,119 @@ Actually this is really useful when you want to import a nested module
...
+```
+
+With this convention, there are no toplevel expressions anymore: any
+module can be written as a sequence of definitions. Consequently, some
+style guidelines consider that `;;` should never be used outside of the
```
hcarty: A link to the style guide on ocaml.org might be useful here so a reader can jump directly to the somewhat official guide.

hcarty (commented):
This seems clearer!  I just found some typos and small suggestions.

site/learn/tutorials/structure_of_ocaml_programs.md
Octachron: I forgot that I introduced another example, I fixed this by swapping these two examples.

site/learn/tutorials/structure_of_ocaml_programs.md
Octachron: Fixed, thanks.

site/learn/tutorials/structure_of_ocaml_programs.md
Octachron: I switched the two constructs and added a short paragraph to explain why "let () = …" is generally better.

site/learn/tutorials/structure_of_ocaml_programs.md
Octachron: Link added (even if the somewhat official guide is not cristal clear on the use of `;;`), thanks for the suggestion.

Octachron:
@hcarty Thanks for the suggestions! (and the spotted typos).

Chris00:
Thanks.

PR #921, opened 2017-07-13
update links for downloading opam_install.sh on french documentation

YiuRULE:
Hello,

I updated the french documentation for the installation of ocaml w/ opam available on [this link](https://ocaml.org/docs/install.fr.html), because when we execute the script, we receive this error:

`opam_installer.sh: Couldn't download http://www.ocamlpro.com/pub/opam-1.1.0-x86_64-Linux`

So changing the download link available on the [english document](https://opam.ocaml.org/doc/Install.html) fix the problem.


Chris00:
Thnaks.

PR #922, opened 2017-07-14
add 4.05.0 release

gasche:
I have not tested the rending (is there an easy way to do this?).

### DIFF
--- a/site/releases/4.05.0.html
+++ b/site/releases/4.05.0.html
@@ -1,0 +1,6 @@
+<!-- ((! set title OCaml 4.05.0 !)) -->
+<html>
+  <head>
+    <meta http-equiv="refresh" content="0; url=4.05.html" />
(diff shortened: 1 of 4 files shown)


agarwal:
> I have not tested the rending (is there an easy way to do this?).

Nothing other than building the site locally, which does require you to have several dependencies installed.

gasche:
Thanks!

As of now, [Install.html](http://ocaml.org/docs/install.html) refers to a 4.05.0 release (if I understand correctly, this is automagically determined from the opam repository, but the link it gives, <http://ocaml.org/releases/4.05.0.html>, is dead. If I understand correctly, this due to the change in this PR not having reached the website.

agarwal:
I'm not sure how the `get LATEST_OCAML_VERSION` is being computed? @Chris00 Do you know?

PR #931, opened 2017-10-11
Removing outdated PPA

superherointj:
Anil, author of ppa:avsm/ppa, reported:
| The PPA is no longer necessary, as Ubuntu 16.04 ships with the latest packages.  Simply try `apt install opam` to get started on Ubuntu.


pw374:
thanks!

PR #942, opened 2017-12-10
More efficient solutions for 61 and 62

roflmaostc:
Removed the inefficient append (@) operator from solutions.

### DIFF
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -1627,6 +1627,10 @@ collect them in a list.
 SOLUTION
 
 > ```ocamltop
+> (* Having an accumulator acc prevents using inefficient List.append.
+>  * Every Leaf will be pushed directly into accumulator.
+>  * Not tail-recursive, but that is no problem since we have a binary tree and
+>  * and stack depth is logarithmic. *)
 > let leaves t = 
 >   let rec leaves_aux t acc = match t with
 >     | Empty -> acc
@@ -1650,6 +1654,10 @@ successors. Write a function `internals` to collect them in a list.
 SOLUTION
 
 > ```ocamltop
+> (* Having an accumulator acc prevents using inefficient List.append.
+>  * Every internal node will be pushed directly into accumulator.
+>  * Not tail-recursive, but that is no problem since we have a binary tree and
+>  * and stack depth is logarithmic. *)
 > let internals t = 
 >   let rec internals_aux t acc = match t with
 >     | Empty -> acc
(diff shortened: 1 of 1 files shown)


site/learn/tutorials/99problems.md
```diff
@@ -1627,10 +1627,13 @@ collect them in a list.
 SOLUTION
 
 > ```ocamltop
-> let rec leaves = function
->   | Empty -> []
->   | Node(c, Empty, Empty) -> [c]
->   | Node(_, l, r) -> leaves l @ leaves r
+> let leaves t = 
+>   let rec leaves_aux t acc = match t with
```
Chris00: Note that, since `leaves_aux` is not tail rec anyway, why not use `List.rev_append` (with a note saying why)?

site/learn/tutorials/99problems.md
```diff
@@ -1647,9 +1650,13 @@ successors. Write a function `internals` to collect them in a list.
 SOLUTION
 
 > ```ocamltop
-> let rec internals = function
->   | Empty | Node(_, Empty, Empty) -> []
->   | Node(c, l, r) -> internals l @ (c :: internals r)
```
Chris00, line 1652: Idem.

site/learn/tutorials/99problems.md
```diff
@@ -1668,11 +1675,16 @@ list.
 SOLUTION
 
 > ```ocamltop
-> let rec at_level t l = match t with
->   | Empty -> []
->   | Node(c, left, right) ->
->      if l = 1 then [c]
->      else at_level left (l - 1) @ at_level right (l - 1)
```
Chris00, line 1675: Idem.

Chris00 (changes requested):

Chris00:
Please rebase your commits on trunk and squash them — will do it while merging otherwise.

site/learn/tutorials/99problems.md
roflmaostc: List.rev_append is still inefficient. Very often it will traverse the output list to reach the end.. 
Using List.rev_append would result in a O(n^2).
Here the stack depth is only in log(m) since we have a binary tree. Runtime of my solution is O(n).

roflmaostc:
I tried to do what you wanted, but I think the result is wrong.
I'm not so familiar with git to fix some curious situations...

roflmaostc:
I think now it is what you want?

Chris00:
For the squashing, yes.  Still looking forward to the answer about using `List.rev_append` instead.

roflmaostc:
Commented above:
List.rev_append is still inefficient. Very often it will traverse the output list to reach the end..
Using List.rev_append would result in a O(n^2).
Here the stack depth is only in log(n) since we have a binary tree. Runtime of my solution is O(n).

Chris00:
OK — looked to fast.  May you add comments in the code (or before the code block) to motivate your choice of implementation?  Thanks.

roflmaostc:
Added some comments

Chris00:
Thanks.

PR #947, opened 2018-01-16
Tutorial: 99 problems : simpler non tail-recursive implementation for problems 7 and 12

christophe-riolo:
The proposed solution does not use List.rev. Unfortunately, there is no tail recursion in the `One` case.

### DIFF
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -162,7 +162,18 @@ type 'a node =
 SOLUTION
 
 > ```ocamltop
-> (* This function traverses the list, prepending any encountered elements
+> (* This naive implementation traverses the list, storing the tail of the 
+> result list in an accumulator*)
+> 
+> let flatten_naive list =
+>   let rec aux acc = function
+>     | [] -> acc
+>     | One x::tl -> x::(aux acc tl)
+>     | Many l::tl -> aux (aux acc tl) l
+>     in aux [] list
+>
+> (* This version however is more efficient by being tail recursive.
+>   It traverses the list, prepending any encountered elements
 >   to an accumulator, which flattens the list in inverse order. It can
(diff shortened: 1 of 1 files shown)


Chris00:
>  there is no tail recursion in the One case.

An none of the cases in the `decode` function is tail recursive (b/c `c::decode`).  I think we should keep the tail-recursiveness of the solution as it is definitely something programmers should think about.  Do you think we should motivate why that style was chosen?

christophe-riolo:
I think it would be good, at least as a general disclaimer, to explain how the solutions have been chosen. I try to put myself in a learner's shoes, it might seem like "ocaml is difficult" if they are not told that there are simpler but less efficient implementations :)
So explaining that there has been a choice of "efficient and good practice" as opposed to "naive and simple" could be good, though I would argue that ideally, from the learner's point of view, seeing both could be interesting .

Chris00:
Would you update this PR with better explanations for the given problem?  I wary a little about a general declaration because solutions are not all optimized.

christophe-riolo:
Sure, I'll take a look at how to formulate it!

christophe-riolo:
I have updated, putting the original algorithms back, with some additional text about tail calls.

pmetzger:
Just an aside: The name of this PR ticket makes it difficult for someone looking through a list of pull requests to know what this is about (and thus whether they should pay attention). I would have named it something like "Tutorial: More straightforward solution to problem 7".

Similarly, the commit messages don't say a lot for someone scanning "git log" about whether they're interested in looking at a particular commit or not.

christophe-riolo:
@pmetzger Thank you for your input, I am indeed new to contributing so that was a good idea.
I have cleaned and reworded my commits and renamed the PR

christophe-riolo:
I will close this PR and try to go through all the exercises to find if I can do other similar simpler naive versions, and submit a more complete PR.

PR #948, opened 2018-01-27
No longer recommend Protz's Windows installer

vp2177:
See http://protz.github.io/ocaml-installer/ as for why.


Chris00:
Thanks.

PR #956, opened 2018-02-19
Added solution for problem 85.

bruderjakob17:


site/learn/tutorials/99problems.md
```diff
@@ -2483,11 +2483,55 @@ adjacent if and only if f(X) and f(Y) are adjacent.
...
-<!-- SOLUTION -->
+SOLUTION
 
-```ocaml
-(* solution pending *);;
-```
+> ```ocamltop
+> (* calculates the cross-product l1 x l2 of two lists *)
+> let cross_product l1 l2 =
+> let rec aux acc = function
```
Chris00, line 2491: 1. Properly indent your code
2. Try to use a meaningful name for inner functions.

site/learn/tutorials/99problems.md
```diff
@@ -2483,11 +2483,55 @@ adjacent if and only if f(X) and f(Y) are adjacent.
...
+>   | [] -> acc
+>   | x::xs -> aux (List.rev_append
+>     (List.combine (List.init (List.length l2) (fun _ -> x)) l2)
```
Chris00, line 2494: You do not need to create a list full of `x` to combine it with `l2`, a simple fold is enough.

site/learn/tutorials/99problems.md
```diff
@@ -2483,11 +2483,55 @@ adjacent if and only if f(X) and f(Y) are adjacent.
...
+>     acc) xs
+> in aux [] l1
+>
+> (* check whether the function f (given by an association list) is an isomorphism between g and h *)
```
Chris00, line 2498: Note that a map to represent `f` will be much faster — if one assume the data to be comparable.

site/learn/tutorials/99problems.md
```diff
@@ -2483,11 +2483,55 @@ adjacent if and only if f(X) and f(Y) are adjacent.
...
+> let is_graphiso f g h =
+>   if (List.length f <> List.length g.nodes) || (List.length f <> List.length h.nodes) then
```
Chris00, line 2500: Your code does not guarantee that `f` is a bijection.  Maybe it is because you will only feed it with such "functions" but then you can probably ensure that `f` also has the right length.

site/learn/tutorials/99problems.md
```diff
@@ -2483,11 +2483,55 @@ adjacent if and only if f(X) and f(Y) are adjacent.
...
+>     false
+>   else
+>     let check (v,w) =
+>       List.mem (v,w) g.edges = List.mem (List.assoc v f, List.assoc w f) h.edges
```
Chris00, line 2504: Considering that the number of edges is often a magnitude smaller than the number of couples in the cross product `g.nodes × g.nodes`, you should design a faster way of checking edge correspondence.

site/learn/tutorials/99problems.md
```diff
@@ -2483,11 +2483,55 @@ adjacent if and only if f(X) and f(Y) are adjacent.
...
+>     in List.for_all (check) (cross_product g.nodes g.nodes)
+>
+> (* searches a bijection f: setA -> setB such that is_iso f evaluates to true.
+>    If found returns Some f (f as association list), else None.
+>    This algorithm checks every possible bijection, thus is very slow. *)
+> let find_iso_opt is_iso setA setB =
+>   let rec aux acc remainingA remainingB =
+>     match remainingA with
+>     | [] -> if is_iso acc then Some acc else None
```
Chris00, line 2513: Why calling `acc` here what you call `f` above?

site/learn/tutorials/99problems.md
```diff
@@ -2483,11 +2483,55 @@ adjacent if and only if f(X) and f(Y) are adjacent.
...
+>     | x::xs ->
+>       let rec aux2 i =
+>         if i < List.length remainingB then
+>           let y = List.nth remainingB i in
```
Chris00, line 2517: You should avoid repeatedly calling `List.length` and `List.nth`—your algorithm will then have a higher asymptotic cost than it needs to have.

site/learn/tutorials/99problems.md
```diff
@@ -2483,11 +2483,55 @@ adjacent if and only if f(X) and f(Y) are adjacent.
...
+>           let ys = List.filter (fun v -> v <> y) remainingB in
+>           match aux ((x,y)::acc) xs ys with
+>             | Some f -> Some f
+>             | None -> aux2 (i+1)
+>         else None
```
Chris00, line 2522: Your algorithm would be better if it used a set to hold remaining elements.

Chris00 (commented):
Another review will need to be done after the remarks are adequately acted upon.

avsm:
@bruderjakob17 I'm going to close this PR for now, as there seems to be no activity towards addressing the review comments for 2 years. However, please do reopen the PR and continue the work if you get a chance!

PR #958, opened 2018-02-26
Reduce the use of gendered pronouns

vicapow:


agarwal:
This is good, and this [article](https://www.washingtonpost.com/news/morning-mix/wp/2017/03/28/the-singular-gender-neutral-they-added-to-the-associated-press-stylebook/?utm_term=.f0eb4353b8d6) states that `they` is becoming accepted even as a singular, gender-neutral pronoun.

Chris00:
Then, time permitting, the other tutorials should also be checked to ensure uniformity.

PR #960, opened 2018-03-12
corrected graphics code

tarptaeya:
The graphics code was not correct and it is not compilable. According to ocaml (I tried with v>=4.05) `let ()` must return with a value of type `unit`, but the code provided here was using `read_line ()` which is of type `string -> unit`. So I used `print_endline ""` which is of type `unit` to correct the code.


Chris00:
The `read_line` is there so that the graphics window stays open until a key is pressed.  Fixed by mean of 9e1818b

PR #962, opened 2018-03-13
Remove an unnecessary line break before hyphen

yudai-nkt:
I've just started learning OCaml with the tutorial here and Real World OCaml, and noticed that there is an unexpected itemization in the Japanese version unlike the English one.

[English version](https://ocaml.org/learn/tutorials/structure_of_ocaml_programs.html#References-real-variables):

[Japanese version](https://ocaml.org/learn/tutorials/structure_of_ocaml_programs.ja.html):

This PR attempts to fix it by deleting a line break before the hyphen which is incorrectly interpreted as Markdown syntax.

By the way, this hyphen should be an em dash, isn't it?


Chris00:
Thanks.

> By the way, this hyphen should be an em dash, isn't it?

Yes.

PR #963, opened 2018-03-14
Wrong Chinese character

HyperLiar:
It should be '像' but not '象'.


Chris00:
Commit messages **must** be in English.  Please update yours.

Chris00:
Thanks.

PR #966, opened 2018-03-14
corrected spelling

tarptaeya:
`highlight` was incorrectly written as `highligh` in the comment of `highlight_error_range` function


Chris00:
Thanks

PR #971, opened 2018-03-25
Replace deprecated Array.create with Array.make

alan-j-hu:
The `Array.create` function has been deprecated in favor of `Array.make`. The OCaml interpreter displays a warning that appears in the compiled HTML at http://ocaml.org/learn/taste.html.


Chris00:
Thanks.

PR #975, opened 2018-03-31
escaped brackets to make them render correctly

tarptaeya:
**Reference** - Omd (About.md)
> If you want a parenthesis or bracket not to count in the balanced parsing, escape it with a backslash


Chris00:
Thanks.

PR #984, opened 2018-04-17
Remove OCaml 2017 announcement

pmetzger:
OCaml 2017 was a long time ago, so this probably doesn't belong on the front page any longer.

### DIFF
--- a/site/index.md
+++ b/site/index.md
@@ -104,16 +104,6 @@
 			<ul class="news-feed" style="margin-bottom: 0px">
 
 			<li class="announcement"><article>
-			  <h1><a title="OCaml Users and Developers Workshop"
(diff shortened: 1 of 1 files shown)


Chris00:
I would prefer to have it commented out — rather than removed — as one will need to put it back when the next meeting will be announced.

pmetzger:
@Chris00 As you have requested, I've changed this to comment out the announcement.

Chris00:
Thanks.

pmetzger:
BTW, longer term, I'd suggest that it might be good if there was a distinction between "real" news (new versions of OCaml, OCaml meeting announcements, etc.) and the OCaml planet stuff, but that's not a good discussion for here.

PR #990, opened 2018-04-26
Fix warnings

redianthus:
Hi,

I fixed warnings I got when doing `make`.

### DIFF
--- a/script/breadcrumb.ml
+++ b/script/breadcrumb.ml
@@ -11,7 +11,7 @@ let get_title lang path =
     ignore(Str.search_forward title_re s 0); (* or Not_found *)
     String.trim (Str.matched_group 1 s)
   with Not_found ->
-    String.capitalize(Filename.basename path)
+    String.capitalize_ascii(Filename.basename path)
 
 let rec breadcrumb_of_path bc lang path =
   if path = "site" || path = "." || path = "/" then ("/", "Home") :: bc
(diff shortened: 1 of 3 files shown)


redianthus:
Is there a special reason why we are testing against `4.02.3` ?

The build is failing with `Error: Unbound value String.lowercase_ascii` which is not surprising. :)

If we don't need to keep compatibility with `4.02.3` then we can just drop this in the CI file and merge the PR. (I can open a PR to drop `4.02.3` if needed)

agarwal:
I'm not sure what version of OCaml is being run on the build server. @avsm @Chris00 I lost track long ago about how the site is actually built and deployed. Can you confirm if testing against 4.02.3 is needed. Thanks.

pmetzger:
I think it would be best if the build server for the site was running 4.06.1 at this point. It would mean we could stick to modern constructs for everything.

Chris00:
4.02.3 is run on the server — at the time because of problems with dependencies.  I'll try to update it tomorrow.

redianthus:
@Chris00, is there some dependency trouble to fix ? Or is it just a matter of updating OCaml on the server ? Just to know if there's something more I can do. :)

Chris00:
It is updated on the server but there is still a problem with Travis.

redianthus:
Yes, I just have to remove it from Travis. Will do.

EDIT: it appears you already did it :) we just have to wait for the build to pass

redianthus:
@Chris00, I'm not sure what's going on with the `4.05.0` build.

On the [last master build](https://travis-ci.org/ocaml/ocaml.org/jobs/388257176#L949), this seems to be handled correctly. [Here](https://travis-ci.org/ocaml/ocaml.org/jobs/388318777#L933), the build is stopped because there's not output produced. I'm not sure if it's because no answer is provided to opam (why ?) or if it's just because the next step is too long...

Chris00:
There is a problem with the script to retrieve the latest opam package.  Unfortunately, that thing breaks regularly...  I'll try to write something taking another approach — but it is unlikely to happen before July.

redianthus:
I'll try to see if I can fix it. Do you want to wait until it's fixed to merge this ?

PR #994, opened 2018-04-27
Update index.md

ghost:
Add some advice. That's what people visit this site for.

### DIFF
--- a/site/learn/tutorials/index.md
+++ b/site/learn/tutorials/index.md
@@ -3,6 +3,9 @@
 
 # OCaml tutorials
 
++If you are just looking for a place to start you should try
++[Real World OCaml](https://dev.realworldocaml.org)
+
 ####  Your Help is Needed
 Many of the tutorials below need updating and tutorials on many new
 topics are needed. Please contribute by visiting this project's repo on


agarwal:
I don't know if we can promote a single book, and some people feel OCaml From the Very Beginning is better to learn the basics of ocaml. What do you think about generalizing this sentence to something like "If you are looking for books, see the [Books](http://ocaml.org/learn/books.html) page."

pmetzger:
Not sure why you have `+`es around your sentence?

pmetzger:
I think @gutsle's point is that he'd like a concrete piece of advice for what a person at the very beginning should do, listed at the very top. Having a long list of options creates paralysis and indecision in many people.

agarwal:
I'm experiencing paralysis on this PR. I'll be happy if someone else wants to make a decision.
Cc: @Chris00, @amirmc.

Chris00:
I'm also a bit reluctant to promote a single book, especially because it also promote an alternative stdlib and other alternatives could feel aggrieved (even though I like most of the design of Base/Core).  If Base/Core had achieved the same acceptance than Dune, OPAM, and Merlin, I would be more willing to single out that book — it is good, the fact that the community could comment on the 1st version before its launch was a plus and it is available online for free.

agarwal:
Given the hesitancy of two of the maintainers and in the interest of moving on, I'm going to close without merging. Apologies if you disagree.

PR #999, opened 2018-05-21
It was confusing for me, it should be changed?

Neuche:
Hello, i 'm learning Ocaml and the wrong indentation made the example look kind of recursive and confusing. I hope i got it right.


Chris00:
Thanks.

PR #1004, opened 2018-06-05
Improve scripts

redianthus:
Hi,

I just changed  few things in the scripts.

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -1,3 +1,6 @@
+#!/bin/sh
+
+set -ex
 
 export OPAMYES=1
 
@@ -5,7 +8,7 @@ export OPAMYES=1
 if [ -f "$HOME/.opam/config" ]; then
     opam update || UPDATE_FAILED="yes"
     opam upgrade || UPGRADE_FAILED="yes"
-    if [ -n "$UPDATE_FAILED"  -o  -n "$UPGRADE_FAILED" ]; then
+    if [ -n "$UPDATE_FAILED" ] || [ -n "$UPGRADE_FAILED" ]; then
         # Something went wrong, restart from scratch
         rm -rf "$HOME/.opam/"
         opam init
@@ -15,15 +18,15 @@ else
 fi
 
 if [ -n "${OPAM_SWITCH}" ]; then
-    opam switch ${OPAM_SWITCH}
+    opam switch "$OPAM_SWITCH"
 fi
-eval `opam config env`
+eval "$(opam config env)"
 
 make deps
 RC=$?
 if [ $RC -ne 0 ]; then
     # Try again after updating the environment (bin path)
-    eval `opam config env`
+    eval "$(opam config env)"
     make deps
 fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ addons:
     - aspcud
     - nettle-dev   # for ocamlnet
 
-script: bash -ex .travis-ci.sh
+script: ./.travis-ci.sh
 env:
   - OPAM_SWITCH=4.04.2
   - OPAM_SWITCH=4.05.0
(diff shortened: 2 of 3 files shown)


.travis-ci.sh
```diff
@@ -1,11 +1,14 @@
...
+set -ex
 
 export OPAMYES=1
 
 # $HOME/.opam is cached, hence always present.
 if [ -f "$HOME/.opam/config" ]; then
     opam update || UPDATE_FAILED="yes"
     opam upgrade || UPGRADE_FAILED="yes"
-    if [ -n "$UPDATE_FAILED"  -o  -n "$UPGRADE_FAILED" ]; then
+    if [ -n "$UPDATE_FAILED" ] || [ -n "$UPGRADE_FAILED" ]; then
```
Chris00, line 11: Do not see the point of this change.

Chris00 (commented):
I do not see the point of your changes.  Could you explain?

.travis-ci.sh
pmetzger: Neither do I.

redianthus:
I ran ShellCheck and fixed things it spotted.

```sh
find -O3 . -type f -name '*.sh' -exec shellcheck {} \;

In ./.travis-ci.sh line 1:

^-- SC2148: Tips depend on target shell and yours is unknown. Add a shebang.


In ./.travis-ci.sh line 8:
    if [ -n "$UPDATE_FAILED"  -o  -n "$UPGRADE_FAILED" ]; then
                              ^-- SC2166: Prefer [ p ] || [ q ] as [ p -o q ] is not well defined.


In ./.travis-ci.sh line 18:
    opam switch ${OPAM_SWITCH}
                ^-- SC2086: Double quote to prevent globbing and word splitting.


In ./.travis-ci.sh line 20:
eval `opam config env`
     ^-- SC2046: Quote this to prevent word splitting.
     ^-- SC2006: Use $(..) instead of legacy `..`.


In ./.travis-ci.sh line 26:
    eval `opam config env`
         ^-- SC2046: Quote this to prevent word splitting.
         ^-- SC2006: Use $(..) instead of legacy `..`.


In ./opam.ocaml.org/link_blog_doc_ext.sh line 7:
ln -n -s -f $BLOG blog
            ^-- SC2086: Double quote to prevent globbing and word splitting.


In ./opam.ocaml.org/link_blog_doc_ext.sh line 8:
ln -n -s -f $DOC doc
            ^-- SC2086: Double quote to prevent globbing and word splitting.


In ./opam.ocaml.org/link_blog_doc_ext.sh line 9:
ln -n -s -f $EXT ext
            ^-- SC2086: Double quote to prevent globbing and word splitting.
```

For each of them, you can check the corresponding ShellCheck wiki page to get explanations on why this should be changed. (Googling e.g. `SC2086` gives you directly the link)

.travis-ci.sh
redianthus: See [SC2166](https://github.com/koalaman/shellcheck/wiki/SC2166).

pmetzger:
> I ran ShellCheck and fixed things it spotted.

Based on prior experience, ShellCheck seems to complain about things that are fine and propose controversial changes as well.

mshinwell:
This patch is ok, although you should maybe revert the changes for SC2006, since I don't think we have consensus on that.  Although I would reiterate that not fixing SC2006 is in my view a mistake: we should be using the most robust shell language constructs which we can, irrespective as to whether or not uses of the constructs being replaced might be problematic, and irrespective of whether such change breaks old versions of a dead operating system (Solaris in the case of SC2006).

You could also put `set -u` at the top of the scripts.  To a first approximation every shell script should have that.

mshinwell:
(OK assuming that Travis works; there was some strange failure, I have restarted one of the jobs.)

PR #1018, opened 2018-08-23
Add OCaml versions

XVilka:


site/docs/install.md
```diff
@@ -120,16 +128,17 @@ pacman -S ocaml
 Other Arch Linux packages related to OCaml are
 [listed here](https://www.archlinux.org/packages/?q=ocaml).
 
-## macOS
+## macOS
```
pmetzger: Why add a space here? I think that's probably a mistake.

site/docs/install.md
```diff
@@ -138,14 +147,16 @@ brew install opam
 apt install ocaml
 ```
```
pmetzger, line 141: Is Fink still active? Maybe we should remove this mention entirely.

pmetzger (changes requested):
Thanks for the changes! A few comments below...

pmetzger:
Any idea why the CI system failed?

site/docs/install.md
XVilka: [Seems](http://www.finkproject.org/) it is. But is not yet added in repology: https://github.com/repology/repology/issues/243

XVilka:
Fixed the space problem and added NetBSD mention. Also grouped FreeBSD, OpenBSD and NetBSD together. Travis CI error seems unrelated.

Chris00:
Thanks.

PR #1023, opened 2018-09-15
Commits to add the OCaml manual to the ocaml.org web site

pmetzger:
This is a bit of a hack, but I think it's the right hack. I've added the html for the 4.07 manual and an infrastructure for later adding the html for later releases. I have _not_ added build infrastructure for the manual, that would be both complicated and (I strongly suspect) not a great savings in labor.

You may wish to examine the commits individually, as the one to add the html files is rather long and not very informative.


pmetzger:
BTW, arguably, we should use the official manual tarball instead of the version I generated locally as my version is somewhat different (different dates, for example.)

Makefile
```diff
@@ -1,4 +1,5 @@
 HEADACHE = headache -c _headache.conf
+LATEST = 4.07
```
Chris00, line 2: There is `LATEST_OCAML_VERSION_MAIN` in `Makefile.common` for this.

Makefile
```diff
@@ -23,6 +26,13 @@ syncotherfiles:
 	  --exclude '*.cmi' --exclude '*.cmo' --exclude '*.annot' \
 	  -rltHprogv ../ocaml.org-media/* ocaml.org/ ; fi
 
+copymanual:
+	rsync -rltHprogv site/manual/* ocaml.org/manual/
```
Chris00, line 30: I rather think the manual (as big "constant" data) rather belongs to https://github.com/ocaml/ocaml.org-media  On the server, this repo is checked out in `../media/` (from the root of this repo).  If the manual files are not present, a warning may be displayed but the build should not fail.

Chris00: If not too much work, it would be good to at least add the navigation header on all pages of the manual.

Chris00: It is fine for the `.md` pages you added to be part of this repository.

Chris00 (changes requested):

Makefile
pmetzger: That's one obvious fix to make.

Makefile
pmetzger: I can move the thing to the media repo.

Should I be sourcing the official manual for this purpose?

 Should I also be adding the manual sections to releases so that each release will have its manual visible?

Lastly, it would be straightforward, using lambdasoup, to alter the manual pages to add the navigation header, but I believe to do that we would need to get a ruling from @xavierleroy and company — I would like to know their opinion since the manual is not (as such) open source.

pmetzger:
BTW, just to be clear, please don't commit this until we've sorted out exactly how we want to handle it. What I did as a PR here is a strawman. I think it's not _too_ far from what we want but we should hash it out carefully first.

Makefile
Chris00: > Should I be sourcing the official manual for this purpose?

Yes, please.  It is it done through a script, this script can be committed to the media repo.

> Should I also be adding the manual sections to releases so that each release will have its manual visible?

That would be great!

For the authorization to add the banner, I'll let @xavierleroy answer — my purpose is simply that it eases going back to other pages of the site from the manual.

pmetzger: The script would probably be nothing more than wget followed by untar so I'm not sure it is _that_ useful, but if it ends up being more complicated than that, I'll commit such a thing. I'll provide instructions in any case.

pmetzger:
I think perhaps if @xavierleroy doesn't respond I'll do it for now without the navigation bar. It will be easy to add it later if he gives permission.

xavierleroy:
I missed this discussion among many others, sorry about that.

I see no licensing issues with adding navigation links.  (You're not changing the contents.)  Please go ahead.

XVilka:
@pmetzger are there any updates regarding this?

pmetzger:
@XVilka I guess I dropped the ball on it, didn't I. Do you want to try taking it up again? My free time is pretty restricted for the next few weeks at least.

avsm:
Fixed in #1110

PR #1025, opened 2018-09-22
Fix horizontal scrolling on mobile

juvirez:
Here is a screenshot of the problem: http://take.ms/Nhup7

### DIFF
--- a/site/css/ocamlorg.css
+++ b/site/css/ocamlorg.css
@@ -478,8 +478,9 @@ ul.translations:after { content: "]"; }
 }
 @media (max-width: 480px) {
   #footer .column {
-    width: 98%;
+    width: 80%;
     padding-left: 20%;
+    margin-right: 0;
   }
 }
 #footer .entry {


site/css/ocamlorg.css
```diff
@@ -478,8 +478,9 @@ ul.translations:after { content: "]"; }
 }
 @media (max-width: 480px) {
   #footer .column {
-    width: 98%;
+    width: 80%;
     padding-left: 20%;
+    margin-right: 0;
```
pw374: I think that a better fix would be to remove the `padding-left` completely and add `margin-right: 0` without touching `width`.

pw374 (changes requested):
I think that a better fix would be to remove the `padding-left` completely and add `margin-right: 0` without touching `width`. Please let me know what you think.

pw374:
@juvirez

avsm:
@patricoferris @kanishka-work would either of you be able to look into this one?

patricoferris:
Sure, it also falls under some other "responsive" bugs I've found such as this `iframe` :))

video-width

avsm:
Fixed by #1210

PR #1037, opened 2018-11-27
added instructions on how to get ocaml

UnixJunkie:
once you have opam

### DIFF
--- a/site/docs/install.md
+++ b/site/docs/install.md
@@ -29,6 +29,20 @@ the recommended way to install the OCaml compiler and OCaml
 packages. Simply follow the
 [OPAM install instructions](https://opam.ocaml.org/doc/Install.html).
 
+Then, use opam to install an ocaml compiler.
+Here are the commands for a Linux user using the Bash shell and opam-2.0.0:
+```
+# environment setup
+opam init
+eval `opam env`
+# install given version of the compiler
+opam switch create 4.06.1
+eval `opam env`
+# check you got what you want
+which ocaml
+ocaml --version
+```
+
 ## Linux
 
 Most Linux distributions allow OCaml and/or OPAM to be installed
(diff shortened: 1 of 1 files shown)


UnixJunkie:
Recently, people were complaining on caml-list about the quality of the documentation.
Let's hope this makes things clearer.
Feel free to edit to whatever you think would be better.

UnixJunkie:
On purpose, I did not install the latest version of the compiler.
Because, if you install the latest version, sometimes some packages are broken.
So, I feel safer to point users to a version which is not the bleeding edge, so that
there are more chances they can install what they are looking for afterwards.

site/docs/install.md
```diff
@@ -29,6 +29,20 @@ the recommended way to install the OCaml compiler and OCaml
 packages. Simply follow the
 [OPAM install instructions](https://opam.ocaml.org/doc/Install.html).
 
+Then, use opam to install an ocaml compiler.
+Here are the commands for a Linux user using the Bash shell and opam-2.0.0:
```
dbuenzli: I don't think that information is relevant. I'd remove it.

site/docs/install.md
```diff
@@ -29,6 +29,20 @@ the recommended way to install the OCaml compiler and OCaml
...
+```
+# environment setup
+opam init
+eval `opam env`
+# install given version of the compiler
+opam switch create 4.06.1
```
dbuenzli, line 39: Why not use the latest version here ?

dbuenzli: Sorry didn't read your message properly. Makes sense.

site/docs/install.md
UnixJunkie: I've updated it but not removed it.
I think eval is a bash thing.
Again, feel free to edit this PR.

pw374:
Thanks!

Issue #1030, opened 2018-10-31
tutorials: Unit not explained


villesau:
In basics https://github.com/ocaml/ocaml.org/blob/master/site/learn/tutorials/basics.md
reads this:
> The unit type is sort of like void in C, but we'll talk about it more below.

But `unit` is actually not explained below. Also would be very good to open up `Option` a bit!

pmetzger:
If you have some proposed language that would probably be helpful; pull requests are best of all!

villesau:
Unfortunately my understanding of OCaml is currently limited to pretty much on the level tutorials/basics provides so I don't consider my self as proficient enough to explain the concepts or submit a PR. Otherwise I would have already opened one :)

pmetzger:
Unit is pretty simple, it's a type with only one value (also named unit). It's used whenever something doesn't really return a value, thus the comparison to C's voids. Option is a way of packaging up something that might or might not return a value, None for when there's no value and Some for when there is one. Based on that, do you think you could try writing some language anyway? I'm happy to help edit you.

NebuPookins:
Perhaps this issue can be closed?

* The issue mentions the file https://github.com/ocaml/ocaml.org/blob/master/site/learn/tutorials/basics.md but this file no longer exists.
* The issue says the text "The unit type is sort of like void in C, but we'll talk about it more below." is confusing, but this text no longer appears anywhere on the site.
* I'm guessing the replacement for `basics.md` is `a_first_hour_with_ocaml.md`, which does explain (at least a little bit) the concept of unit: "There is exactly one thing of type unit, and it is written (). We use unit to call a function which needs no other argument, and is only used for its imperative side effect."

pmetzger:
If the original file is gone, there is no need for the open issue.

PR #1039, opened 2018-12-22
Added tutorials: Unit not explained #1030

Theomat:
Added some explanations on unit and option types.

Resolving #1030.

Please provide feedback on explanations if you find them to be unclear or confusing.

Thanks.

### DIFF
--- a/site/learn/tutorials/basics.md
+++ b/site/learn/tutorials/basics.md
@@ -503,7 +503,7 @@ type 'a option =
   | None
   | Some of 'a;;
 ```
-So `'a` indicates that is a polymorphic type like a function. So we can use the `option` type with `int`, `float` or whatever but at the same tyme a `int option` is not the same type as a `float option`. Now let's write a `safe_sqrt` that checks that the input parameter is positive, if it is we will return `Some (sqrt x)` otherwise we will return `None`.
+`'a` indicates that it is a polymorphic type. So we can use the `option` type with `int`, `float` or anything else, but at the same time `int option` is not the same type as `float option`. Now let's write a `safe_sqrt` that checks that the input parameter is positive, if it is we will return `Some (sqrt x)` otherwise we will return `None`.
 ```ocaml
 let safe_sqrt x =
   if x > 0. then Some (sqrt x)
--- a/site/learn/tutorials/basics.md
+++ b/site/learn/tutorials/basics.md
@@ -252,7 +252,7 @@ network stack), OCaml provides a `nativeint` type which matches the
 native integer type for your platform.
 
 OCaml doesn't have a basic unsigned integer type, but you can get the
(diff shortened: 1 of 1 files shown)


site/learn/tutorials/basics.md
```diff
@@ -479,5 +479,37 @@ removes a whole class of errors which cause segfaults,
...
+
+Let's say we have the `sqrt` function which given a `float` return it's square root as a `float`. Now the issue is what happens when we call `sqrt` with a negative `float` (for those who don't know the square root is undefined on negative reals) ? Let's suppose when we call the `sqrt` function with a negative number it exits the program. That's bad, we don't want our program to terminate just because of bad input.
+
+ Let's delegate the issue to the called of `sqrt` and use the `option` type :
+```ocaml
+type 'a option =
+  | None
+  | Some of 'a;;
+```
+So `'a` indicates that is a polymorphic type like a function. So we can use the `option` type with `int`, `float` or whatever but at the same tyme a `int option` is not the same type as a `float option`. Now let's write a `safe_sqrt` that checks that the input parameter is positive, if it is we will return `Some (sqrt x)` otherwise we will return `None`.
```
redianthus:
```suggestion
`'a` indicates that it is a polymorphic type. So we can use the `option` type with `int`, `float` or anything else, but at the same time `int option` is not the same type as `float option`. Now let's write a `safe_sqrt` that checks that the input parameter is positive, if it is we will return `Some (sqrt x)` otherwise we will return `None`.
```

redianthus:
I'm not sure that putting the explanation of the option type here is desirable. Variants hasn't been mentioned at this point, and it may be unclear to the reader who doesn't know how to use them.

Theomat:
Adding `option` type was mentioned in #1030.
However I strongly agree with you, the explanation of `option` type isn't really desirable, I found it quite hard to write explanations for an unexperienced reader.

redianthus:
To explain the `option type`, I would rather start by showing how to use a function that returns an `option`, without talking about the definition of the type or the definition of the function.

Starting with something really simple like `int_of_string`, explaining that it will fail in some cases, so we have `int_of_string_opt`. Then, showing how to get the result using pattern-matching, and that if forces you to handle both cases, otherwise the compiler will tell you there's missing cases.

Theomat:
Well, I was trying to avoid  pattern matching, it is true that it is very idiomatic but it is discussed in another [section](http://ocaml.org/learn/tutorials/data_types_and_matching.html).
In that case I think that the `option` type has no place in basics and should be placed elsewhere.

ghost:
So this needs removal of option section and then it can be merged?

avsm:
This now conflicts with changes made to the tutorials elsewhere, so closing for now as there is no activity. @johnwhitington, if you have time when you are going through your tutorial changes to factor in the comments/changes in this PR, that would be appreciated.

PR #1040, opened 2018-12-28
Change the order of the suggested Ubuntu packages.

Zimmi48:
It seems that students looking to install OCaml on Ubuntu will find this page, not know what X11 means and proceed to install using the first suggested command. By swapping the two suggested commands, we ensure that people who don't know what they are doing are getting X11 support and the Graphics package.


pmetzger:
Is there a similar change to make in the English version of the page?

pw374:
thank you

pw374:
@pmetzger yes there is

pmetzger:
@pw374 It should be done, then. :)

PR #1044, opened 2019-01-20
Fix (and add) FreeBSD install instructions

kjakatter:
This changes the use of `pkg_add` to `pkg` since the latter replaced the former years ago.  No currently supported release of FreeBSD even has `pkg_add` anymore.

This also adds install instructions for getting opam using `pkg`.


avsm (approved):
Thanks very much! These changes look correct to me.

Chris00:
Thanks.

PR #1050, opened 2019-03-23
Modules tutorial: Fix "Displaying the interface of a module"

smindinvern:
The described method no longer works.  Update the tutorial to use the `#show` directive instead[1].

[1] https://stackoverflow.com/a/43292437


Chris00:
Thanks.

PR #1051, opened 2019-03-31
update opam metadata to 2.0 format

avsm:


ocaml-org.opam
```diff
@@ -25,10 +21,11 @@ depends: [
   "cow" {>= "2.2.0"}
   "camlp4"
 ]
-depopts: [
-# for the Gtk tutorial:
```
Chris00: I think this comment should stay otherwise it is hard to understand why `lablgtk` (the `2` should be removed, it is a typo) is needed.

Chris00 (changes requested):
Thanks for submitting this PR!

ocaml-org.opam
avsm: good point! done in 58bf577eb967e7f4494aa6589ac30c295db73294

avsm:
I just noticed that the dockercloud CI is still hooked in; i've disabled it now to stop the X marks on every PR

PR #1060, opened 2019-04-18
update Latest version of OCaml on front page

pw374:
I don't know if that would actually work...


pw374:
apparently it won't be enough

Chris00:
Thanks but it will not be enough to update the version.

Issue #1027, opened 2018-10-07
Put 4.07.1 announcement on the front page?


pmetzger:
Since 4.07.1 is out, perhaps it should be announced on the front page?

Chris00:
Except for the date, it should happen automatically.  I'll check it out — maybe a consequence of the OPAM 2 migration.

pw374:
Something's broken, since it hasn't happened automatically.

That being said, it seems the date of the latest release will still have to be manually edited.

Chris00:
@avsm If there was a convention to determine the release date of OCaml compilers (say to add “Released on ...” to its description), that would be really great.

pw374:
Ummm...  now it shows 4.06.1 ! It’s even worse than yesterday

avsm:
@Chris00 perhaps the best way to do this is to add a `x-released-on` header to the `opam` file for the compilers, and query that from the website build for the descriptive information

Chris00:
@avsm Fine with me!

PR #1061, opened 2019-04-18
determine OCaml's latest version

pw374:
this might be better, I'm not sure


Drup:
Why not just `opam show ocaml-base-compiler --field version` ?

Drup:
Actually, that might not work if you already have an installed ocaml package.

`opam list -V ocaml-base-compiler --no-switch -s --columns version` avoids the use of grep and sed.

Chris00:
@Drup I have merged you change.  (Nothing to do with the [current OCaml version not being detected](https://github.com/ocaml/ocaml.org/issues/1027), this is due to an automatic opam upgrade failing.)

PR #1066, opened 2019-05-10
Update latest distribution versions in INSTALL.md

XVilka:
Because of the recent releases of OCaml itself, Fedora, and Ubuntu.


Chris00:
Thanks.

PR #1069, opened 2019-05-18
Cull dead company entries

yminsky:
Cull some dead entries.  Only culled entries where the links were dead or pointing to the wrong place. Some of these are companies that went out of business, some (like monoidics) were acquired. But it seems sensible to keep this list as current entities that use OCaml.


Chris00:
Thanks.  (I removed the logos too.)

PR #1072, opened 2019-05-30
-debug is not a valid option for ocamlbuild

co-dan:
The way to enable debugging information is to set the 'debug' tag, as described in the ocamlbuild manual.


Chris00:
Thanks.

PR #1090, opened 2019-10-21
Add Ahrefs's tech blog to planet

Khady:
name for the feed: Ahrefs
url: https://tech.ahrefs.com/feed

The tech blog is focused on articles related to technologies we are using: ocaml, reasonml, bucklescript.


pw374:
Thanks

PR #1098, opened 2019-12-17
Update URL for js_of_ocaml toplevel

citizen428:
The previous URL results in a 404, since it's missing /manual in the path.


avsm:
Thank you for the fix!

PR #1107, opened 2020-02-22
Update install.md

tk-425:
Instead of 
`# check you got what you want
which ocaml`,
changed to 
`#check your ocaml version`
because previous comment was confusing.


avsm:
The `which ocaml` is important as it helps the user confirm that their PATH is set correctly.

avsm:
No response on this PR, so I'll go ahead and close it. Please re-open if necessary.

PR #1110, opened 2020-03-17
Host reference manual on ocaml.org instead of caml.inria.fr

xavierleroy:
Following COVID-19 containment measures, Inria Paris is closed, and there is a possibility that the facility will be shut down completely in the near future, including the server room that contains the caml.inria.fr Web server.

It is therefore high time to stop hosting crucial resources for the OCaml community on caml.inria.fr, as I have repeatedly argued in the past.  

This PR adds a copy of the reference manual (various formats, various versions) to ocaml.org and rewrites URLs all over so that the reference manual is accessed from ocaml.org and the source tarballs from github.com.  

The PR is best reviewed commit by commit.  The first commit is huge (it's the import of all the files for the reference manual) but the other commits have reasonable size.

I do not know how to rebuild and test the Web site locally, so all this is untested.  Please tell me how to test.

Special mention: @Octachron @pmetzger

PR #1111, opened 2020-03-18
OCaml 4.09.1 release

Octachron:
This PR adds a page for the release of OCaml 4.09.1 (a bugfix release). I reused the same short format as the 4.08.1 version. Since there is no link to binaries file, this PR commutes with the great migration of #1110 .


avsm:
Thanks!

PR #1114, opened 2020-03-25
Fix documentation about stack length complexity

giltho:
Hi !

See https://github.com/ocaml/ocaml/blob/trunk/stdlib/stack.ml#L50


avsm:
Thanks!

Issue #1105, opened 2020-02-07
Makefile.common is executed too many times


sanette:
Because of the recursive structure of the Makefile (which calls make inside the Makefile),
it turns out that Makefile.common is executed for each processed file.
In this way, if you have 100 files to process, the command

`LATEST_OCAML_VERSION=$(shell opam list -V ocaml-base-compiler --no-switch -s --columns version | tail -n 1)`

is executed 100 times, which slows things down quite a lot.

pmetzger:
Good catch.

PR #1115, opened 2020-03-26
include all OCaml manuals 4.00 - 4.10

sanette:
This PR is the result of a lot of work, so obviously it's not going to be perfect. However, as a try (or experimentation) I believe it is ready to be online. 
* It contains all html files of the html-bundled OCaml manuals from version 4.00 to 4.10 available on the INRIA web site at this moment, slightly post-processed.
* the post-processing is mainly done for improving design but also for providing "search as you type" capabilities for the API parts of the manual.
* the other formats (PDF, info and txt) are also archived for downloading.

This PR dos *not* contain the scripts used for post-processing. My idea is to do one thing at a time. Once we get more feedback on the website, we can work to make the scripts more robust and stabilize the process. We can also trash everything and revert to the previous documentation.

Following the advice of some people, the structure of the Manual is *not* modified. The numbering of chapters is the same as usual. To my opinion, navigating these chapters is a bit confusing, but again, one thing at a time. As a step toward a re-organization, I tried to highlight the important parts of the manual in the docs/index page.

There are some broken links the the `compilerlibref` part, but they were already present in the INRIA version; I believe that this part is work in progress, so I have not attempted to correct it.

The main goal for me was to make OCaml doc easier to reach for beginners and nicer to read, and to try to convey a modern image of the language. For others, the goal was simply to archive the manual in a safe place.


Octachron:
You probably want to update the existing manual files in `releases/` rather than add a second copy of all manuals.

sanette:
> You probably want to update the existing manual files in `releases/` rather than add a second copy of all manuals.

At this point I wanted to interfere the least possible with the existing site, but if you prefer, I can do this, yes. (note also that I use `.md` files instead of `.html` files in order to be better integrated into the site design.)

sanette:
The build failed because it took too much time. I had noticed earlier two possible reasons:
https://github.com/ocaml/omd/issues/199#event-3126092853
https://github.com/ocaml/ocaml.org/issues/1105#issuecomment-584295101

sanette:
> You probably want to update the existing manual files in `releases/` rather than add a second copy of all manuals.

in fact, after thinking, if you allow I would rather not use the `releases` dir. I find it better to separate the `manual` part and the `api` part. (and hence have `ocaml.org/docs` and `ocaml.org/api` URLs).

Concerning the files that have been uploaded very recently in `releases/X.XX/htmlman` we could either remove them, or, in fact it would be better to keep them, so that the post-process script could be run on them to generate the `.md` files.

sanette:
?

sanette:
(sorry, wrong manipulation.)
Did anyone have the chance to test this PR?

Octachron:
One thing is clear, we are not removing the source of the manual in `releases`. They were moved because the `caml.inria.fr` site cannot be relied upon anymore. I am not sure if this means that we want to have temporarily two copies of the files, or if this means that we should start by integrating the preprocessing scripts right away.

sanette:
If we are ok with hosting 2 copies (I know this is not environmental friendly, but after all the whole manual is no so huge) I would suggest keeping the "original" html in `release` and run the processing script on them to populate `docs` and`api`.

bikallem:
I have been using this at https://sanette.github.io/ocaml-api/4.10/index.html for a while now - possibly for about a month now. The layout and general aesthetics is much better than before but the feature that I found most useful is the API search functionality. Previously I would do CTRL+F on a page by page basis and needless to say it was a tiresome exercise. Are there plans to publish other package API in ocaml.org, for e.g. lwt, js_of_ocaml, mirage, etc?

sanette:
not yet. The first goal is still to work on the manual. But the good news is that this includes making the search API even more robust and portable. So, yes, then it should be possible to use it for other libraries. I don't think it's difficult.

sanette:
In order to make this potential big change easier to control, digest, and modify, the decision was to split it up: one (easier) part concerning only the presentation of the docs/ page, but not modifying the manual itself. The more challenging part (the manual itself) will now occur upstream in the manual section of ocaml.
A first proposition for part 1 is now ready ==> https://github.com/ocaml/ocaml.org/pull/1124

sanette:
I'm closing this because it will be much easier to discuss https://github.com/ocaml/ocaml.org/pull/1124 first.

PR #1116, opened 2020-03-30
Use #require instead of #load to load graphics module in tutorial

RobbieClarken:
I am learning ocaml by following the ocaml.org tutorial. I have found it excellent so far but got very stuck at the point:

> To use Graphics in the interactive toplevel, you must first load the library with
> ```
> #load "graphics.cma";;
> ```

It took me some time to discover I needed to run `#require "graphics";;` before `#load` so I think it is worth documenting this. Should we also mention running `#use "topfind";;` first? In my case I didn't need to do this but I'm not sure if that will be the same for everyone.

### DIFF
--- a/site/learn/tutorials/structure_of_ocaml_programs.md
+++ b/site/learn/tutorials/structure_of_ocaml_programs.md
@@ -262,7 +262,7 @@ first load the
 library with
 
 ```ocaml
-#load "graphics.cma";;
+#require "graphics";;
 ```
 Windows users: For this example to work interactively on Windows, you
 will need to create a custom toplevel. Issue the command `ocamlmktop


Octachron:
It is not necessary to use `#load` after `#require "graphics"`. I agree that it is probably better to mention `#require` since it works across more OCaml versions. However, setting up topfind requires more than the basic compiler bundle. One solution might be to add a link to a tutorial for setting up the REPL (if it exists?).

RobbieClarken:
Thanks @Octachron - I've update the PR to replace `#load` with `#require`. I see that `#require` comes from the `findlib` library. Is it safe to assume that this library will be installed with most OCaml installations?

In my case, I installed OCaml on macOS by first installing opam using homebrew and then using opam to install ocaml 4.10.0 as described on the [`ocaml.org Installation page`](https://ocaml.org/docs/install.html#OPAM).

pw374:
Personally I'm used to the old  fashion way, which is to invoke the REPL like this
```
ocaml graphics.cma
```

But I guess I'm kinda old school now, and my ways are deprecated 😆

Octachron:
The main issue is that `graphics` is nowadays a library independent of the compiler distribution. Consequently, for recent version of OCaml, graphics.cma is not anymore in the default path and `ocaml graphics.cma` doesn't work.

Findlib is the traditional way to find and use OCaml library and it should be installed on any OCaml installation that uses at least one library.

pw374:
@Octachron thank you for confirming that 

> my ways are deprecated

😆

PR #1117, opened 2020-03-30
adds the OCaml 2020 page

ivg:
Also updates the indices, both English and French, to point to the
new conference.


avsm:
Thanks @ivg!

PR #1119, opened 2020-05-05
Add reasonml.org blog feed

ryyppy:
Name: Reason Documentation Blog
Url: https://reasonml.org/blog/feed.xml
Kind: Personal

Even though the content is not solely OCaml related, it might still be interesting to add the blog RSS feed for the whole Reason platform.


agarwal:
I'm wondering about the entry as "Reason _Documentation_ Blog". Is the blog about documentation only? Also, the feed declares the title as "ReasonML Blog". Wondering if you want to list it with the same title.

PR #1121, opened 2020-05-20
Update taste.md

nicholasjpaterno:
This seems out of order.

It would be useful for the higher-order section to go before the polymorphism section.


agarwal:
Ok, I merged this.

PR #1123, opened 2020-05-27
Update installation instruction on Windows

### LINKED ISSUE #1097: Bash on Windows installation tutorial link is broken
On the [installation page](https://fdopen.github.io/opam-repository-mingw/installation/) there's a part that says:

> Use « Bash on Windows » and install OCaml as on Ubuntu. Main OCaml versions and OPAM work. The disadvantage of this method is that you don't build native Windows binaries but Linux binaries. See tutorial.

However the "see tutorial" link is broken, leads to some random spam.

mdesharnais:
Delete reference to unmaintained OCPWin.

Rename and update reference from "Bash on Windows" to "Windows Subsystem for
Linux". Move up in the list of alternatives as it is a much simpler solution
than compiling one's own version of OCaml.

Fix link to compiling from source.

close #1097

### DIFF
--- a/site/docs/install.md
+++ b/site/docs/install.md
@@ -191,22 +191,19 @@ Under Windows, the following direct solutions are available to use OCaml:
 * [OCaml for Windows](https://fdopen.github.io/opam-repository-mingw/)
   provides an experimental OPAM repository and opam build for Windows.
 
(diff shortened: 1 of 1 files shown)


avsm (approved):
Thanks for this contribution!

PR #1124, opened 2020-06-06
update Docs/index pages

sanette:
Modification to docs/index pages to direct the newcomer to the most relevant parts of the Manual
(which means the usual manual recently uploaded in /releases).


mseri:
This looks good, but there seems to be a lot of replication between the two files. Would moving the javascript snippet to a separate file help its maintenance? I know it’s few lines, but it seems easy to forget to fix both files in case of further changes by somebody else.

In any case, I think this would be very useful to have

sanette:
good idea, I'll do this

sanette:
on the other hand, do we really want to keep a French version?
If so, it would probably be better to have some kind of translation mechanism

sanette:
notice for instance that currently there are 3 pictures of books in the English version and only 2 in the French version
(this was corrected in the PR)

sanette:
As suggested, I have put the javascript in a separate file (I think it makes sense to leave it in the same directory, since it's only useful for that particular page).
I also corrected the link to "tools" for older versions, and made the selected version more permanent.

sanette:
I have added 4.11
Here is a screenshot.

Screenshot_20200831_093310

sanette:
the merge conflicts concern recent modifications made by this commit https://github.com/ocaml/ocaml.org/commit/635ce2f019cbbc6ecfa7c557dc66261176a591ae
the point they address (dates too old) was also corrected in this PR, but in a slightly different manner: removing the incriminated sentences. I'm OK with both ways.

avsm:
@sanette if you could merge this against the `master` branch of ocaml.org I'd be glad to review/merge it.

sanette:
sure I'll do this when I have some time. Maybe this week-end

sanette:
@avsm , the merge seems to be ok

avsm:
Thanks, pushed to https://staging.ocaml.org to preview.  We'll need an issue to remember to update the OCaml manual version list and cross-reference it to https://github.com/ocaml/opam-repository/issues/17530 for the OCaml 4.12 release readiness tracking.

sanette:
sorry, I had left some spelling typos, that I have just corrected

sanette:
if https://github.com/ocaml/ocaml/pull/9755 gets accepted, we will also need to update the links to the manual (or maybe not, depending where the new `webman` is located)

avsm:
This is really nice work, thank you for all the effort @sanette! I've edited  ocaml/opam-repository#17530 to remind us to update the site for 4.12. Given that @Octachron indicates that he would like your manual changes included in 4.12.0, there's no point automating the JavaScript in this page until that settles down.  So this PR is just fine by me. Pushing it live!

ghost:
@sanette I saw one site use weblate for translation

PR #1125, opened 2020-06-11
Update structure_of_ocaml_programs.md

bowsersenior:
Make docs for 3rd snippet consistent with code

### DIFF
--- a/site/learn/tutorials/structure_of_ocaml_programs.md
+++ b/site/learn/tutorials/structure_of_ocaml_programs.md
@@ -524,7 +524,7 @@ let scrollbar = GRange.scrollbar `VERTICAL ~packing:hbox#pack ()
 ```
 
 Third snippet: The author imports all the symbols from the `GdkKeysyms`
-module. Now we have an unusual let-binding. `let _ = expression` means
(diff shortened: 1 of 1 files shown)


samoht:
Thanks!

sanette:
on the other hand the text is not completely exact. In fact the `let ()=` does check that the result is indeed `()`.

samoht:
hum indeed, maybe I've merged that one too quickly. @sanette do you think I should revert this change? (`let () =` is used as-is just after, so I think that's ok)

pw374:
I think this should be reverted indeed.

because "_but throw away the result_"  makes sense only with `let _ =` (it doesn't with `let () =`)

pw374:
I reverted that merge :)

pw374:
https://github.com/ocaml/ocaml.org/pull/1125/files#diff-c7075d3266aa322ac73a724bb06e6924L543

To achieve consistency, it's the `let () =` from the "3rd code snippet" that should've been changed to `let _ =`

bowsersenior:
Thanks for the comments. I opened a [new PR](https://github.com/ocaml/ocaml.org/pull/1157) that updates the 3rd snippet instead.

PR #1129, opened 2020-06-24
Correct OCaml 4.10.0 release date in index.md

lingocoder:
According to [_its individual release page_](https://ocaml.org/releases/4.10.0.html) and [_the main release page_](https://ocaml.org/releases/), OCaml 4.10.0 was released on Feb 21, 2020; not Flag Day, 2019.


avsm:
Thanks!

Issue #1128, opened 2020-06-24
Simple alias for the latest manual


dbuenzli:
Since @xavierleroy's #1110 the manuals are now hosted on this website. 

However the link to the latest version of the manual remains as hard to track as before and somehow https://ocaml.org/releases/latest/manual.html is a bit unconvenient to type. 

Could we please have the obvious URI https://ocaml.org/manual redirect to the latest version of the manual ? 

Also the link to the manual in the footer still redirects on caml.inria.fr

xavierleroy:
I fully support this request, especially if it supports "deep linking", e.g. https://ocaml.org/manual/native.html -> https://ocaml.org/releases/4.10/htmlman/native.html .  A symbolic link  on the server side would do fine.  FWIW, I couldn't figure how to do it as part of #1106

pw374:
Let's see if https://github.com/ocaml/ocaml.org/commit/9b8245a415ff864dc61b5538536ff571450bb5ee works... If it doesn't, I'll revert the commit...

dbuenzli:
Works ! Thanks. 

That being said I'm not sure the "deep linking" idea is such a good thing with respect to link rot. The problem is that users will effectively browse the manual from `http://ocaml.org/manual/`. So they will end-up cut and pasting links which could no longer work in the future because of structural changes to the manual.

Personally I think a simple `303` to ` /releases/latest/manual.html` which itself redirects to the latest version is better idea -- it also means the `/manual` link won't need an update on each new release provided  `/release/latest` is updated.

pw374:
I'm almost surprised the symbolic link works (it was like a cheap 50-50 bet, it just depends on how the server is configured, and I haven't had access to the server for quite a while).

Now, that symbolic link doesn't actually provide a satisfactory solution.

As @xavierleroy pointed out (on the commit itself), it'll be a pain to update that manually for each release. It doesn't seem difficult to update it automatically though. However, as @dbuenzli points out, that will generate dead links when the structure of the manual changes.

So, a `303`? Yes, I think that would be the best solution here (not ideal because it won't show up in the git repo but that's a tiny detail). It seems that ocaml.org is served by a lighttpd server, therefore gotta access lighttpd configuration file (can't just edit some `.htacess` file, afaik).

Intermediate solution, which doesn't require direct access to server, would be to drop deep linking and just create a file `site/manual/index.html` that redirects (on the client side) to `release/latest/manual.html`.

What are your thoughts?

[...]

PR #1134, opened 2020-07-06
use /manual/ as alias for latest manual via symlink

pw374:
related to https://github.com/ocaml/ocaml.org/issues/1128

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,11 +8,16 @@ local: script/relative_urls
 production: pre-build
 	$(MAKE) gen_md gen_html
 	$(MAKE) syncotherfiles
+	$(MAKE) link-to-latest-manual
 	$(RM) ocaml.org/robots.txt
 
 staging: pre-build
 	$(MAKE) gen_md gen_html SET_STAGING='-set staging'
 	$(MAKE) syncotherfiles
+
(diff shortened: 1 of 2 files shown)


avsm:
I resolved conflicts and pushed this to staging.ocaml.org to make sure it works.

Makefile
```diff
@@ -8,12 +8,17 @@ local: script/relative_urls
...
+	$(MAKE) link-to-latest-manual
 	$(RM) ocaml.org/robots.txt
 
 staging: pre-build
 	$(MAKE) gen_md gen_html SET_STAGING='-set staging'
 	$(MAKE) syncotherfiles
 
+link-to-latest-manual:
+	$(RM) site/manual
+	cd site && ln -s "$(shell cd site && ls -d releases/[0-9].[0-9][0-9]/htmlman |sort|tail -n 1)" manual
```
ghost, line 20: Not sure if this will work for version 10.0, but worry about it when we get there

avsm:
This isn't quite the right fix; the use of symlinks isn't working right. I think we need a redirect elsewhere for this (e.g. in the Caddy config), so I'll close this one out for now.

PR #1140, opened 2020-08-11
Also compile the site with OCaml 4.09.1 and 4.10.0

komainu8:
Hello,

I added ocaml compiler for using on TravisCI.

Thanks,


avsm (approved):
Thanks!

PR #1142, opened 2020-08-19
OCaml 4.11.0 release

Octachron:
This PR adds the release page and manual for OCaml 4.11.0 .
It should also update the various index with the new release date.


Octachron:
I  have added the release info for 4.10.1 too.

pw374:
Thanks!

PR #1143, opened 2020-08-20
basics ja: Modify to more suitable link

komainu8:
Currently, this link moves us from the page of Japanese to the page of English.

I think suitable that the language of the page is the same source of link and destination.


pw374:
thanks

PR #1144, opened 2020-08-20
Add dmbaturin's blog

dmbaturin:
I've made a feed with only posts tagged "ocaml" to avoid syndicating unrelated stuff.

### DIFF
--- a/planet_feeds.txt
+++ b/planet_feeds.txt
@@ -76,3 +76,4 @@ Gabriel Radanne|http://drup.github.io/feed-ocaml.xml
 The BAP Blog|https://binaryanalysisplatform.github.io/feed.xml
 Tarides|https://tarides.com/feed.xml
 Ahrefs|https://medium.com/feed/ahrefs/tagged/ocaml
+Daniil Baturin|https://baturin.org/blog/atom-ocaml.xml


dmbaturin:
@pw374 Thanks for merging! 

However, for some reason, it's marked as broken with "Nettls_gnutls.TLS.Error(GNUTLS_E_PUSH_ERROR)" error message.

The W3C validator can fetch my feed and considers it valid: https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fbaturin.org%2Fblog%2Fatom-ocaml.xml

Do you have any ideas what the issue may be?

pw374:
It seems your feed isn't the only one affected by that issue :(

It probably doesn't have anything to do with the validity of the contents, but (according to the error message) more with the communication between your server and the process that fetches the data. Perhaps an issue with the TLS implementation or handshake? @Chris00 any ideas?

(I don't have time to dig into this issue at this time. I _might_ later...)

dmbaturin:
Some feeds are genuinely dead, too. I'm going to make a PR to remove (or update, when possible) the obviously dead ones.

Chris00:
It's a problem with [OCamlnet](http://projects.camlcity.org/projects/ocamlnet.html) which has been reported (site down ATM) but not yet fixed.  It affects several of my projects.  I'll move to [cohttp](https://github.com/mirage/ocaml-cohttp) when I'll find some time.

PR #1157, opened 2020-09-10
Fix code sample to match docs

bowsersenior:
Following up on [previous PR[(https://github.com/ocaml/ocaml.org/pull/1125) based on [this comment](https://github.com/ocaml/ocaml.org/pull/1125#issuecomment-689889651)


pw374:
thanks!

PR #1163, opened 2020-09-14
basics.ja: add missing translate to Japanese

komainu8:
I added a translation.
Because the following sentences had not been translated into Japanese yet.

```
 Using let (without rec) allows you to re-define a value in terms of the previous definition. For example:

# let positive_sum a b = 
    let a = max a 0
    and b = max b 0 in
    a + b;;

This redefinition hides the previous "bindings" of a and b from the function definition. In some situations coders prefer this pattern to using a new variable name (let a_pos = max a 0) as it makes the old binding inaccessible, so that only the latest values of a and b are accessible.
```


avsm:
Many thanks for the contribution!

PR #1166, opened 2020-09-19
add logo images

sanette:
This is the image used in https://discuss.ocaml.org/
I'm not sure where to find them.
 Having them here could be useful for instance when creating a local version of the manual, see https://github.com/ocaml/ocaml/pull/9755#discussion_r484764944


gasche:
I just looked at this. The OCaml logos are available at https://github.com/ocaml/ocaml-logo . Could we use the logos there directly directly (for example, downloading logos from that repository in build scripts, if you need to), instead of hosting them on the ocaml.org repository as well?

sanette:
oh thanks! then yes, of course, this is even better

PR #1180, opened 2020-10-02
Change "industrial strength" to "industrial-strength"

eli-b:
I think this is more correct grammatically, and it's in line with the other md files.


pw374:
thanks

PR #1194, opened 2020-10-24
Get rid of more forge.ocamlcore.org

gildor478:
This is an update to the new repositories on github (when possible).


Chris00:
Thanks.  I think there are various links on the meeting pages that should be updated.

PR #1197, opened 2020-10-28
Update Windows section in french "install" page

emillon:
This is a translation of 5815fa6 and 49b144e so that both languages are consistent.


samoht:
Merci! ;-)

Issue #1199, opened 2020-11-02
Improve contrast ratio


patricoferris:
## Problem

Modern web accessibility tooling flags many of OCaml.org's pages as failing in some easily fixable areas. Most notably (and the topic of this issue) in contrast ratio between two elements, mainly the orange links and the white background. This is thanks to @ILeandersson's work in this area. 

To reproduce the web accessibility reports one can either: 
 - Visit this link: https://wave.webaim.org/report#/ocaml.org 
 - Or as a second option a [Google Lighthouse](https://developers.google.com/web/tools/lighthouse) audit can be run from Chrome's Dev tools. 

As it currently stands the links are `#c45641` with the background being plain white, this has a contrast ratio of `4.42`. The WCAG standards require (at level AA) [4.5](https://www.w3.org/TR/WCAG21/#contrast-minimum) or (at level AAA) [7.0](https://www.w3.org/TR/WCAG21/#contrast-enhanced). 

## Solution 

For the most part, these links are found on a white background, so [changing the colour](https://github.com/ocaml/ocaml.org/blob/master/site/css/bootstrap.css#L223) will fix the problem. Below I have listed three colours, the original, a level AA compliant colour and a level AAA compliant colour. 

 - Current: `#c45641`
 - Level AA: `#c15540`
 - Level AAA: `#914030`

Other areas flagged as low contrast (again with AA and AAA solutions) are: 

- White text and green background of buttons (ratio `2.53`)
    + Current background colour: `#8eaf20`
    + Level AA: `#008000`
    + Level AAA: `#015f01`
- Focused Navbar item (text colour is white and ratio `3.35`): 
    + Current background colour: `#c87a27`
    + Level AA: `#b15b01`
    + Level AAA: `#844400`

There are quite a few within the code blocks for different identifiers: 
 - Modules (background colour `#f5f5f5` and ratio `3.03`) :
    + Current text colour: `#829f1c`
    + Level AA: `#008000`
    + Level AAA: `#006700`
- Variables (same background colour and ratio `3.25`): 
  + Current text colour: `#b8860b` (this is particularly hard to achieve a good ratio)
  + Level AA: `#ce4813`
  + Level AAA: `#98340c`
- Strings (same background colour and ratio `4.27`): 
  + Current text colour: `#bb255d` 
  + Level AA: `#a95472`
  + Level AAA: `#941a45`
- Types (same background colour and ratio `3.72`): 
  + Current text colour: `#c746cc` 
[...]

PR #1201, opened 2020-11-03
improve contrast ratios

patricoferris:
This PR fixes quite a few of the contrast ratio problems outlined in #1199 to the WCAG AA level.


avsm:
image

Pushed to staging.ocaml.org and the above screenshot is the differences in my browser. Looks good to merge, thanks!

PR #1206, opened 2020-11-07
POY repository was moved

veeara282:
AMNH has moved the POY repository [here](https://github.com/amnh/poy5), but hasn't updated it on [their own web page for it](https://www.amnh.org/research/computational-sciences/poy). I changed the link to the new GitHub URL.


avsm:
Thanks!

PR #1210, opened 2020-11-09
fix responsive layout and add iframe titles

patricoferris:
This PR: 
 - Hopefully finishes the work started at #1025 using @pw374's approach which seems to work well (be good to test on a device once this is in staging, but laptop testing suggests it works)
 - Fixes the `<iframe>` bug also brought up in that PR and whilst I was there I added titles to the `<iframe>` elements as WCAG standards recommend https://www.w3.org/TR/WCAG20-TECHS/H64.html (I double-checked the French version with people who can speak it :) )

### DIFF
--- a/site/css/ocamlorg.css
+++ b/site/css/ocamlorg.css
@@ -494,7 +494,7 @@ ul.translations:after { content: "]"; }
 @media (max-width: 480px) {
   #footer .column {
     width: 98%;
-    padding-left: 20%;
+    margin-right: 0px;
   }
 }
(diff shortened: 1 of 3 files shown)


avsm:
Pushed to staging.ocaml.org to preview.

patricoferris:
Thanks @avsm, all looks good on my end: 

## iPhone Width Problem



## Tablet Video Overlap Problem
tablet-video

avsm:
Thanks!

PR #1221, opened 2020-11-23
Add Odoc to tools and fix malformed html

patricoferris:
This PR: 

 - Adds Odoc to the *Incubate* platform tools! 
 - Fixes some malformed HTML that was in the page ...


patricoferris:
Thanks to @talex5 for pointing out the incorrect href links for some of the Platform tools which are also fixed in this PR.

avsm:
Thanks!

PR #1217, opened 2020-11-15
trim redundant or non essential links from footer

ghost:
#1216 

I don't expect most visitors to need to visit logos page.

The credits link is redundant (and git log credit generation is not working, btw) with about the site. Find us on GitHub is redundant with website issues and edit this page links.

PR #1224, opened 2020-12-07
Accessibility improvements

patricoferris:
This PR continues with the improvements started in #1199 in particular: 

 - It darkens comments in code sections as the contrast ratio isn't WCAG compliant -- I believe having them in italics is sufficient to distinguish them as a comments rather than code.
 - Quite a few links have no "name" (alt text, title attribute etc.) which [is not compliant with the same set of standards](https://web.dev/link-name/). To fix this I've added title attributes to the image icon for news feed items (and similar). 
 - Alt tags for the Jane Street and Facebook images have also been added (in French too). 
 - The dangling link referenced in https://github.com/ocaml/ocaml.org/pull/1217#issuecomment-730354544 has been removed, I didn't see any changes to layout but a trip to staging would be useful for this PR to double-check :) 

A benefit of this PR is that the homepage and learn page should now score 100 on [Google Lighthouse's](https://developers.google.com/web/tools/lighthouse) accessibility metric 🎉  There is still a lot of improvements in terms of accessibility to be made however.

### DIFF
--- a/script/rss2html.ml
+++ b/script/rss2html.ml
@@ -472,8 +472,9 @@ let headline_of_post ?(planet=false) ?(img_alt="") ~l9n ~img e =
     else match get_alternate_link e with
          | Some l -> Uri.to_string l
          | None -> "" in
+  let title = string_of_text_construct e.Atom.title in
   let html_icon =
-    [Element("a", ["href", link],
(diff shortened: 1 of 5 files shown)


avsm:
Really nice changes, thank you!

PR #1227, opened 2020-12-11
Add a section on `drom` in the up-and-running tutorial

lefessan:
`drom` may be useful for newbies to create their first OCaml project with many tools already configured. I added a section in the up-and-running tutorial to explain how to install it and how to create a simple project.

### DIFF
--- a/site/learn/tutorials/up_and_running.md
+++ b/site/learn/tutorials/up_and_running.md
@@ -129,7 +129,7 @@ downloaded
 Done.
 ```
 
-## A first project
+### A first project with Dune
 
 Let's begin the simplest project with Dune and OCaml. We create a new directory
 and ask `dune` to initialise a new project:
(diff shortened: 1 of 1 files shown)


ghost:
Maybe add it to incubate on this page: https://ocaml.org/platform/ ?

lefessan:
@kanishka-work you mean "add it also" or "add it instead" ? Of course, it can be added on this page too, but I think it's already fairly stable on the scaffolding part (the one put forward in this description), while it is indeed a bit less stable for day to day usage.

ghost:
I was thinking of "add it also". To be clear about my role, I am not a reviewer / maintainer, I just came across this PR and noticed this.

avsm:
Thanks @lefessan, this is very good and useful. However, I do think we need to get a view on the other tool (`spin`) before making a recommendation on the default page here at ocaml.org, especially since they are all relatively new.

I don't want to delay this information going up. however To progress this PR, I'd suggest:
- put the information on this PR in the `drom` homepage.
- I can modify this PR to link to "There are several project generators available to construct the metadata files: - drom - spin"
- Users can click through and get the latest information from them there, and decide which one to use.
- When there is consensus on one tool to recommend, the information can go directly on ocaml.org in a future iteration.

avsm:
cc @tmattio regarding Spin going up here as well.

avsm:
Given that there are multiple "scaffolding generator" projects, @tmattio has kindly volunteered to find consensus with @lefessan @craigfe and the others to see if we can make a common recommendation from ocaml.org. Until then, closing this particular change as it's not clear which one to recommend to new users just yet.

PR #1232, opened 2021-01-22
Fix hosting providers

patricoferris:
This PR: 
 - Updates the hosting providers to the current ones (and includes a note on Rackspace's support from 2012-2020)
 - Fixes a very minor accessibility bug for mobile layouts

One question - the hosting information has never been on the `fr` page. Do we want to add it there?


avsm:
thanks!

PR #1234, opened 2021-01-27
OCaml 4.12.0 release and manual

Octachron:
This draft PR is mostly here to check the release process for the manual and start writing down the highlights for OCaml 4.12.0 in preparation to the upcoming 4.12.0 release.

I have also tried to simplify the release page that accumulated quite lot of clutters: Do we really need to mention all opam switch variants? The question is even more accute now that 4.12.0 has switched to using `ocaml-option-*` as configuration packages.
Similarly, I have dropped the links to ocaml-java and ReScript, since both of them are stuck to old version of OCaml.

Finally, the changes to the indexes in `docs/` are here to fix a SEO issue: search engines tend to rank random versions of the manual higher than the latest version. One reason seems to be that this latest version of the manual is not linked directly from the site. In order to fix this, I have updated the html structure to link by default to the latest version and added a "latest" version in the version selector.  I have also fixed the javascript selector to work on the local version of the site, to make testing easier.


avsm:
You can push this to the `staging` branch if you'd like to see it live on staging.ocaml.org.

Octachron:
Good idea, thanks for the suggestion.

avsm:
>  Do we really need to mention all opam switch variants? The question is even more accute now that 4.12.0 has switched to using ocaml-option-* as configuration packages.

I don't think we need to mention all the switch variants, but listing the valid ocaml-option packages would still be useful. We try to track them systematically here for CI: https://github.com/ocurrent/ocaml-version/blob/master/ocaml_version.ml#L395 so could autogenerate some of this page from the ocaml-version library for 4.13 to reduce the manual load.

I agree about removing ocaml-java and ReScript.

The SEO changes for the docs looks good to me. If you want, we can merge the `latest` changes immediately ahead of 4.12, or do it at the same time as the release, as you prefer.

Octachron:
I think a good compromise would be to list the various `ocaml-option-*` in a new section between `What's new` and `Source distribution`. The current layout put too much emphasis on the switch list in my opinion.

avsm:
Agreed.

Octachron:
As far as I am concerned, this is ready to be merged.

ghost:
Maybe the source download links can point directly to the assets hosted on the github release page, rather than uploading it to the ocaml.org website?

Issue #1237, opened 2021-02-09
Footer width on mobile


patricoferris:
This would be a **good first issue** for anyone looking to get started improving and contributing to OCaml.org.

Currently when viewing the site on a mobile device (or from some quick testing anything less than `768px` in width) the footer no longer stretches completely across the bottom leaving some white margins on either side. I believe it would look better if it (like on larger screen sizes) stretched completely across the bottom. 

### Before 

<img width="329" alt="Screenshot 2021-02-09 at 11 59 30" src="https://user-images.githubusercontent.com/20166594/107360755-45b18080-6ace-11eb-983a-30163cb81ffa.png">

### After 

<img width="318" alt="Screenshot 2021-02-09 at 12 06 50" src="https://user-images.githubusercontent.com/20166594/107361450-4d255980-6acf-11eb-9b73-3fcf745f3167.png">

It may be useful to look into how other full width elements get around this problem such as the navbar at the top.

Shubham-Vishwakarma:
@patricoferris 
I would like to analyse this first, before taking it up
I am very much new to ocaml
But would like to give it try

Shubham-Vishwakarma:
I think I am able to find the issue

In bootstrap.css, there is padding-left and padding-right defined for smaller screen

@media (max-width: 767px) {
  body {
    padding-left: 20px;
    padding-right: 20px;
  }
....... }

If we remove this the footer will be fixed, but other elements will be affected

We could override this in ocamlorg.css where other footer.css are defined
This will fix the issue

@media (max-width: 767px) {
  #footer {
    margin-left: -20px;
    margin-right: -20px;
  }
..........}


Please let me know if this solution will be acceptable, and it's correct way to do it
If everything looks good, please let me know, I will raise a PR

patricoferris:
Hi @Shubham-Vishwakarma thanks for the contribution 🎉 

This is looking good to me, in fact it is how the navbar does it https://github.com/ocaml/ocaml.org/blob/master/site/css/bootstrap.css#L6708-L6713 so I don't think anyone would have anything against this. 

A good next step will be the raise that PR then we can push it to `staging` to double-check everything looks good and then hopefully merge!

Thanks again for doing this.

Shubham-Vishwakarma:
@patricoferris I have created PR to staging branch
Please let me know, if this solves the issues

Thanks for guidance,

Shubham-Vishwakarma:
My PR shows error
Will you please guide me.
[...]

PR #1239, opened 2021-02-25
Update dockerfiles to ocaml/opam and latest opam-repo commit

patricoferris:
This PR: 

 - Changes the source of the docker image from the deprecated `ocurrent/opam` to the continuously updated `ocaml/opam` image. This is needed because if the build misses the cache it is using an old `apt-get` package list (see this failing build https://deploy.ocamllabs.io/job/2021-02-25/160625-ocluster-build-8d226d thanks to @Shubham-Vishwakarma for bringing it to my attention in #1237).
 - Bumps the commit of the opam-repository to the latest, because the new base-image is using `4.10.2` which isn't in the old checkout of the repo that is currently being used. 

Seemed to build the docker files locally for me. 

cc @avsm


avsm:
Spot on! Thanks

Issue #1236, opened 2021-02-09
Small Improvements for Docs Page


patricoferris:
This would be a good first issue for anyone looking to get started improving and contributing to OCaml.org. 

### IFrame Titles 

Using [`pa11y`](https://github.com/pa11y/pa11y) to perform an accessibility audit on the https://ocaml.org/docs page using the [WCAG2AA](https://www.w3.org/TR/WCAG21/) standard revealed a some small improvement that can be made: 

 - The videos (which are `iframe` elements) should contain a descriptive `title` attribute -- you can read more about this [from Google's lighthouse automation tool](https://web.dev/frame-title/?utm_source=lighthouse&utm_medium=devtools). Note that titles will also need to be added to the `fr` page.

### Mobile Layout

Whilst using the page on mobile I noticed the buttons linking to the OCaml manual appear a bit oddly. Currently they overlap with the title of the following section leading to a somewhat confusing layout. This image should convey this: 

<img width="335" alt="Screenshot 2021-02-09 at 11 25 02" src="https://user-images.githubusercontent.com/20166594/107357019-7511be80-6ac9-11eb-8a7a-1d0ddafa9a72.png">

It would be nice if this was cleaner and well-separated on mobile, but retained its layout for desktop which I think works nicely.

nibble0101:
@avsm  Can I work on this? Pretty much my first time here.

avsm:
@nibble0101 please do! Just ping on here if you run into any blockers trying to fix it.

Shubham-Vishwakarma:
@nibble0101 Are you working on this?
If not please let me know
I would like to give it a try

nibble0101:
@Shubham-Vishwakarma  I am working on it.

nibble0101:
@avsm  I have managed to clone this repository and set up the project on my computer. Is there a script I can run to generate the html file to display in the browser instead of the markdown preview?

patricoferris:
Hi @nibble0101,

Just jumping in to try and unblock you quickly. The site uses a Makefile to run the necessary scripts to generate the HTML version of the site. This actually might be a great opportunity for you to try out the WIP [contributing document](https://github.com/ocaml/ocaml.org/pull/1241) that I am drafting and provide feedback on it? 

The summary of it is you need to: 
 - make sure you have `opam` installed (links in that document) and `make`
 - run `make deps` to get the dependencies for this project 
[...]

PR #1248, opened 2021-03-05
Fix HTML validation errors

nibble0101:
This PR fixes HTML validation errors emitted when the [documentation page](https://ocaml.org/docs/) is run through the [HTML validator](https://validator.w3.org/).

Running `make` generates the `ocaml.org` directory but doesn't "relativise" the value of `action` attribute of the `form` elements which have been added. Having an `input` element of type `button` nested within an  `a` tag is not valid. As a result, I have modified  such `input`s  to be nested inside `form` elements.   I suspect a script which generates the HTML files from markdown needs to be worked on to fix this. @patricoferris

### DIFF
--- a/site/css/ocamlorg.css
+++ b/site/css/ocamlorg.css
@@ -145,7 +145,7 @@ h1 img.rss {
 .documentation-video iframe{
   border: 0;
 }
-.documentation-form{
+.doc-link-wrapper{
   display: flex;
   justify-content: flex-end;
 }
--- a/site/css/ocamlorg.css
+++ b/site/css/ocamlorg.css
@@ -142,7 +142,13 @@ li p {
 h1 img.rss {
   height: 1.3ex;
 }
-
+.documentation-video iframe{
+  border: 0;
+}
+.documentation-form{
+  display: flex;
+  justify-content: flex-end;
+}
 .syndication img {
   height: 1.5ex;
   margin-left: 0.3ex;
--- a/site/docs/index.fr.md
+++ b/site/docs/index.fr.md
@@ -29,11 +29,10 @@
 	programmation OCaml, avec le système des modules, les
 	objets, le polymorphisme, etc.
     </p>
-	  <form id="tutref_b" action="/manual/index.html#sec6" class="documentation-form">
-	    <input type="submit" class="btn btn-default"
-		  name="button-tut"
-		  value="Lire les tutoriels">
-	</form>
+	<p class="doc-link-wrapper">
+        <a id="tutref_b" href="/manual/index.html#sec6" class="btn btn-default">
+			Lire les tutoriels</a>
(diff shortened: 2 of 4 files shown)


patricoferris:
Thanks @nibble0101 !

I'm wondering if the use of `form` and `input` are correct for this situation. I appreciate the previous implementation was just an `input` in an `a` which as you pointed out is not valid, but also these "buttons" are essentially just links to the manual/API. Would something like the following work: 

```html
<a href="/manual/index.html#sec6" id="tutref_b" style="float:right;" class="btn btn-default">
   Read the tutorials
</a>
```

It seems simpler and we retain the `relative_urls` functionality. Am I missing something?

nibble0101:
I do agree what you are suggesting is the most  straight forward way. I went with the `form` implementation because I wasn't quite sure why the `input` was wrapped in `a` tag in the very first place. From what I have observed so far, something as simple as formatting a markdown file can break code elsewhere in the application. I therefore wanted to maintain the `input` and give it a wrapper to fix the [button alignment issue](https://github.com/ocaml/ocaml.org/issues/1236). If removing the `input` doesn't break anything, I will remove the `input` and style the `a` tag to look like a `button`.

To fix button misalignment in mobile view, I believe it is better to remove `style="float:right;"` and give the `a` tag a wrapper and apply some styling to the wrapper otherwise I will have to come up with a *hacky* solution and *hacky* is not good either.

patricoferris:
Sounds good to me @nibble0101 👍

nibble0101:
@patricoferris  I have fixed the issues you raised in the review.

avsm:
Many thanks for your contribution @nibble0101! This has been pushed to the `staging` branch and will appear shortly on https://staging.ocaml.org/ as a preview. If you could verify your fix appears correctly there with a comment here and/or run the staging site through the online validator, we can merge this PR and have it go live.

avsm (approved):

nibble0101:
Thanks @avsm. Everything looks good at the staging area.
image

avsm:
Wonderful, thank you for the contribution @nibble0101! Now merged and live on ocaml.org.

patricoferris:
@nibble0101 thanks for the contribution :)) 

It seems this changed a few parts of the layout that we missed. For example there is no `.condensed > h2.ruled` so currently the site has the following docs page:

Screenshot 2021-03-11 at 09 17 24

Rather than the original:

Screenshot 2021-03-11 at 09 17 38

The button layout looks much better imo, but I think we want the ruled parts back again as it makes the page easier to parse. Would you be up for fixing this, if not I can hand it on to others so don't feel obliged :))

nibble0101:
@patricoferris  Sure. I can get straight to fixing it. Should I open a new PR?

Issue #1247, opened 2021-03-03
Git log of contributors is empty


patricoferris:
Currently, the site has a [contributors page](https://ocaml.org/contributors.html) and as part of that page there should be a list of authors extracted for them git log, but as you can see from the screen-shot 

![Screenshot 2021-03-03 at 11 56 59](https://user-images.githubusercontent.com/20166594/109802340-96248580-7c17-11eb-8b24-6cf847d7bb0f.png)

that list on the production site is currently empty. However, when developing locally at least on my machine (`make production`) and serving the site from the `ocaml.org` directory the git log is there. This makes me think there could be an issue in the Dockerfile used for deploying the site, so to tackle this issue you may need to try building the docker-ized site locally (`docker build . --file=Dockerfile.deploy -t ocamlorg-deploy`) and having a look at how we deploy the site using [OCurrent](https://deploy.ocamllabs.io/?repo=ocaml/ocaml.org).

patricoferris:
@IIITM-Jay would you be interested in looking at this?

IIITM-Jay:
@patricoferris Thank you so much for pinging me here. I am going through what you have suggested and asked above and will let you know soon.

IIITM-Jay:
@patricoferris why we are using opam-repository, in the following lines below :point_down: 

```RUN git -C /home/opam/opam-repository pull origin master && git -C /home/opam/opam-repository checkout c1952fb63c5a89c8c4230aac0360884518d72d91 && opam update -u -y```

can't we use the latest commit present in this repository itself like the way to build the site given in OCurrent deployer while building ```live_www_ocaml_org```

patricoferris:
Good question. 

The main reason, as far as I know, for this command is reproducibility and stability. The OCaml.org site doesn't have too many dependencies and certainly doesn't add or change dependencies very often. With this command, if the deployment fails, it's much easier to reproduce the exact environment locally to test it rather than trying to work out which commit of opam-repository was used. 

> like the way to build the site given OCurrent while building live_www_ocaml_org

I don't quite follow what you mean by this, the live deployment of the site does reset the opam-repository commit if I'm not mistaken. See:  https://deploy.ocamllabs.io/job/2021-02-26/175930-ocluster-build-c20284

IIITM-Jay:
[...]

PR #1253, opened 2021-03-07
Sorted out the Contributors list alphabetically and Optimize the command to display the same

IIITM-Jay:
## Issues 
1. The ***Contributors List*** is not displaying the names of the contributors in alphabetical order.
2. The ```sort``` command used to sort in ASCII order is not resulting in the expected output.
3.  The ```sed``` command used to print the same is not the flexible one and complicated too.
4. The following is the screenshot of the above discussed points:
Screenshot from 2021-03-07 10-50-55

5. The overall command is not the optimized one.
*(In Continuation with the issue #1247 )*
## Changes Incorporated
- [x] Made necessary corrections to the ```sort``` command in order to get the desired output.
- [x] Used ```awk``` instead of ```sed``` command to decrease complexity and to achieve optimization.
- [x] Used ```sort -u``` in place of ```sort | uniq``` . Thanks to @pw374 for pointing me about this.

## Resultant Output as expected
contributors

## Additional Changes
- [x] Appropriate changes  made to ***french version*** of the same.

### DIFF
--- a/site/contributors.fr.md
+++ b/site/contributors.fr.md
@@ -5,7 +5,7 @@ Contributeurs
 
 Les contributeurs à ce site :
 <ul>
-((! cmd git log --format="%aN" | sort -k 2 | uniq | sed -e "/^[^ ]*$/ d" -e "s|\(.*\)|<li>\1</li>|" !))
+((! cmd git log --format="%aN" | sort | uniq | awk '{print "<li>"$1,$2,$3"</li>"} !))
 </ul>
 
 Cette liste est obtenue depuis [le log GIT](https://github.com/ocaml/ocaml.org/commits/master), pensez à vérifier que votre
--- a/site/contributors.fr.md
+++ b/site/contributors.fr.md
@@ -5,7 +5,7 @@ Contributeurs
 
 Les contributeurs à ce site :
 <ul>
-((! cmd git log --format="%aN" | sort | uniq | awk '{print "<li>"$1,$2,$3"</li>"} !))
+((! cmd git log --format="%aN" | sort | uniq | awk '{print "<li>"$1,$2,$3"</li>"}' !))
 </ul>
 
 Cette liste est obtenue depuis [le log GIT](https://github.com/ocaml/ocaml.org/commits/master), pensez à vérifier que votre
--- a/site/contributors.md
+++ b/site/contributors.md
@@ -44,7 +44,7 @@ The contributors to this site, extracted from the
 [Git log](https://github.com/ocaml/ocaml.org/commits/master), are:
 
 <ul>
-((! cmd git log --format="%aN" | sort | uniq | awk '{print "<li>"$1,$2,$3"</li>"} !))
+((! cmd git log --format="%aN" | sort | uniq | awk '{print "<li>"$1,$2,$3"</li>"}' !))
 </ul>
 
 
--- a/site/contributors.md
+++ b/site/contributors.md
@@ -44,7 +44,7 @@ The contributors to this site, extracted from the
 [Git log](https://github.com/ocaml/ocaml.org/commits/master), are:
 
 <ul>
-((! cmd git log --format="%aN" | sort -k 2 | uniq | sed -e "/^[^ ]*$/ d" -e "s|\(.*\)|<li>\1</li>|" !))
+((! cmd git log --format="%aN" | sort | uniq | awk '{print "<li>"$1,$2,$3"</li>"} !))
 </ul>
 
 


IIITM-Jay:
@patricoferris May I request to kindly have a look. Thank you once again @patricoferris and @pw374 for guiding me for the same.

cc: @avsm

patricoferris:
Hi @IIITM-Jay,

Thanks for the contribution. A few points and questions. Whilst I appreciate the old `sed` command was not simple, please can I ask you to refrain from calling other people's work 'ugly' -- more useful comments would be 'too complex' or 'hard to maintain'. At the end of the day this is people's hard-work and we should be respectful about it :)) 

Unless I'm mistaken the old command sorted by surname, I don't mind changing that as you have in this PR and I agree that your solution is easier to read and understand. 

Have we gotten to the bottom of why this is not being seen in production though? 

Thanks again for your contribution :))

IIITM-Jay:
@patricoferris I am really very sorry for this, I never meant to hurt someone... I'm really sorry if this had brought to feel something like that 🙇.. updated the description as well..

Thank you and sorry once again.. 🙏

patricoferris:
No worries, I know you didn't mean any ill intent :)) 👍 

Your contribution is looking great! Did you get it building with the dockerfile by any chance? I'm still confused as to what is causing this not to show up in the deployed production site.

IIITM-Jay:
@patricoferris sorry to disturb at the moment especially in your weekends.

I want to discuss and share some of my experience regarding the issue of the production site not showing up the expected results. I have given a lot of time since last 4 days, searching for the error and the reason for why not building up the same , and finally got something that I can share with you.

While playing with the commands for the files that I have changed in this PR, I found some strange things. Some of the commands worked out really well for both **Terminals** as well as the **Local Build**. But some commands resisted to show the desired output on the local machine. 

Such incident I experienced with the command ```git shortlog``` which worked perfectly with the terminal but when I visited the HTML page it's all blanked and empty. This is due to its uncertain behavior with scripting languages. I haven't gone deeper into this because our main aim is to produce the output using ```sed``` or ```awk```.

But this incident made me to think that there can be the possibility that ***awk can work perfectly for the production site instead of sed***, though I am not sure but yes, if possible, can we give a chance to this PR to perform, may be it will go good. Sometimes ```awk``` goes exceptionally well as expected with the scripts in lieu of ```sed```. Also, when I was going through the differences it's mostly written that ```awk``` can work nicely with scripting languages and acts as powerful than ```sed```. In my opinion(I have a strong feeling though) that this issue might get resolved by using ```awk```.

Waiting for your valuable thoughts and opinions... :pray:
Thanking you.
-Jay

IIITM-Jay:
Hi, @patricoferris 
Any idea on above ....

patricoferris:
I must confess my `sed` and `awk` knowledge is zero :)) If you think it might just be fixed by this, then I see no harm in giving it a spin in `staging` to see if works there (as [you can see it didn't for the last commit pushed there](https://staging.ocaml.org/contributors.html)) this seems like a good next step to me. 

We'll get round to testing it in staging shortly.

patricoferris:
Ah, I think I've gotten to the bottom of this... sorry it probably wasn't the best `good-first-issue`. However, **I think your proposed command is more maintainable so hopefully we can still merge it shortly**.

For completeness sake let me quickly explain the problem.

We use a tool called [OCurrent](https://github.com/ocurrent/ocurrent) to deploy the site -- this takes care of building the site using the Dockerfiles and something called [OCluster](https://github.com/ocurrent/ocluster) to schedule the actual work to be done (in this case building and pushing the docker image).

As part of this, the build needs some notion of build context i.e. an environment in which to run the Dockerfile. There are some options which get associated with this and [you can see them here](https://github.com/ocurrent/ocluster/blob/master/api/docker.ml#L38-L43). The most important (for this problem) being the `include_git` parameter, the deployer uses the default which is [false](https://github.com/ocurrent/ocluster/blob/master/api/docker.ml#L55). Without `.git` in the build context our `git log` command whenever it is run will return nothing.

I'm submitting a PR shortly to enable the `.git` directory for our builds, this should hopefully fix things. Thank you for your perseverance and contributions.

avsm:
Excellent work on finding this issue @IIITM-Jay -- dealing with sed and awk and ASCII ordering isn't the easiest thing to figure out.  As a next step, I'll get the related ocurrent/ocurrent-deployer#46 PR merged and live, and then we can push your changeset to the staging site and see it working.

avsm:
Now pushed to staging.ocaml.org, where we should hopefully be able to spot the difference from the live ocaml.org site.

IIITM-Jay:
First of all thank you so much much...(must say 100 times),  @avsm for your kind words.

 And yes now I am able to see the difference between the contributors page at https://staging.ocaml.org/contributors.html   and the same on the https://ocaml.org/contributors.html.

Special thanks to @patricoferris for giving me his full support despite of his busy schedules ( I might guess) and giving me such a nice opportunity to work on this. I can't even imagine of doing this without his kind support and guidance.

Many many thanks to @pw374 too.

Thank you all... :pray:

avsm:
Thank you for checking the staging site, and now merged to live. We're all very glad to have your contribution @IIITM-Jay -- come back again soon with more PRs :-)

Issue #1215, opened 2020-11-13
Download link missing (or hard to find)


suany:
In https://ocaml.org/docs/install.html, under From Source, it says to "Download an OCaml Release", which links to https://ocaml.org/releases/, but none of the pages under releases has an obvious link to download a distribution. The distrib link is hidden in the "Earlier releases are available here" bullet at the bottom of releases/.

ghost:
So each link in the release notes should have two links?
`
OCaml 4.11.1(release notes link, source tarball link), released Aug 31, 2020.
`

Should the source download link be from inria or github releases?

avsm:
That's right -- two links is fine. The source download link should be from GitHub Releases.

rockingrohit9639:
Hii There, Can you please assign this to me?

patricoferris:
Hi @rockingrohit9639 👋 

Thanks for taking interest in this issue. There's no need for assignment to contribute :)) So if you want to open a PR and reference this issue that would be great! Feel free to ask any question on the PR or this issue if you are having trouble, we appreciate getting started isn't necessarily as straightforward as other sites.

victorphoenix3:
@rockingrohit9639  if you are not working on this issue may I submit a PR?

patricoferris:
Fixed in #1256

PR #1256, opened 2021-03-09
fix: download link missing from the releases page

meenal06:
fixes #https://github.com/ocaml/ocaml.org/issues/1215

- I have added a download link for each of the releases as requested.

### DIFF
--- a/site/releases/index.fr.md
+++ b/site/releases/index.fr.md
@@ -10,33 +10,33 @@ instructions pour installer OCaml par d'autres moyens que la
 compilation des sources, comme par exemple le gestionnaire de paquets
 OPAM et les gestionnaire de paquets spécifiques à une plateforme.
 
-* OCaml [4.12.0](4.12.0.html), publiée le 24 février 2021.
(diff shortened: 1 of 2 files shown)


meenal06:
Hi @avsm can you please review this pull request?

patricoferris:
Hi @meenal06 thanks for the contribution :))

A few points. I believe in the original issue you linked to we decided to use the Github releases as the download links (at least for the compilers that have them). You will need to check the individual pages to work out which ones are on there and which aren't.

In addition to this, there's also a french page that needs updating with the same information https://ocaml.org/releases/index.fr.html -- I believe the french word is téléchargé.

avsm:
Confirmed -- the caml.inria.fr site is deprecated, so the download links need to match the 'source tarball' on the linked page. For example, see https://ocaml.org/releases/4.10.2.html for the source tarball link for the 4.10.2 release.

meenal06:
@avsm @patricoferris  I have done the requested changes ,please have a look

avsm (approved):
Diff looks good, and now merged to live. Thanks for the contribution!

PR #1264, opened 2021-03-11
Bringing tutorials up to date - round three

johnwhitington:
This covers the "Data Structures" and "Errors" sections.

### DIFF
--- a/site/learn/tutorials/error_handling.md
+++ b/site/learn/tutorials/error_handling.md
@@ -68,7 +68,7 @@ Functions that can raise exceptions should be documented like this:
 ```ocaml
 val foo : a -> b
 (** foo does this and that, here is how it works, etc.
-    @raise Invalid_argument if a doesn't satisfy ...
+    @raise Invalid_argument if [a] doesn't satisfy ...
     @raise Sys_error if filesystem is not happy *)
 ```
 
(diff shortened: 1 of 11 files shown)


site/learn/tutorials/error_handling.md
```diff
@@ -71,7 +68,7 @@ Functions that can raise exceptions should be documented like this:
 ```ocaml
 val foo : a -> b
 (** foo does this and that, here is how it works, etc.
-    @raise Invalid_argument if a doesn't satisfy blabla
+    @raise Invalid_argument if a doesn't satisfy ...
```
avsm:
```suggestion
    @raise Invalid_argument if [a] doesn't satisfy ...
```

site/learn/tutorials/functional_programming.md
```diff
@@ -47,10 +47,7 @@ function as one of its arguments.
 So far so simple. If you're familiar with C/C++ then this looks like
 passing a function pointer around. Java has some sort of abomination
```
Octachron, line 48: This part is still true for `C`, but `C++ 2011` has moved to `std::function` arguments rather than raw function pointer.

johnwhitington: Thanks. Fixed in another PR. I hoped that all these PRs were seperable, but this file seems to appear in two. Let's merge them in numerical order, and fix the conflict as we go.

avsm (approved):
Lgtm; I'll merge this and any conflicts can be fixed in a followup

PR #1268, opened 2021-03-11
Bringing tutorials up to date - round six

johnwhitington:
This is the final PR of the first fix to the tutorials.

### DIFF
--- a/site/learn/tutorials/calling_c_libraries.md
+++ b/site/learn/tutorials/calling_c_libraries.md
@@ -3,11 +3,6 @@
 
 # Calling C libraries
 
-## Note from Rich
-Chapters 12 and 13 really need to be rewritten from scratch.
-
-I've left them here — you can read them, but the formatting is screwy.
-
 ## MiniGtk
 While the structure of lablgtk outlined in [Introduction to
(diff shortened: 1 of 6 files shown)


johnwhitington:
By the way, what do we want to do about the translations for these six pull requests? I could make a decent fist of doing the big deletions and rearrangements in the french version, but some of the individual sentence changes would be hard for me.

Do we have stats on the translated pages? How many people look at them? Is French the most viewed?

cc @gasche due to the monetary aspect

gasche:
I think that "what do we do with translations?" is a question for ocaml.org caretakers to answer. If their answer is that we should strive to keep them in sync, then of course you should feel free to spend the work on what you can do there.

(Maybe contacting the main contributors of the respective translations and asking them, maybe they would be interested in updating their work.)

avsm:
We don't currently have the stats on the usage of translations, but it's also not a popularity contest: if the translations are published, we should make a best effort to keep them up to date.  The only way to scale this is with maintainers for the sub areas. After the dust has settled on your content refreshes @johnwhitington, I can post to discuss and request a new round of translation assistance with a focus on the tutorial.

avsm (approved):

Issue #1258, opened 2021-03-09
Improve the Success Stories page


patricoferris:
There are probably other success stories we would want to add, but for this issue if we could tidy up the [success stories](ocaml.org/learn/success.html) page that would be great.

There are two main issues I can see (happy to have further improvements too): 

 1. On mobile devices, the logos/images associated with each success story could have a better layout to avoid small bits of text appearing beside them. For example:

<img width="308" alt="Screenshot 2021-03-09 at 09 39 34" src="https://user-images.githubusercontent.com/20166594/110452587-3c133c80-80bd-11eb-84d3-d182d5879350.png">

2. Some of the stories link to images that don't exist (e.g. the coq proof assistant) -- we should probably find and add these images folder.

@yashi-hub would you be interested in taking a look at this in relation to #1245? If yes or no please leave a comment below :))

yashi-hub:
hey @patricoferris. I would love to work on this issue. provided that I am just working on setting up the local project environment right now, so it might take me a day or two before I get started on working on this issue. is that fine?  thank you for assigning it to me. :)

patricoferris:
@yashi-hub take all the time you need, there's no rush :))

ghost:
This page might have the missing images: http://caml.inria.fr/about/successes.en.html

yashi-hub:
hey @patricoferris . as referred to in https://discuss.ocaml.org/t/error-while-setting-up-ocaml-project-using-wsl/7466/2 .
I tried to install everything again according to the guide https://www.cs.umd.edu/class/spring2018/cmsc330/ocamlInstallationGuide.pdf. but now when I run opam install ocamlfind I get the following error- 
![WhatsApp Image 2021-03-15 at 4 15 01 PM](https://user-images.githubusercontent.com/69522036/111142803-18636100-85ab-11eb-951f-06a45e02804e.jpeg)

I installed m4 using the command - sudo apt-get install -y m4.
and now when I run make, I'm getting the same error as in https://discuss.ocaml.org/t/error-while-setting-up-ocaml-project-using-wsl/7466.
continuing the issue here as requested by you to take it to the repo.
I'm unable to understand the steps required to move ahead for setting up the project, things seem to be very new to me.
hoping that I can soon start working on the issues :) :).

patricoferris:
Hi @yashi-hub 👋 

It seems the document I linked is a little out of date, this [guide seems better](https://github.com/janestreet/install-ocaml). 

[...]

PR #1286, opened 2021-03-22
add ocaml.org-media to docker images

patricoferris:
This PR clones the https://github.com/ocaml/ocaml.org-media repository into the build directory of the deploy and staging images, this fixes:

 - The dead links on pages like https://ocaml.org/meetings/ocaml/2011/ 
 - It also adds the missing images on pages like https://ocaml.org/learn/success.html (part of #1258)

Tested it locally and seemed to work -- would be good to push to staging to double check.


avsm:
Thanks!

Issue #1254, opened 2021-03-08
Improve the layout of the community page


patricoferris:
The [community page of the ocaml.org site]() could use an improved layout, particularly on mobile devices. Currently the first section containing the community icon, the introductory text, the logos and stickers link (and images) render a little awkwardly on smaller screens (see screenshot). 

<img src="https://user-images.githubusercontent.com/20166594/110302336-099c0d80-7ff1-11eb-9e70-21009a19a577.png">

Hopefully you can agree this looks awkward, it would be good to improve this. Some ideas are: 

 - To remove the community icon altogether
 - Ensure good layout of the two logos on smaller screens
 
I'm open to any other design options too so please feel free to suggest some. 

@Aditi-1400 is this something you'd be interested in doing as part of #1245? Please leave a comment if you are interested or if you are not :))

Aditi-1400:
Yes, I would definitely like to work on this one :).  I'll get started and get back to you if I have queries. 
Thank you :)

Aditi-1400:
@patricoferris 
So, to proceed with improving the layout, I built the website locally, I made a few changes in the styles and the end product looks like this.
![image](https://user-images.githubusercontent.com/62544124/110370615-ebc2bd00-8071-11eb-9e1c-728f6d540d79.png)
However before proceeding further, I have a few doubts, since it's my first time contributing to a large codebase.
1.I am not really familiar with makefile, but, if I am getting it right, after making the changes, I'll have to run the make command, copy the changes in style to site/css/ocamlorg.css and then create a pull request.
2.  How many changes can I make? For now, I've just added a few classes to the elements and added their styles in ocamlorg.css file itself with a comment to separate it from the rest. It looks like:
![image](https://user-images.githubusercontent.com/62544124/110374659-eddb4a80-8076-11eb-8e2a-6bfd29fe08b3.png)
3. Do I create a new issue for other suggestions that I have for the page, or should I just specify them here? 

Also, I haven't changed the fonts, it's my system font, but, for some reason, I am not getting Domine on my local build.

patricoferris:
Great work @Aditi-1400 🎉 

Responding to your questions: 

> 1.I am not really familiar with makefile, but, if I am getting it right, after making the changes, I'll have to run the make command, copy the changes in style to site/css/ocamlorg.css and then create a pull request.

[...]

PR #1296, opened 2021-03-23
Community page

Aditi-1400:
Modified the styles of community Page


avsm:
Thanks for this contribution @Aditi-1400!  It would be useful to have a little more motivation behind the PR -- what are you trying to fix with this (was there a broken layout before?) and link to the URLs that are affected. If you can include a screenshot of the before/after from your fixes, that would make the improvement more obvious too.  Once that's done, I can push the resulting change to the staging.ocaml.org site for more testing.  Thanks!

patricoferris:
Thanks @Aditi-1400 -- just for future reference, as Anil points out, it's always good to provide a little context. When managing lots of issues and PRs it can be hard to remember who is working on what. In this instance a link back to the original issue in the description of the PR would suffice.  All of the conversation is over on #1254 :))

avsm:
Confirmed on staging that this looks better on mobile now, so merging. Thanks for the contribution, @Aditi-1400!

image

PR #1301, opened 2021-03-26
Use a newer POSIX form when evaluating `opam`

diohabara:
Modified the styles of shell scripting.

### DIFF
--- a/site/docs/install.md
+++ b/site/docs/install.md
@@ -34,10 +34,10 @@ Example using the Bash shell and opam-2.0:
 ```
 # environment setup
 opam init
-eval `opam env`
+eval "$(opam env)"
 # install given version of the compiler
 opam switch create {{! get LATEST_OCAML_VERSION !}}
-eval `opam env`
+eval "$(opam env)"
 # check you got what you want
 which ocaml
 ocaml -version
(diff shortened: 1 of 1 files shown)


site/docs/install.md
```diff
@@ -34,10 +34,10 @@ Example using the Bash shell and opam-2.0:
 ```
 # environment setup
 opam init
-eval `opam env`
+eval "$(opam env)"
```
avsm:
```suggestion
eval $(opam env)
```

The equivalent to the backtick is this; no need for extra quoting.

site/docs/install.md
```diff
@@ -34,10 +34,10 @@ Example using the Bash shell and opam-2.0:
...
 # install given version of the compiler
 opam switch create {{! get LATEST_OCAML_VERSION !}}
-eval `opam env`
+eval "$(opam env)"
```
avsm:
```suggestion
eval $(opam env)
```

same as above.

avsm (changes requested):

avsm (approved):
Thanks for the contribution! The revised change looks fine to me, so merging to the live site.

PR #1303, opened 2021-03-26
A first hour with OCaml

johnwhitington:
This PR adds a basic tour of OCaml. It re-uses a few bits from older tutorials, but is mostly new. It aims to instruct and inform those interested in OCaml.

It works with just a basic OCaml installation or even just TryOCaml for the first 90%.

It should be fairly easy to review in the .md source.

I have yet to update the tutorial list to point to this new tutorial, but it would replace the old "Basics" and "Structure of OCaml Programs" tutorials from the "Getting Started" section.


site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,877 @@
...
+
+A recursive function is one which uses itself in its own definition. An OCaml
+function isn't recursive unless you explicitly say so by using `let rec`
+instead of just `let`. Here's an example of a recursive function:
+
+```ocamltop
+let rec range a b =
+  if a > b then []
+  else a :: range (a + 1) b;;
+```
```
gasche, line 186: I would find it natural to have a small input-output example included in the toplevel session, something like `let digits = range 0 9`.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,877 @@
...
+
+```text
+OCaml type  Range
+
+int         63-bit signed int on 64-bit processors, or 31-bit signed int on
+            32-bit processors
+float       IEEE double-precision floating point, equivalent to C's double
+bool        A boolean, written either 'true' or 'false'
+char        An 8-bit character
+string      An string (sequence of 8 bit chars)
```
gasche: A string

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,877 @@
...
+
+## Pattern matching
+
+Instead of using one or more `if` ... `then` ... `else` ... constructs to make
+choices in OCaml programs, we can use the `match` keyword to match on multiple
+possible values. Consider the factorial function:
+
+```ocamltop
+let rec factorial x =
+  if x <= 1 then 1 else x * factorial (x - 1);;
```
gasche: Nitpick: I find `x` a natural name for real numbers (in OCaml, floats), and `n` more natural for natural numbers / integers.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,877 @@
...
+
+```ocamltop
+map (map (fun x -> x * 2)) [[1; 2]; [3; 4]; [5; 6]];;
+```
+
+In fact, we can turn `*` or any other operator into an ordinary function by
+surrounding it with parentheses and spaces, to give the most concise version:
+
+```ocamltop
+map (map (( * ) 2)) [[1; 2]; [3; 4]; [5; 6]];;
```
gasche: I have mixed feelings about this last version. `( * )` is tricky, because `(*)` wreaks havoc. In general, infix operators are easy to misuse: `List.map (( + ) 2)` is fine, but `List.map (( - ) 1)` may not do what the user expects, with a similar confusion possible for `List.filter (( > ) 0)`.

If you do show this example, I think it's important the lambda-abstraction above be exactly equivalent to the partially-applied version: `(( * ) 2` is `fun x -> 2 * x`, not `fun x -> x * 2`. This would matter highly for a non-commutative operator.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,877 @@
...
+```ocamltop
+type person =
+  {first_name : string;
+   surname : string;
+   age : int};;
+
+let frank =
+  {first_name = "Frank";
+   surname = "Smith";
+   age = 40};;
```
gasche, line 474: Personally I think that the superior formatting style for records (because it gives readable line diffs) is:

```ocaml

type person = {
  first_name : string;
  surname : string;
  age : int;
};;

let frank = {
  first_name = "Frank";
  surname = "Smith";
  age = 40;
};;
```

but it's also not super-important to show version-control-friendly code in a first introduction.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,877 @@
...
+  | Yellow
+  | RGB of int * int * int;;
+let l = [Red; Blue; RGB (30, 255, 154)];;
+```
+
+Data types may be polymorphic and recursive too. Here is an OCaml data type for
+a binary tree carrying any kind of data:
+
+```ocamltop
+type 'a tree = Lf | Br of 'a tree * 'a * 'a tree;;
```
gasche: This may be a new concept for many readers, so I think it would make sense to have each constructor on its own line (as with `colour` just above) to make it easier to unpack/process.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,877 @@
...
+let all = total t;;
+
+let flipped = flip t;;
+
+t = flip flipped;;
+```
+
+Notice that we do not need to explicitly free memory for such data structures
+when we no longer need it: OCaml is a garbage-collected language, and will free
+memory for data structures when they are no longer needed.
```
gasche: It may help to point out which memory is now reclaimable and will be freed implicitly: the value of `flip flipped` in the last test, which is not reachable from the rest of the computation.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,877 @@
...
+```
+
+When an exception is not handled, it is printed at the top level:
+
+```ocamltop
+f 10 0;;
+```
+
+The other way OCaml deals with exceptional situations is to use the built-in
+polymorphic `option` type, which is defined as:
```
gasche: I would say "is by returning a datatype that represents either a correct result or an error, for example using the built-in polymorphic `option` type[..]". This leaves the door open for `('a, 'b) result` or home-grown result types.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,877 @@
...
+
+We can use exception handling to build an option-style function from one which
+raises an exception, the built-in `List.find` function (which finds the first
+element matching a given boolean test):
+
+```ocamltop
+let list_find_opt p l =
+  try Some (List.find p l) with
+    Not_found -> None;;
+```
```
gasche, line 601: Nowadays the better way to write this code is 

```ocaml
let list_find_opt p l =
  match List.find p l with
  | v -> Some v
  | exception Not_found -> None
```

I suppose that you are not planning to show this "match with exception" form in this short introduction, but then I maybe would not show the example at all, following the general idea of only showing code that we would consider "good" code in a real project. (Okay, we let the non-tail-recursive `append` version be.)

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,877 @@
...
+let print_number n =
+  print_string (string_of_int n);
+  print_newline ();;
+```
+
+We can look at the type of the built-in function `print_newline`:
+
+```ocamltop
+print_newline;;
+```
```
gasche, line 672: If you have the space, here may be a good place to point out that `print_newline;;` does *not* print anything, `print_newline ();;` does. To some beginners it may not be clear that the computations (and effects) of a function are only "performed" when the function is called, and that it can otherwise be passed around like any inert value.

gasche (commented):
Nice! See some minor comments inline.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,877 @@
...
+  for i = 0 to 39 do
+    let x_init = Random.float 1.0 in
+    let x_final = iterate r x_init 500 in
+    let y = int_of_float (x_final *. 480.) in
+      Graphics.plot x y
+  done
+done
+
+read_line ()
+```
```
gasche, line 824: Note: as currently written, none of the examples can actually be "tried", they are not valid toplevel programs (no `;;`) and they cannot be compiled either, due to the ambiguous use of expressions.

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: Fixed.

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: Done.

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: Fixed.

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: Fixed.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,877 @@
...
+
+We can work with floating-point numbers too, but we must write `+.`, for
+example, instead of just `+`:
+
+```ocamltop
+let average a b =
+  (a +. b) /. 2.0;;
+```
+
+This is rather unusual. In other languages integers get promoted to floating
```
Octachron: Later, you are using C as an example of "other languages", would it make sense to mention it here?

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: Ok, I added in the double semicolons but also tweaked the language so as not to give the impression double semicolons are part of the language.

gasche: Another approach would be to avoid stray expressions at the toplevel, using `let () = foo` (and then either `let _ = read_line ()` or `let () = ignore (read_line ())`, I prefer the latter). You don't need double-semicolons then because parsing is unambiguous, and actually I think that pasting this form in the toplevel (with just one final `;;`) works well.

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: Ok, I agree. I'll cut this one.

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: Ok, done.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,877 @@
...
+
+'c';;
+
+"Help me!";;
+
+```
+
+Each expression has one and only one type.
+
+OCaml works out types automatically so you will rarely if ever need to
```
Octachron: The `if ever` modal seems a bit too strong when explicit types are needed for module types, type-directed disambiguation, GADTs, higher rank polymorphism, and polymorphic recursion....

johnwhitington: Fixed.

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: Thanks. Fixed.

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: Good idea. Done.

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: I didn't want to delete this (because meaningful small exception examples are hard enough to find), so instead I added yours afterward.

johnwhitington:
Ok, all fixed. Thanks everyone.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,893 @@
...
+using [TryOCaml](http://try.ocamlpro.com), with no installation required.
+
+## Running OCaml programs
+
+To try small OCaml expressions, you can use an interactive top level, or REPL
+(Read-Eval-Print Loop). The `ocaml` command provides a basic top level (you
+should install `rlwrap` through your system package manager and run `rlwrap
+ocaml` instead to get history navigation.)
+
+The alternative REPL [utop](https://github.com/diml/utop) may be installed
```
avsm:
```suggestion
The alternative REPL [utop](https://github.com/ocaml-community/utop) may be installed
```

new homepage for utop

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,893 @@
...
+through [OPAM](/docs/install.html#OPAM) or your system package manager. It has
```
avsm:
```suggestion
through [opam](/docs/install.html#OPAM) or your system package manager. It has
```

lowercase preferred for opam

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,893 @@
...
+# #use "program.ml"
+```
+
+Note that `#use` is not part of the OCaml language proper; it is an instruction
+to the OCaml top level only.
+
+## Expressions
+
+Our phrase `50 * 50` was an expression, which evaluated to `2500`. OCaml told
+us that the type was `int`, an integer. To avoid repetition, we can give a name
```
avsm: This is the first mention of "type". It might be worth a sentence explaining that every value has a type in OCaml before using the term.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,893 @@
...
+call to `range` would have tried to look for an existing (previously defined)
+function called `range`, not the currently-being-defined function.
+
+## Types
+The basic types in OCaml are:
+
+```text
+OCaml type  Range
+
+int         63-bit signed int on 64-bit processors, or 31-bit signed int on
```
avsm, line 205: a subtlety perhaps, but it can also be 32-bits on js_of_ocaml. Perhaps note that this is for native code generation, but that might be too much info here.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,893 @@
...
+```ocaml
+type 'a option = None | Some of 'a
+```
+
+So we may write:
+
+```ocamltop
+let f a b =
+  if b = 0 then None else Some (a / b)
+```
```
avsm, line 591: It's a little odd not to mention type `result` here, which is the preferred way of not "gobbling" errors by returning a None with no further information.

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,893 @@
...
+let birthday p =
+  p.age <- p.age + 1;;
+```
+
+## The Standard Library
+
+OCaml comes with a library of useful modules which are available anywhere OCaml
+is. For example there are standard libraries for functional data structures
+(such as maps and sets) and imperative data structures (such as hash tables),
+and for making system calls. We use them by writing the module, followed by a
```
avsm:
```suggestion
and for interacting with the operating system. We use them by writing the module, followed by a
```

`Unix` is not quite as raw as making syscalls directly :-)

site/learn/tutorials/a_first_hour_with_ocaml.md
```diff
@@ -0,0 +1,893 @@
...
+List.iter print_length ["one"; "two"; "three"];;
+```
+
+You can find the full list of standard library modules in the
+[manual](/releases/latest/manual.html).
+
+## A module from OPAM
+
+Apart from the standard library, a much wider range of modules are available
+through the OCaml Package Manager, OPAM. You must have OCaml on your computer
```
avsm:
```suggestion
through the OCaml Package Manager, opam. You must have OCaml on your computer
```

avsm (commented):
Really nice work, John! Comments left in the review.

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: Yes, I think probably too much info.

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: Fixed. Thanks.

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: I just looked at the Stdlib documentation for Result, and it looks like we need to improve it -- there's really no explanation of the expected way to use the functions.

Do you think introducing just the result type without introducing the monadic stuff (which I don't want to do) is worth it?

avsm: It's difficult to use `Result.t` without pointing to one of the libraries that consume the type:
- Rresult: https://github.com/dbuenzli/rresult/blob/master/src/rresult.mli
- Core: http://dev.realworldocaml.org/error-handling.html#encoding-errors-with-result

It's probably worth deferring this to the "second hour with OCaml" indeed

avsm:
Pushed to staging.ocaml.org

site/learn/tutorials/a_first_hour_with_ocaml.md
johnwhitington: Ok. By the way, what do you think of this blog post?

https://keleshev.com/composable-error-handling-in-ocaml

I wonder if we might persuade him to let us reuse this content in another of the tutorials...

gasche: A minor note: this style (result monad with polymorphic variants for errors) was advertized in the [Catch me if you can](https://hal.inria.fr/inria-00432575) paper of Teller, Spiwack and Varoquaux back in 2008.

(My suspicion is that it didn't catch on because it is a bit too complex to convince programmers in simple ecosystems. But maybe the lack of traction was also due to the tools of the time (no opam, no standard "result" type, a camlp4 syntax extension instead of binding operators...), and it could have more success today.)

dbuenzli: A minor note on the minor note :-)  

This was actually discussed before the paper on the caml-list mailing list [here](https://inbox.ocaml.org/caml-list/F5231A73-DA8B-4FE8-8B7F-C478F1A65585@erratique.ch/). The page for the mentioned OSR is archived [here](https://web.archive.org/web/20090109050841/http:/cocan.org/osr/exceptionless_error_management).

avsm (approved):
Looks good to merge. @johnwhitington, would you be able to hook up the indexes first?

site/learn/tutorials/a_first_hour_with_ocaml.md
avsm: Sounds like a separate section on error handling in a future PR is appropriate with all these rich references. There's also a mini-guide to using errors with polymorphic variants over at https://mirage.io/wiki/mirage-3.0-errors.  But for now, this particular PR comment is resolved.

dbuenzli: FWIW there's also `rresult` [guidelines](https://erratique.ch/software/rresult/doc/Rresult.html#usage).

johnwhitington:
Thanks @avsm 

Done, although I haven't tested it yet, because merging master in broke my build:

```
File "http.ml", line 41, characters 4-23:
41 |   | `Permanent_redirect
         ^^^^^^^^^^^^^^^^^^^
Error: This pattern matches values of type [? `Permanent_redirect ]
       but a pattern was expected which matches values of type
         Cohttp__.Code.status_code
       The second variant type does not allow tag(s) `Permanent_redirect
make[1]: *** [script/rss2html] Error 2
make: *** [local] Error 2
```

I suppose I need to `make deps` again, or re-install something via opam? Any idea?

avsm:
Yes you need to `opam update` to pull in the latest version of cohttp; I merged in some changes to support cohttp 4.0 in #1309

avsm:
Also need to update the tutorial list over at https://staging.ocaml.org/learn/ to include the new content

johnwhitington:
Ah yes. Coming up...

johnwhitington:
> Yes you need to `opam update` to pull in the latest version of cohttp; I merged in some changes to support cohttp 4.0 in #1309

I updated and upgraded successfully with opam, but now I get:

```
File "code.ml", line 142, characters 7-28:
142 |     | [Omd.Html_block(_,_,o)] -> o
             ^^^^^^^^^^^^^^^^^^^^^
Error: This pattern matches values of type Omd.block_desc
       but a pattern was expected which matches values of type Omd.block
make[1]: *** [script/md_preprocess] Error 2
make: *** [local] Error 2
```

johnwhitington:
Ok, I fixed up the main learn page. I'm not sure about the selection chosen for the front page generally, but I'll look at that separately.

johnwhitington:
It looks like the issue above is due to a very recent version of omd where the AST has changed...

```
omd                     2.0.0~alpha1 A Markdown frontend in pure OCaml
```

avsm:
That's right -- the constraints on the `ocamlorg.opam` were updated: https://github.com/ocaml/ocaml.org/blob/master/ocamlorg.opam#L12

It would be quite nice if an opam workflow noticed that the constraints changed and had some way to prompt the user that the current packages no longer satisfied the build needs without an `opam update`. I've made a note to bring this up in the opam 2.2 design discussions.

avsm:
Verified in staging, and now live. Thanks @johnwhitington and all the reviewers!

PR #1309, opened 2021-03-29
Port to cohttp 4

mseri:


patricoferris:
Thank you for the fix! Hmm looks like we're bumping into the same problem as before -- for deployment (both production and staging) we reset the opam-repository  to a specific commit for example: https://github.com/ocaml/ocaml.org/blob/master/Dockerfile.deploy#L2. This means the deploy builds cannot find the latest and greatest cohttp https://deploy.ocamllabs.io/job/2021-03-29/080935-ocluster-build-34c27c -- @mseri could you bump these as part of this PR please? I'm guessing the reset is a reproducibility thing.

mseri:
Sure!

avsm:
Looks good to me. We need a CI test to use the latest opam-repo `master` branch to spot dependency breakages too. I'll queue that up on my CI list...

PR #1332, opened 2021-03-30
Added PR template

### LINKED ISSUE #1329: Issue and PR templates
It would be nice to have some good Issue and PR templates for contributors to use. There have been some stellar examples already from contributors that we could definitely borrow from as inspiration. For example

 - @nibble0101's PR for adding some borders https://github.com/ocaml/ocaml.org/pull/1262 -- the before and after screen shots are great in providing us with some notion that the work has been done and that locally it appears to be working. Also issues like https://github.com/ocaml/ocaml.org/issues/1315 go a long way in explaining the exact problem trying to be fixed.
 - @IIITM-Jay's PR for fixing the git log of contributors https://github.com/ocaml/ocaml.org/pull/1253 again the before and after along with careful explanations are fantastic and really help use understand what's going on. 
 - @nayaldivya's PR for improving accessibility https://github.com/ocaml/ocaml.org/pull/1252 follows the same idea and is very helpful.
 
I've just hand-picked a few here but hopefully you can see that by spending a little time on a PR's layout (or an issue) it can really help speed up the review process. 

--- 

We can use [Github's PR and Issue templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates) to do this.

TildaDares:
Fixes #1329

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,9 +1,16 @@
 # Issue Description
 
-Please include a summary of the issue. You can also include images of the issue if there are any.
+Please include a summary of the issue.
 
 Fixes # (issue)
 
 ## Changes Made
 
-Please describe the changes that you made. You can also include images of the changes made if there are any.
+Please describe the changes that you made.
+
+* **Please check if the PR fulfills these requirements**
+
+- [ ] PR is descriptively titled and links the original issue above
+- [ ] Before/after screenshots (if this is a layout change)
+- [ ] Details of which platforms the change was tested on (if this is a browser-specific change)
+- [ ] Context for what motivated the change (if this is a change to some content)
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,0 +1,9 @@
+# Issue Description
+
+Please include a summary of the issue. You can also include images of the issue if there are any.
+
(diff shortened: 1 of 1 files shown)


pull_request_template.md
```diff
@@ -0,0 +1,9 @@
+# Issue Description
+
+Please include a summary of the issue. You can also include images of the issue if there are any.
```
avsm: Thanks for this! It would be worth providing a little more detail on the nature of the information we'd like (based on the example PRs @patricoferris pointed out in #1329). For example:
- if there is a layout change, provide a before/after screenshot
- if there is a browser-specific change, provide details of which platforms the change was tested on
- if there is a change to some content, provide some context for what motivated that change

avsm (approved):
Thanks for this! Merging, and we can iterate on it within the repository for any future comments.

ghost:
This is really nice.

If you end up working on this further at a later point, I like taking screenshots at mobile and desktop sizes when I have the time when I create PRs.

TildaDares:
Thank you @kanishka-work. I'll keep that in mind

PR #1347, opened 2021-03-31
scripts/http: error input_value: not_a_binary_channel [Windows]

mseri:
This is addressing a windows only failure in the HTTP module.
You can see the error in the screenshot below

The compilation failure on windows is due to incompatibility with non-binary channels.

The issue was reported [here](https://discuss.ocaml.org/t/error-while-trying-to-set-up-ocaml-project-on-windows-10-via-cygwin-terminal/7567/12) and triaged [here](https://discuss.ocaml.org/t/error-while-trying-to-set-up-ocaml-project-on-windows-10-via-cygwin-terminal/7567/13).

Tested locally on mac, which did not and keep not having issues. Would be nice if somebody can actually test it on windows.
Feel free to close it if it is not relevant, I have decided to send the patch to ease the onboarding for the outreachy participants by trying to minimise potential setup issues.

### DIFF
--- a/script/http.ml
+++ b/script/http.ml
@@ -93,7 +93,7 @@ let get ?(cache_secs=cache_secs) url =
   let fn = Filename.concat (Filename.get_temp_dir_name ()) ("ocamlorg-" ^ md5) in
   eprintf "Downloading %s ... %!" url;
   let get_from_cache () =
-    let fh = open_in fn in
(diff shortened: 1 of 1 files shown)


avsm:
This looks strictly better, so I'll go ahead and merge this quickly without directly testing on Windows as it passes CI. Thanks @mseri!

PR #1348, opened 2021-03-31
script/http: don't leak response body

mseri:
With the port to cohttp 4.0 I had not noticed the appearance of many warnings about 'Cohttp_lwt: body not consumed - leaking stream!'. This commit ensures that unused response bodies are drained before following the redirects or logging errors and are not leaked.

See https://github.com/mirage/ocaml-cohttp/issues/730 for more details and the reference to the PR introducing the leak check.

Signed-off-by: Marcello Seri <marcello.seri@gmail.com>

### DIFF
--- a/script/http.ml
+++ b/script/http.ml
@@ -33,7 +33,10 @@ let rec http_get_and_follow ~max_redirects uri =
 
 and follow_redirect ~max_redirects request_uri (response, body) =
   let open Lwt in
-  match Cohttp.Response.status response with
+  let status = Cohttp.Response.status response in
+  (* The uncosumed body would otherwise leak memory *)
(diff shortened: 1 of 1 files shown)


script/http.ml
```diff
@@ -33,7 +33,10 @@ let rec http_get_and_follow ~max_redirects uri =
 
 and follow_redirect ~max_redirects request_uri (response, body) =
   let open Lwt in
-  match Cohttp.Response.status response with
+  let status = Cohttp.Response.status response in
+  (* The uncosumed body would otherwise leak memory *)
```
avsm:
```suggestion
  (* The unconsumed body would otherwise leak memory *)
```

mseri: oops :P

avsm:
thanks!

PR #1364, opened 2021-04-01
Made the sizes of the book images in the books and paper section equal

### LINKED ISSUE #1344: Improve Books and Paper section in https://ocaml.org/docs/ on screen sizes less than 768px
On screen sizes less than 768px, the sizes of the book images are not equal. I'd like to improve that section by making the first book image be the same size as the last two.

image

TildaDares:
# Issue Description

On screen sizes less than 768px, the sizes of the book images are not equal.

## Before

image

Fixes #1344 

## Changes Made

Added a class that sets the sizes of the book images to a fixed width on screen sizes less than 768px.

## After

image


gs0510:
Thanks for your PR @TildaDares! The image for Real World OCaml looks very pixelated, can you lower the size of all images rather than increasing it? Thanks!

TildaDares:
@gs0510 Can I look for an image of that book cover that isn't pixelated? Even with the small size it still looks pixelated.

gs0510:
Thanks @TildaDares! This LGTM :)

avsm:
Nice work @TildaDares! Pushed to the live site :)

TildaDares:
Thank you @avsm.

PR #1369, opened 2021-04-02
Added context to links in header and footer 

### LINKED ISSUE #1327: Link text is unclear
The text "Learn" in the nav-bar and the footer is unclear without context and may be confusing to screen readers.  You can add a special screen reader text than can help with accessibility.

Relevant code:
```
<a href="/learn/">Learn</a>
```

Srinithyee:
# Issue Description
Fixes #1327 
The text "Learn" in the nav-bar and the footer is unclear without context and may be confusing to screen readers. You can add a special screen reader text than can help with accessibility.
 If there's alt text that can help provide context for screen readers it will make the link more accessible.

## Summary of the issue.
The current code is not friendly to screen readers. The texts in the nav-bar and footer does not have context. 

## Changes Made
Although the issue focused on "Learn", I've fixed the other texts in the nav-bar and footer as well.
* **I have added relevant alt text to each text in the nav-bar and footer**

### DIFF
--- a/template/footer.mpp
+++ b/template/footer.mpp
@@ -12,7 +12,7 @@
 </div>
 <div class="column">
   <div class="entry">
-    <h1><a title="Official Documentation for Ocaml" href="{{! cmd script/link_of_lang ((! get filename !)) "/docs/" !}}">{{! cmd script/translate ((! get filename !)) Documentation !}}</a></h1>
+    <h1><a title="Official Documentation for OCaml" href="{{! cmd script/link_of_lang ((! get filename !)) "/docs/" !}}">{{! cmd script/translate ((! get filename !)) Documentation !}}</a></h1>
     <ul>
       <li><a href="{{! cmd script/link_of_lang ((! get filename !)) /docs/install.html !}}">{{! cmd script/translate ((! get filename !)) "Install" !}}</a></li>
       <li><a href="http://caml.inria.fr/pub/docs/manual-ocaml/">{{! cmd script/translate ((! get filename !)) "Manual" !}}</a></li>
--- a/template/footer.mpp
+++ b/template/footer.mpp
@@ -1,7 +1,7 @@
 <!-- footer							-*-html-*- -->
 <div class="column">
   <div class="entry">
-    <h1><a alt="Beginner's guide to OCaml" href="{{! cmd script/link_of_lang ((! get filename !)) "/learn/" !}}">{{! cmd script/translate ((! get filename !)) Learn !}}</a></h1>
+    <h1><a title="Beginner's guide to OCaml" href="{{! cmd script/link_of_lang ((! get filename !)) "/learn/" !}}">{{! cmd script/translate ((! get filename !)) Learn !}}</a></h1>
     <ul>
       <li><a href="{{! cmd script/link_of_lang ((! get filename !)) "/learn/taste.html" !}}" >{{! cmd script/translate ((! get filename !)) "Code Examples" !}}</a></li>
       <li><a href="{{! cmd script/link_of_lang ((! get filename !)) "/learn/tutorials/" !}}">{{! cmd script/translate ((! get filename !)) "Tutorials" !}}</a></li>
@@ -12,7 +12,7 @@
 </div>
 <div class="column">
   <div class="entry">
-    <h1><a alt="Official Documentation for Ocaml" href="{{! cmd script/link_of_lang ((! get filename !)) "/docs/" !}}">{{! cmd script/translate ((! get filename !)) Documentation !}}</a></h1>
+    <h1><a title="Official Documentation for Ocaml" href="{{! cmd script/link_of_lang ((! get filename !)) "/docs/" !}}">{{! cmd script/translate ((! get filename !)) Documentation !}}</a></h1>
     <ul>
       <li><a href="{{! cmd script/link_of_lang ((! get filename !)) /docs/install.html !}}">{{! cmd script/translate ((! get filename !)) "Install" !}}</a></li>
       <li><a href="http://caml.inria.fr/pub/docs/manual-ocaml/">{{! cmd script/translate ((! get filename !)) "Manual" !}}</a></li>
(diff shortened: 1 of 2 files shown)


nibble0101:
Hi @Srinithyee. Thanks for the PR.

I think it would be better if you provided a link to the corresponding issue/issues in the body of the PR rather than in the header. You can do it by using the `Fixes` keyword followed by `#` character and the issue number. For example  `Fixes #1327` will automatically link to issue `1327`. 

About the changes you have submitted, I noticed the  `alt` attributes are  on `a` tags. In the final generated HTML, will the `alt` attribute still be on `a` tags or it will be on `img` tags?  I am asking because as far as I know, `alt` is not valid as HTML attribute for `a` tag.  You can check the MDN documentation [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a).

Srinithyee:
> Hi @Srinithyee. Thanks for the PR.
> 
> I think it would be better if you provided a link to the corresponding issue/issues in the body of the PR rather than in the header. You can do it by using the `Fixes` keyword followed by `#` character and the issue number. For example `Fixes #1327` will automatically link to issue `1327`.
> 
I've made the asked changes. Do check if this is okay now. Thanks :)
> About the changes you have submitted, I noticed the `alt` attributes are on `a` tags. In the final generated HTML, will the `alt` attribute still be on `a` tags or it will be on `img` tags? I am asking because as far as I know, `alt` is not valid as HTML attribute for `a` tag. You can check the MDN documentation [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a).
>
Ah yes, looks like it won't be generated. I think I could use the 'title' attribute instead. I'll make the changes and resubmit. Thanks for the review :)

Srinithyee:
@nibble0101 
I've replaced 'alt' attribute with 'title'. Hope this should work as expected!

nibble0101:
Thanks @Srinithyee . I personally don't know any other way of improving the accessibility in this situation than making the text 
 within the  `a` tag more descriptive. Probably something like `Learn Ocaml` which in this case may not be possible  because we are dealing with a menu item. Using `title` attribute surely does help  but it also has its own shortcomings as described [here](https://www.a11yproject.com/posts/2019-02-15-creating-valid-and-accessible-links/#toc_What-about-the-title-attribute-). Let us wait and see what the other reviewers think.

Srinithyee:
> Thanks @Srinithyee . I personally don't know any other way of improving the accessibility in this situation than making the text
> within the `a` tag more descriptive. Probably something like `Learn Ocaml` which in this case may not be possible because we are dealing with a menu item. Using `title` attribute surely does help but it also has its own shortcomings as described [here](https://www.a11yproject.com/posts/2019-02-15-creating-valid-and-accessible-links/#toc_What-about-the-title-attribute-). Let us wait and see what the other reviewers think.

 Yes, I don't know any other way to make the link text clear either, we could wait for the other reviewers to give their opinions. But since @gs0510 opened the issue, I think we could ask her if this was what she was looking for or if she was expecting something more than this. I could then work on it, based on her suggestions.

gs0510:
Thanks for your PR @Srinithyee! @nibble0101  has a very good point wrt title, which I missed earlier. I think renaming the navbar items is not going to possible, so we don't really have an alternative here other than to use title. I think we can keep this change, until we come across a better alternative.

template/footer.mpp
```diff
@@ -12,7 +12,7 @@
 </div>
 <div class="column">
   <div class="entry">
-    <h1><a href="{{! cmd script/link_of_lang ((! get filename !)) "/docs/" !}}">{{! cmd script/translate ((! get filename !)) Documentation !}}</a></h1>
+    <h1><a title="Official Documentation for Ocaml" href="{{! cmd script/link_of_lang ((! get filename !)) "/docs/" !}}">{{! cmd script/translate ((! get filename !)) Documentation !}}</a></h1>
```
avsm:
```suggestion
    <h1><a title="Official Documentation for OCaml" href="{{! cmd script/link_of_lang ((! get filename !)) "/docs/" !}}">{{! cmd script/translate ((! get filename !)) Documentation !}}</a></h1>
```

correcting the spelling of "OCaml"

template/navbar.mpp
```diff
@@ -1,7 +1,7 @@
...
-              <li ((! ifdef learn class="active" !))><a href="{{! cmd script/link_of_lang ((! get filename !)) "/learn/" !}}">{{! cmd script/translate ((! get filename !)) Learn !}}</a></li>
-              <li ((! ifdef documentation class="active" !))><a href="{{! cmd script/link_of_lang ((! get filename !)) "/docs/" !}}">{{! cmd script/translate ((! get filename !)) Documentation !}}</a></li>
-              <li ((! ifdef packages class="active" !))><a href="https://opam.ocaml.org/">{{! cmd script/translate ((! get filename !)) Packages !}}</a></li>
-              <li ((! ifdef community class="active" !))><a href="{{! cmd script/link_of_lang ((! get filename !)) "/community/" !}}">{{! cmd script/translate ((! get filename !)) Community !}}</a></li>
-              <li ((! ifdef news class="active" !))><a href="{{! cmd script/link_of_lang ((! get filename !)) "/community/planet/" !}}">{{! cmd script/translate ((! get filename !)) News !}}</a></li>
+              <li ((! ifdef learn class="active" !))><a title=" Beginner's guide to OCaml" href="{{! cmd script/link_of_lang ((! get filename !)) "/learn/" !}}">{{! cmd script/translate ((! get filename !)) Learn !}}</a></li>
```
avsm:
```suggestion
              <li ((! ifdef learn class="active" !))><a title="Beginner's guide to OCaml" href="{{! cmd script/link_of_lang ((! get filename !)) "/learn/" !}}">{{! cmd script/translate ((! get filename !)) Learn !}}</a></li>
```

Removing an errant space from the title

template/navbar.mpp
```diff
@@ -1,7 +1,7 @@
...
+              <li ((! ifdef documentation class="active" !))><a title="Official Documentation for Ocaml" href="{{! cmd script/link_of_lang ((! get filename !)) "/docs/" !}}">{{! cmd script/translate ((! get filename !)) Documentation !}}</a></li>
```
avsm:
```suggestion
              <li ((! ifdef documentation class="active" !))><a title="Official Documentation for OCaml" href="{{! cmd script/link_of_lang ((! get filename !)) "/docs/" !}}">{{! cmd script/translate ((! get filename !)) Documentation !}}</a></li>
```

avsm:
Pushed to staging.ocaml.org to verify the titles work (they do!), and merging with fixes to minor typos. Thanks @Srinithyee for the contribution and @nibble0101 and @gs0510 for the reviews.

PR #1419, opened 2021-04-05
Delete tutorial content which is no longer linked

johnwhitington:
(This should not be merged until https://github.com/ocaml/ocaml.org/pull/1303 has been merged)

This PR deletes files containing out of date or superseded tutorials, none of which are linked from the main tutorial page, and none of which should be linked from other tutorials. A recent discussion pointed out that some of these out-of-date tutorials appear on first-page Google results, and so mislead searchers into wasting time on out-of-date technologies.

After this and https://github.com/ocaml/ocaml.org/pull/1303 are merged, I will propose another PR to clean up and erroneous links to this deleted content - there may be one or two around.


avsm (approved):
Farewell, ocamlbuild tutorial! It would be nice to have this preserved on https://github.com/ocaml/ocamlbuild, but can leave that to some hypothetical future ocamlbuild maintainer unless @gasche is feeling very motivated.

PR #1434, opened 2021-04-06
Fix the translation for 'Books' in /learn/index.fr.html #1408

### LINKED ISSUE #1408: Translation error in 'Learn' page.
The heading for 'Books' in the French translation of the page https://ocaml.org/learn/index.html is:
h
which translates to 'English and French books' but there are books in other languages too. I think the heading should be 'Livres' according to the online translator.

gurleennsidhuu:
# Issue Description

The translation for 'Books' in https://ocaml.org/learn/index.fr.html was written to be 'Livres anglais et français' which translates to 'English and French books'. The heading should be 'Livres' (books in en).

Fixes #1408

Before: 
Screenshot 2021-04-06 154526
After:
Screenshot 2021-04-06 154735

### DIFF
--- a/site/learn/index.fr.md
+++ b/site/learn/index.fr.md
@@ -45,7 +45,7 @@ OCaml est un langage générique de programmation, de puissance industrielle, qu
             </footer>
         </section>
         <section class="span4 condensed">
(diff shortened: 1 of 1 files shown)


avsm:
This looks right, given that there are non English and French books in the section. Thanks for the contribution!

PR #1449, opened 2021-04-06
Reinstate debugging tutorial

johnwhitington:
There is a decent debugging tutorial in `site/learn/tutorials/debug.md` which appears to be unused.

This PR includes a lightly edited version of it, and links it back into the tutorials.

PR #1451, opened 2021-04-06
Split modules tutorial in two

johnwhitington:
Modules are simple; functors are complicated. They currently exist in the same tutorial, and in an order of presentation which is confusing.

This PR splits it into two tutorials. I have also done a pass over them to improve them somewhat.

The functors tutorial could do with extending in the future.


johnwhitington:
(I appear to have dragged in an unrelated change to debugging. This will disappear once https://github.com/ocaml/ocaml.org/pull/1449 is merged)

avsm:
#1449 has been merged, so you should be able to rebase this freshly against the `master` branch now.

johnwhitington:
Ok, my Git knowledge has failed, so I think I'm going to dump this PR and represent it in a new one.

PR #1458, opened 2021-04-06
Changed all occurences of real-world-ocaml.jpg to real-world-ocaml.jpeg

### LINKED ISSUE #1450: Change all occurences of real-world-ocaml.jpg to real-world-ocaml.jpeg
Since I replaced real-world-ocaml.jpg with real-world-ocaml.jpeg in issue #1344, the real-world-ocaml image like the one in https://ocaml.org/learn/ is not being displayed.

Screenshot 2021-04-06 at 16 03 32

I'd like to change all img tags that are still referencing the old image to the new one.

TildaDares:
# Issue Description

Some images of the Real World Ocaml book were not being displayed.

Fixes #1450

## Changes Made

Changed all occurences of real-world-ocaml.jpg to real-world-ocaml.jpeg

### DIFF
--- a/site/docs/index.fr.md
+++ b/site/docs/index.fr.md
@@ -138,7 +138,7 @@
       <div class="row">
 	<div class="span2 documentation-book">
 	  <a href="https://realworldocaml.org">
-	    <img src="/img/real-world-ocaml.jpg" alt="Real Worl OCaml book">
+	    <img src="/img/real-world-ocaml.jpeg" alt="Real Worl OCaml book">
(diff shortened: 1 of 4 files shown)


avsm:
Thanks for the fix, @TildaDares! We intersected and I simultaneously applied this fix in #1456. I preferred to keep the original filename (with your updated image) rather than rename all the places where it might have been called from, since that also preserves the link for any external links to the image from outside of ocaml.org.

PR #1464, opened 2021-04-07
Change breadcrumb position and differentiate languages

### LINKED ISSUE #1378: Repositioning of breadcrumbs on non top level pages and differentiate languages
I suggest we move the breadcrumbs to the right of the page and also differentiate the languages

Capture

bfore

If it's a good issue, I can make the fix

dinakajoy:
# Issue Description
On pages with  left sidebar,  the breadcrumbs appear on the middle of the page which isn't a good user experience as users are used to seeing breadcrumbs on the left side of the page. Also the language are clustered and difficult to differentiate. This can confuse users on which one to select.

Fixes #1378 

This PR positions the breadcrumbs on the left side of the page
  
Before
before-breadcrumbs
  
After
breadcrumbs-after

This PR adds | character to languages to differentiate them
  
Before
bfore

After
before

Test Platforms:
- Mobile, Tablet and desktop devices using chrome devices
- Chrome and Firefox browsers
  
### DIFF
--- a/site/css/ocamlorg.css
+++ b/site/css/ocamlorg.css
@@ -458,6 +458,12 @@ ul.translations:after { content: "]"; }
   text-shadow: 0 1px 0 #ffffff;
 }
 
+.translations > li:not(:last-child):after {
+  display: inline-block;
+  content: "\00a0|\00a0";
+  font-weight: 600;
+}
+
 /* Accessibility Helpers */
 .visually-hidden {
   clip: rect(0 0 0 0);
--- a/site/css/ocamlorg.css
+++ b/site/css/ocamlorg.css
@@ -427,6 +427,8 @@ header + .core-running-header {
 /* Translations and breadcrumb */
 div.running-header {
   margin-top: -10px;
+  width: 100%;
+  box-sizing: border-box;
 }
 .breadcrumb {
   display: inline;
--- a/template/core.mpp
+++ b/template/core.mpp
@@ -2,10 +2,7 @@
 <div class="container">
   <div class="row">
     {{!x ifdef toc
-    <div class="span4"></div>
-    <div class="span8">
-      ((! input template/running-header.mpp !))
-    </div>
+    ((! input template/running-header.mpp !))
     <div class="span4">
       <nav id="nav-secondary">
         <ul class="nav nav-list">
--- a/template/running-header.mpp
+++ b/template/running-header.mpp
@@ -1,18 +1,23 @@
 <!--								-*-html-*- -->
-<div class="running-header">
-  {{!z ifndef nobreadcrumb
-  <div class="row">
-    <ol class="breadcrumb">
-      {{! cmd script/breadcrumb "((! get filename !))" !}}
-    </ol>
+
+<div class="row">
+  <div class="running-header">
+    {{!z ifndef nobreadcrumb
+    <div class="span7">
+     <ol class="breadcrumb">
+        {{! cmd script/breadcrumb "((! get filename !))" !}}
+      </ol>
+    </div>
+    <div class="span5">
+      <ul class="translations">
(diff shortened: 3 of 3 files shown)


pitag-ha:
Thanks for the PR! :) 

I didn't know @IkehAkinyemi was also changing the layout of the language selector. I thought they would only change its position and fix the overflow that was provoked by the change of position. So now, there is another [PR](https://github.com/ocaml/ocaml.org/pull/1466) also changing the layout of the language selector. I'm really sorry for that! I'm not saying any of them is better, I'm just pointing out that there's a conflict now.

0xull:
I'm sorry about the conflict in UI layout. I will make out more time making sure to read up on other issues raised to know the most recent update on the project. Just resume back from tooth surgery.

dinakajoy:
I made different commits for each changes. One for the breadcrumbs and one for the languages. You can simply discard the language commit and integrate the breadcrumb commit @pitag-ha 
Or should I revert the commit myself?

gs0510:
I think it's okay for now @dinakajoy, depending on whose PR gets merged first you might have to rebase. :)

pitag-ha:
> I think it's okay for now @dinakajoy, depending on whose PR gets merged first you might have to rebase. :)

The problem with that is that it's not only a conflict in code, but also in content :) @IkehAkinyemi's moves the language selector to the top right and fixes its size to avoid overflow on pages with lots of languages. If we merged that PR and afterwards merged this PR (just resolving the code conflict), we would get the overflow that @IkehAkinyemi is avoiding.

dinakajoy:
I think it's best to leave the language at it's current position (ie beneath the navbar) rather than close to the search bar. As the navbar looks great currently without the language. On mobile, the hamburger is in the middle which is not right.
Instead, I suggest we add the dropdown (and instead of using ... you can use select html tag so it shows the drop down arrow) to the language at it's current location. 
I will revert my commit and only leave that of the breadcrumbs so this pull request can be merged for the breadcrumb.
@pitag-ha and @IkehAkinyemi  please what do you think?

dinakajoy:
I just made a new PR. I renamed the branch for clarification thinking it will simply update the remote branch name but instead it directed me to create a new PR - #1508
This can be closed now. Thank you.

pitag-ha:
> On mobile, the hamburger is in the middle which is not right.

If I understand you correctly, I think that's not exactly right. The language selector would always be in the top right corner. So on mobile, it would be on the right to the hamburger menu. See [here](https://github.com/ocaml/ocaml.org/pull/1466#issuecomment-817240701). Does that sound better?

> can be merged for the breadcrumb

I don't fully understand the idea to move the breadcrumbs, so I'll leave that to other mentors :)

> and instead of using ... you can use select html tag so it shows the drop down arrow

Yes, I also think that the `...` design is improvable. Thanks for the select html tag idea, that might be a good solution! @IkehAkinyemi , have you tried how that would look like?

dinakajoy:
@pitag-ha here is what I am trying to explain - 
The proposed changes
Capture

What other platform looks like - 
Capture2
Capture3

I am just suggesting the language be side by side with the breadcrumb as it is now but with dropdown.
But if you prefer it on the the navbar, it's okay too.

Please check #1378 for the initial issue details raised for the breadcrumb.

pitag-ha:
@dinakajoy , thanks for your input! It's great to discuss different ideas! :)

I've just moved this discussion to the [PR](https://github.com/ocaml/ocaml.org/pull/1466) which it's about.

avsm:
This change looks fine, so I'll go ahead and merge this first so that the discussion about the dropdown bar can be figured out. Thanks @dinakajoy and @IkehAkinyemi for the collaborative discussion (and also hope you are feeling better from the tooth surgery @IkehAkinyemi!)

PR #1484, opened 2021-04-08
Updated dimension of colour-transparent-icon.svg to realign links in community

### LINKED ISSUE #1441: Realigning links in Contributors
@patricoferris 
I'd like to realign the two links to make them in line. It looks quite odd now.
image

Do let me know if I can fix this!

Srinithyee:
# Issue Description
To realign the two links to make them in line since it looked quite odd.

Fixes #1441 

## Before:
image

## After:
* Without aligning the top of the images
image
* After aligning the top of the images
image

## Changes Made
The dimensions of color-transparent-icon.svg and OCaml_Sticker.svg has been changed.

## Testing Platform
This has been tested on desktop and mobile devices.

### DIFF
--- a/site/img/OCaml_Sticker.svg
+++ b/site/img/OCaml_Sticker.svg
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- Generator: Adobe Illustrator 20.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
-<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" width="100px" height="100px" viewBox="0 0 150 150" style="enable-background:new 0 0 141.7 141.7;" xml:space="preserve">
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+     width="100px" height="100px" viewBox="0 0 150 150" style="enable-background:new 0 0 141.7 141.7;" xml:space="preserve">
 <style type="text/css">
 	.st0{fill:#FFFFFF;}
 	.st1{fill:#494545;}
--- a/site/img/OCaml_Sticker.svg
+++ b/site/img/OCaml_Sticker.svg
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- Generator: Adobe Illustrator 20.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
-<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
-	 viewBox="0 0 141.7 141.7" style="enable-background:new 0 0 141.7 141.7;" xml:space="preserve">
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" width="100px" height="100px" viewBox="0 0 150 150" style="enable-background:new 0 0 141.7 141.7;" xml:space="preserve">
 <style type="text/css">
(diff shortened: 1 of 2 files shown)


pitag-ha:
Thanks for your contribution, that already looks better! Have you seen how it looks like to also align the top of the two icons?

Srinithyee:
> Thanks for your contribution, that already looks better! Have you seen how it looks like to also align the top of the two icons?

@pitag-ha I haven't checked that out yet. But, yes, I'll try to get that right too :)

Srinithyee:
@pitag-ha I have made the asked changes. Do review and let me know if it's fine!

site/img/OCaml_Sticker.svg
```diff
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- Generator: Adobe Illustrator 20.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
-<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
-	 viewBox="0 0 141.7 141.7" style="enable-background:new 0 0 141.7 141.7;" xml:space="preserve">
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" width="100px" height="100px" viewBox="0 0 150 150" style="enable-background:new 0 0 141.7 141.7;" xml:space="preserve">
```
pitag-ha: Is there a reason why you wrote this in one line instead of two as it was before? In general, it's good to avoid very long lines. In particular, here I have to scroll to the right to see the whole line.

site/img/colour-transparent-icon.svg
```diff
@@ -2,7 +2,7 @@
 <!-- Generator: Adobe Illustrator 17.0.2, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
 <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
 <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
-	 width="165.543px" height="144.277px" viewBox="0 0 165.543 144.277" enable-background="new 0 0 165.543 144.277"
+	 width="100px" height="100px" viewBox="0 0 175 190" enable-background="new 0 0 165.543 144.277"
```
pitag-ha, line 5: You've changed the width-height-ratio here (and maybe in the other icon too): before width > height, now they're equal. That distortions the image. I'm not sure if that distortion is visible though.

pitag-ha (commented):
Apart from my two "nitpicking comments": LGTM, thanks! :)

site/img/OCaml_Sticker.svg
Srinithyee: No particular reason. I could make it two lines, if it makes it easily readable :)

site/img/colour-transparent-icon.svg
Srinithyee: I had that in mind too. But since we are dealing with svg here, the distortion isn't visible. Would you want me to change it? I'm not too sure about how I could change it though. Considering that the distortion isn't visible, should we let it be?

avsm: This looks a lot better already, thank you! I do think it's worth keeping the right aspect ratio, since this is a logo and the proportions matter.  If you could restore the same width/height ratio while keeping the existing alignment from your improvements, that would be perfect.

Srinithyee: @avsm Okay! I'll try to make the change :)

Srinithyee: image

@avsm This is the best I was able to come up with, when I made changes only in the parameters of the viewBox without altering the original height and width of the image. I think it's because the two SVGs are of dimensions by itself. Moreover, the width and height of "OCaml_Sticker" isn't given explicitly. What would you suggest my next steps to be?

avsm:
That looks like an improvement to me, so merging! Thanks for the contribution.

PR #1486, opened 2021-04-08
Made the function of the version selector on the documentation page more obvious

### LINKED ISSUE #1310: Documentation section: explain version selector
This would be a **good first issue** for anyone looking to get started improving and contributing to OCaml.org.

The Documentation section offers a drop-down to choose the OCaml version. Extract:

Screenshot from 2021-03-29 17-55-45

Selecting a version in the drop-down adapts the links in that section (such as <ins>OCaml tutorials</ins> as seen in the extract) to the selected version. But currently, visually nothing changes when selecting a version. So one might doubt if it has an impact at all. I suggest to make clear that and what impact it has.

Two possible approaches for that are the following:
- Either: add the selected version number to the text of every link that is affected by the selection; e.g. change <ins>OCaml tutorials</ins> to <ins>4.10 OCaml tutorials</ins> or, more precisely, to <ins><selected_version> OCaml tutorials</ins>.
- Or: add an explanatory text next to the drop-down explaining what it does.

Goodiec:
# Issue Description
When users select a version from the version selector on the documentation page, the changes that are made to the page such as the links to the resources changing are not visible so the users might be confused about what the version selector is really doing
Changes made by the version selector not visible
Fixes # (issue)
Fixes #1310 
## Changes Made
Added the selected version to all the link texts whose links change when a version is selected.
Created a function to get the selected option's value and add it in front of the affected link texts.

Before

ver-before

After

ver1
ver

For Firefox
fire2
fire1

### DIFF
--- a/site/css/ocamlorg.css
+++ b/site/css/ocamlorg.css
@@ -48,23 +48,23 @@
     background-color:lightgray;
 }
 
-#opamdoc-contents .expander {
-  width:1.5em;
-  height:1.5em;
-  border-radius:0.3em;
-  font-weight: bold;
-  font-family:Sans;
-  line-height:16px;
+#opamdoc-contents .expander { 
+  width:1.5em; 
+  height:1.5em; 
+  border-radius:0.3em; 
+  font-weight: bold; 
+  font-family:Sans; 
+  line-height:16px; 
 }
 #opamdoc-contents .expanding_sig { border-spacing: 5px 1px }
 #opamdoc-contents .expanding_sig td { vertical-align: text-top }
-#opamdoc-contents .expanding_include_0 td { vertical-align: text-top }
-#opamdoc-contents .expanding_include_1 td { vertical-align: text-top }
-#opamdoc-contents .expanding_include_2 td { vertical-align: text-top }
-#opamdoc-contents .expanding_include_3 td { vertical-align: text-top }
-#opamdoc-contents .expanding_include_4 td { vertical-align: text-top }
-#opamdoc-contents .expanding_include_5 td { vertical-align: text-top }
-#opamdoc-contents .expanding_include_6 td { vertical-align: text-top }
+#opamdoc-contents .expanding_include_0 td { vertical-align: text-top } 
+#opamdoc-contents .expanding_include_1 td { vertical-align: text-top } 
+#opamdoc-contents .expanding_include_2 td { vertical-align: text-top } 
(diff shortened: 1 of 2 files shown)


pitag-ha:
Thanks @Goodiec, that's great! :) Could you properly link the issue to the PR (i.e. write "Fixes #1310" instead of "Fixes # (issue) #1310")?

Goodiec:
> Thanks @Goodiec, that's great! :) Could you properly link the issue to the PR (i.e. write "Fixes #1310" instead of "Fixes # (issue) #1310")?

Done @pitag-ha, thank you.

patricoferris:
This is looking great, one suggestion is that perhaps we don't need to add the version in quite so many places. For example `Many latest tools are bundled...` doesn't quite make sense. The ones I think we could drop are:

 - `Many latest tools`
 - `check the latest language extensions`
 - `latest Standard Library` and `latest core library`
 - `The official latest OCaml Tutorial`
 
 Which I think is most of the ones in the paragraphs. The buttons could maybe read `Some Button Text (<ocaml-version>)`? This might help with the problem at the moment where we have `latest Read the tutorials` which would become `Read the tutorials (latest)`? What are your thoughts @Goodiec?

Goodiec:
> This is looking great, one suggestion is that perhaps we don't need to add the version in quite so many places. For example `Many latest tools are bundled...` doesn't quite make sense. The ones I think we could drop are:
> 
> * `Many latest tools`
> * `check the latest language extensions`
> * `latest Standard Library` and `latest core library`
> * `The official latest OCaml Tutorial`
> 
> Which I think is most of the ones in the paragraphs. The buttons could maybe read `Some Button Text (<ocaml-version>)`? This might help with the problem at the moment where we have `latest Read the tutorials` which would become `Read the tutorials (latest)`? What are your thoughts @Goodiec?

Hi @patricoferris, I think your suggestion makes sense. I think instead of putting the version in brackets for just the buttons as you suggested, we could use the same style for those in the paragraphs that we are not dropping as well. I was also thinking that maybe adding a tooltip on the hover of the version controller to say that relevant links in the page will change when a particular version is selected, what do you think?

Goodiec:
@patricoferris, made the changes based on your suggestion. Here's the image of how it looks:
ver

I didn't implement the tooltip I suggested but here's an image showing how that would look:
tooltip

I think the tooltip might be a bit too much too though, do you think it will be useful in addition to the change that has been made already?

Issue #1489, opened 2021-04-09
Offscreening the navbar menu when not required which solves the contrast issue of lighthouse


jyotibalodhi:
When we go to the `ocaml.org/learn/` and generating the lighthouse report, it gives an issue for the contrast ratio being too low between background and foreground colors as can be seen in the image below. I am solving the issue by off-screening the navbar menu when not in use and toggle it back when clicked. 
In the image below the contrast ratio issue can be seen on the right side as a lighthouse report.

![img1](https://user-images.githubusercontent.com/63946672/114132875-c8b25400-9922-11eb-8c08-83f142d59a1d.png)

I have tried to solve the issue by modifying the collapsed menu JS and making some changes in the bootstrap.css.

![img 2](https://user-images.githubusercontent.com/63946672/114133044-0b742c00-9923-11eb-9287-a8c25e8d2188.png)

@patricoferris @pitag-ha @gs0510 Kindly review the PR I am raising regarding the issue as I mentioned in the commet.

jyotibalodhi:
I have raised a PR #1490 solving this issue, kindly have a look!

PR #1490, opened 2021-04-09
Navbar low contrast ratio lighthouse issue resolved

jyotibalodhi:
# Issue Description

When we go to the `ocaml.org/learn/` and generating the lighthouse report, it gives an issue for the contrast ratio being too low between background and foreground colors as can be seen in the image below. I am solving the issue by off-screening the navbar menu when not in use and toggle it back when clicked.
In the image below the contact ratio issue can be seen on the right side as a lighthouse report.

Fixes # (issue)

#1489 

## Changes Made

I am solving the issue by off-screening the navbar menu when not in use and toggle it back when clicked.
In the image below the contrast ratio issue can be seen on the right side as a lighthouse report.

img1
 
I have tried to solve the issue by modifying the collapsed menu JS and making some changes in the `bootstrap.css`.

img 2

### DIFF
--- a/site/css/bootstrap.css
+++ b/site/css/bootstrap.css
@@ -6993,6 +6993,8 @@ body {
   .nav-collapse.collapse {
     overflow: hidden;
     height: 0;
+    position: relative;
+    right: 1000px;
   }
   .navbar .btn-navbar {
     display: block;
--- a/site/css/bootstrap.css
+++ b/site/css/bootstrap.css
@@ -6992,9 +6992,10 @@ body {
   .nav-collapse,
   .nav-collapse.collapse {
     overflow: hidden;
-    height: 0;
+    height: auto;
     position: relative;
     right: 1000px;
+    display: none;
   }
   .navbar .btn-navbar {
     display: block;
--- a/site/js/collapsed-menu.js
+++ b/site/js/collapsed-menu.js
@@ -3,13 +3,12 @@
 
 function toggleVisibilityMenu(id) {
     var e = document.getElementById(id);
-    if (e.style.height == "auto") {
-	e.style.height = "0";
+    if (e.style.display == "block") {
     e.style.right = "1000px";
-
+    e.style.display = "none";
     }
     else {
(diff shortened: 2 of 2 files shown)


gs0510:
@jyotibalodhi I can't really find a difference in the screenshot, can you also add a screencast so the change is more apparent? Thanks!

jyotibalodhi:
> @jyotibalodhi I can't really find a difference in the screenshot, can you also add a screencast so the change is more apparent? Thanks!

Hello @gs0510 Actually the difference is not on the mobile screen view it's on the lighthouse report. In the first screenshot, on the right hand side you can see there's a error saying background and foreground colours low ratio. That is happening because of the navbar technically being there.  And in the second screenshot I have resolved rhe issue by toggling the nav menu items. 

This issue was raised by @patricoferris in a previous issue which I was solving that was closed but this point could not be worked upon. So I thought to work on this seperately.

patricoferris:
Hi @jyotibalodhi, thanks! Seems like this does fix lighthouse issue from your images. I'm wondering what the accessibility implications of this are? The content is technically still on the screen somewhere right? So would screen-readers for mobile still see it (and read it)? These are open questions I don't actually know the answer to and was wondering if you do?

jyotibalodhi:
Hi @patricoferris,  I just read about correctly hiding the off-screen content and made them inaccessible using tabs from here - https://web.dev/offscreen-content-hidden/. Adding the `display: none;` should make it unreadable for screen readers. Updating the Implementation in a while :).

patricoferris:
Nice work @jyotibalodhi 👍 -- I think we could probably push this to staging to double-check soon.

jyotibalodhi:
Yes, I think that would be great @patricoferris ! 
Also it would be great if you can assign me some medium issues, as I am much more familiar with the codebase now. :)

avsm:
This fix looks good on staging (no visible changes, but Lighthouse improvement), so merging. Thanks @jyotibalodhi!

PR #1494, opened 2021-04-09
made a 404 page

### LINKED ISSUE #109: Make a 404 page?
Would it be worth making a 404 page?

Having it styled similarly to any of the other OCaml.org pages with some basic content that points people to the front page (or site index etc) might be more useful than getting the github 404.

gurleennsidhuu:
# Issue Description

Made the front end of 404 page.

Fixes #109

## Changes Made

Added a md file name 404page.

Screenshot 2021-04-09 183030

### DIFF
--- a/site/404page.md
+++ b/site/404page.md
@@ -1,0 +1,25 @@
+<!-- ((! set title Docs !)) ((! set documentation !)) ((! set nobreadcrumb !)) -->
+<style>
+#center{
+display: table;
+text-align: center;
+width: 100%;
+height:53vh;
+}
+#head{
+display: table-cell;
+vertical-align:middle;}
+#txt{
+color: #FF8C00;}
+#but{
+font-size: 15px;
+cursor: pointer;
+}
+</style>
+<div id="center">
+    <div id="head">
+        <font size=+10><b>404</b></font>
+        <h5 id="txt"><b>OOPS, PAGE NOT FOUND!</b> <br> The page doesn't exist or some other error occured.</h5>
+        <a href="https://ocaml.org/" target="_blank"><button style="font-family:Arial, Helvetica, sans-serif" type="button" id="but">GO HOME</button></a>
+    </div>
+</div>


pitag-ha:
That's nice! Does the page ever get displayed? :)

gurleennsidhuu:
> That's nice! Does the page ever get displayed? :)

As of now, I have just made the front end of the page. I don't think it will get displayed.

patricoferris:
Hi @gurleennsidhuu thanks for this, a couple of quick comments: 

 - I think the file needs to be called `site/404.md` so that the build process makes a `404.html` page which is what most file servers expect (including [caddy](https://caddyserver.com/docs/caddyfile/directives/handle_errors) which we use in [deployment](https://github.com/ocaml/ocaml.org/blob/master/Dockerfile.deploy#L13)).
 - Whilst technically I believe you are [allowed `<style>` elements in the body](https://www.w3.org/TR/html52/document-metadata.html#the-style-element), I've never seen it in practice (doesn't this cause the "[flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content#:~:text=A%20flash%20of%20unstyled%20content,before%20all%20information%20is%20retrieved.)" problem?). I think you'd be better moving these to `ocamlorg.css`. Maybe even better would be re-using some of the styles from the rest of the site we already use like buttons, leaving the font as the normal one etc. 
 
What do you think ?

gurleennsidhuu:
> Hi @gurleennsidhuu thanks for this, a couple of quick comments:
> 
> * I think the file needs to be called `site/404.md` so that the build process makes a `404.html` page which is what most file servers expect (including [caddy](https://caddyserver.com/docs/caddyfile/directives/handle_errors) which we use in [deployment](https://github.com/ocaml/ocaml.org/blob/master/Dockerfile.deploy#L13)).
> * Whilst technically I believe you are [allowed `<style>` elements in the body](https://www.w3.org/TR/html52/document-metadata.html#the-style-element), I've never seen it in practice (doesn't this cause the "[flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content#:~:text=A%20flash%20of%20unstyled%20content,before%20all%20information%20is%20retrieved.)" problem?). I think you'd be better moving these to `ocamlorg.css`. Maybe even better would be re-using some of the styles from the rest of the site we already use like buttons, leaving the font as the normal one etc.
> 
> What do you think ?

Sure. I will make the following changes. Thanks. :)

gurleennsidhuu:
> Hi @gurleennsidhuu thanks for this, a couple of quick comments:
> 
> * I think the file needs to be called `site/404.md` so that the build process makes a `404.html` page which is what most file servers expect (including [caddy](https://caddyserver.com/docs/caddyfile/directives/handle_errors) which we use in [deployment](https://github.com/ocaml/ocaml.org/blob/master/Dockerfile.deploy#L13)).
> * Whilst technically I believe you are [allowed `<style>` elements in the body](https://www.w3.org/TR/html52/document-metadata.html#the-style-element), I've never seen it in practice (doesn't this cause the "[flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content#:~:text=A%20flash%20of%20unstyled%20content,before%20all%20information%20is%20retrieved.)" problem?). I think you'd be better moving these to `ocamlorg.css`. Maybe even better would be re-using some of the styles from the rest of the site we already use like buttons, leaving the font as the normal one etc.
> 
> What do you think ?

I used the already existing class 'btn' for the 'go home' button, but as there were no classes for centering the text on the page, I had to create two new classes in ocamlorg.css. Is that okay?

0xull:
> I used the already existing class 'btn' for the 'go home' button, but as there were no classes for centring the text on the page, I had to create two new classes in ocamlorg.css. Is that okay?

Hello @gurleennsidhuu, this is a great feature.

Few suggestions if it's approved by the mentors, having 404 pages for other languages that are available on the ocaml.org website would be great, that makes it inclusive. 

Also with your recent changes on the `GO HOME` button, the text being strongly capitalized can be loosened a little by just capitalizing the initial, `Go home` or `Go Home`. 

If one thing I have learnt while working on this project would be to include screenshots or screencast of what your implemented changes/features look like on a desktop or mobile screen would help your PR review. Screenshots or a screencast of your recent changes would be helpful. (I'm just suggesting)

The word `occured` on the page is a typo, you could correct it to the right word, `occurred`.

gurleennsidhuu:
> > I used the already existing class 'btn' for the 'go home' button, but as there were no classes for centring the text on the page, I had to create two new classes in ocamlorg.css. Is that okay?
> 
> Hello @gurleennsidhuu, this is a great feature.
> 
> Few suggestions if it's approved by the mentors, having 404 pages for other languages that are available on the ocaml.org website would be great, that makes it inclusive.
> 
> Also with your recent changes on the `GO HOME` button, the text being strongly capitalized can be loosened a little by just capitalizing the initial, `Go home` or `Go Home`.
> 
> If one thing I have learnt while working on this project would be to include screenshots or screencast of what your implemented changes/features look like on a desktop or mobile screen would help your PR review. Screenshots or a screencast of your recent changes would be helpful. (I'm just suggesting)
> 
> The word `occured` on the page is a typo, you could correct it to the right word, `occurred`.

Hi @IkehAkinyemi. Thank you for your suggestions. :) I will fix the typos and I also feel that I should change the capitalization but I don't know other languages, so I will not be able to make 404 page in other languages.

0xull:
Concerning other languages, once this current PR is confirmed/merged, you can raise an issue to implement the 404 pages in other languages and let anyone who knows the other languages implement it.

Glad you found the suggestions helpful. Great work, fellow.

site/404page.md
```diff
@@ -0,0 +1,25 @@
...
+color: #FF8C00;}
+#but{
+font-size: 15px;
+cursor: pointer;
+}
+</style>
+<div id="center">
+    <div id="head">
+        <font size=+10><b>404</b></font>
+        <h5 id="txt"><b>OOPS, PAGE NOT FOUND!</b> <br> The page doesn't exist or some other error occured.</h5>
```
avsm, line 22: Generally, it's good practise NOT TO SHOUT AT USERS by using all capital letters. You can revise this by simply calling it "Page Not Found" -- the bold is enough to emphasise that portion of the text.  Similar comment for the "GO HOME" button below as well.

avsm (changes requested):

PR #1495, opened 2021-04-09
Rewrite of "Compiling OCaml Programs"

johnwhitington:
This PR is a rewrite of the "Compiling OCaml Programs" tutorial. It is the last piece of the important "Getting Started" section of the manual to be modernized.

One particular difficulty is the explanation of the connection between ocaml, ocamlfind, opam, and dune. I hope this needle has been threaded, but I welcome fixes.

After this is merged, there will need to be one more pass over the whole of the "Getting Started" section to ensure consistency.

(A small change is made to modules.md too, to avoid contradictory statements in two parts of the tutorials).

### DIFF
--- a/site/learn/tutorials/compiling_ocaml_projects.md
+++ b/site/learn/tutorials/compiling_ocaml_projects.md
@@ -5,7 +5,7 @@
 # Compiling OCaml projects
 
 This tutorial explains how to compile your OCaml programs into executable form.
-It addresses:
+It addresses, in turn:
 
 1. The compilation commands `ocamlc` and `ocamlopt` provided with OCaml. It is
    useful to learn these commands to understand OCaml's compilation model.
@@ -14,9 +14,10 @@ It addresses:
    about where libraries have been installed on your particular system. 
 
 1. Automatic build systems for OCaml, such as `dune`, which release us from
-   details of compiler command invocation.
+   details of compiler command invocation, so we never touch `ocamlc`,
+   `ocamlopt`, or even `ocamlfind`.
 
-In our [up and running tutorial](up_and_running.html) we jumped straight using
+In our [up and running tutorial](up_and_running.html) we jumped straight to using
 the automated build system `dune`. Now we shall look under the hood.
 
 ## Compilation basics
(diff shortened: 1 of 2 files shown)


site/learn/tutorials/compiling_ocaml_projects.md
```diff
@@ -4,144 +4,171 @@
...
+command line. You must also indicate the libraries on which these libraries
+depend. You must also pass the -I option to `ocamlopt` for each directory where
+they may be found. This becomes complicated, and this information is
+installation dependent. So we will use `ocamlfind` instead, which does these
+jobs for us.
+
+###  Using the ocamlfind front-end
+
+Using `ocamlfind` is often used for compiling any program or library that uses
+third-party OCaml libraries. Library authors themselves make their library
```
avsm, line 88: It might be worth connecting ocamlfind to the previous section: that is, we use ocamlfind to glue together compilation units (cmx, cmxa, cmi, etc) into a "library", which can be then be addressed using the ocamlfind library namespace.

site/learn/tutorials/compiling_ocaml_projects.md
```diff
@@ -4,144 +4,171 @@
...
-following packages: lablGL, sdl, sdl.sdlimage, sdl.sdlmixer, sdl.sdlttf.
-The bigarray package is needed by the sdl package, but we don't need to
-worry about that since `ocamlfind` knows it.
+This shows the list of package names, with their versions. Note that most
+opam packages install software using ocamlfind, so your list of ocamlfind
+libraries will be somewhat similar to your list of installed opam packages
+obtained by `opam list`.
 
-The command for compiling our program will be:
+The command for compiling our program using package `package` will be:
```
avsm: It's syntactically a little confusing to see `package` used twice here. Why not call it `pkg1` or `package1` instead?

site/learn/tutorials/compiling_ocaml_projects.md
```diff
@@ -4,144 +4,171 @@
...
+
+## Dune: an automated build system
+
+The most popular modern system for building OCaml projects is
+[dune](https://dune.readthedocs.io/en/stable/) which may be installed with
+`opam install dune`. It allows one to build OCaml projects from a simple
+description of their elements. For example, the dune file for our project might
+look like this:
+
+```shell
```
avsm: I usually use ```scheme for this

site/learn/tutorials/compiling_ocaml_projects.md
johnwhitington: Fixed.

site/learn/tutorials/compiling_ocaml_projects.md
johnwhitington: Thanks. Fixed.

I'm a bit concerned about the words "library" and "package" because they are overloaded ("ocamlfind library" but "-package" "-linkpkg", then "opam package" but people say "please install so-and-so library from opam") but I think we're a but stuck with this state of affairs.

avsm: I think that's a very good point to bear in mind as we curate the tutorials on ocaml.org itself. We can't control what the Internet says, but at least we can be consistent in our own definitions of what an "OCaml library" (findlib) and "OCaml package" (opam) are.

avsm:
Merging, thanks!

PR #1496, opened 2021-04-09
Fixed issue with focus on the book images

### LINKED ISSUE #1357: Issue with focus on clickable image
Noticed an issue in the Books column on the learn page. Clicking on the first image applies the style for focus on the second image. Clicking on the second image applies to the second and also the outline from the style displays off the outline of the images. This issue is shown in the images below:

issue2
issue2 1

This can cause users to be confused and actually doubt if they clicked on the right image and doesn't give the best user experience. If this issue is approved, I can go ahead to try to fix this.

Goodiec:
# Issue Description
Noticed an issue in the Books column on the learn page. Clicking on the first image applies the style for focus on the second image. Clicking on the second image applies to the second and also the outline from the style displays off the outline of the images. This issue is shown in the images below

The style for focus wasn't displaying properly for the book images.

Fixes # (issue)
Fixes #1357 

## Changes Made
Added a class to the affected images and then used CSS in the CSS file to fix the issue.

Remove the inline styling that was added to both images that were contributing to the issue and added new classes to display the images properly.
Before

before-focus
before-focus1

After

afterfocus-2
afterfocus-2 1

### DIFF
--- a/site/css/ocamlorg.css
+++ b/site/css/ocamlorg.css
@@ -48,23 +48,23 @@
     background-color:lightgray;
 }
 
-#opamdoc-contents .expander { 
-  width:1.5em; 
-  height:1.5em; 
-  border-radius:0.3em; 
-  font-weight: bold; 
-  font-family:Sans; 
-  line-height:16px; 
+#opamdoc-contents .expander {
+  width:1.5em;
+  height:1.5em;
+  border-radius:0.3em;
+  font-weight: bold;
+  font-family:Sans;
+  line-height:16px;
 }
 #opamdoc-contents .expanding_sig { border-spacing: 5px 1px }
 #opamdoc-contents .expanding_sig td { vertical-align: text-top }
-#opamdoc-contents .expanding_include_0 td { vertical-align: text-top } 
-#opamdoc-contents .expanding_include_1 td { vertical-align: text-top } 
-#opamdoc-contents .expanding_include_2 td { vertical-align: text-top } 
-#opamdoc-contents .expanding_include_3 td { vertical-align: text-top } 
-#opamdoc-contents .expanding_include_4 td { vertical-align: text-top } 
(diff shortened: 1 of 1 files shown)


Goodiec:
@pitag-ha, @gs0510, I Just noticed that the contents from the books section down on the docs page of the ocaml.org website do not get displayed properly on smaller screen sizes, there's way too much empty space on the right. Can I add a commit to this request to fix that or should I open an issue for it?

doc2
doc1

nibble0101:
Hi @Goodiec.

Thanks for the PR. I tried going through the changes you made but looks like some of them are a result of formatting the code. Is that correct? If so,  It would be best to always avoid committing formatting changes otherwise reviewing your code becomes difficult.

Goodiec:
> Hi @Goodiec.
> 
> Thanks for the PR. I tried going through the changes you made but looks like some of them are a result of formatting the code. Is that correct? If so, It would be best to always avoid committing formatting changes otherwise reviewing your code becomes difficult.

Hello @nibble0101, thank you for the review. Please could you elaborate more on this so I can make the corrections?

nibble0101:
@Goodiec  What i mean is that there seems to be some changes in your commit which are not related to the issue you are trying to fix though I could be wrong. From the top right corner, you will notice github is indicating you have changed 58 lines of code. You removed 23 lines and added 35 lines. Are all those changes fixing issue #1357 or some of them are as a result of formatting the code?

Goodiec:
> @Goodiec What i mean is that there seems to be some changes in your commit which are not related to the issue you are trying to fix though I could be wrong. From the top right corner, you will notice github is indicating you have changed 58 lines of code. You removed 23 lines and added 35 lines. Are all those changes fixing issue #1357 or some of them are as a result of formatting the code?

OK @nibble0101, I think it is adding the whitespace changes that my editor is making. I will look at how to stop that from happening, thank you!

avsm:
Merged to the live site. Thanks!

Issue #1353, opened 2021-04-01
Different button colors on different pages


dinakajoy:
I just noticed that the button color on other pages are green while it's grey on news page.
Accessibility wise, I think it needs to be fixed. Because from the home page down, users already have it registered in their mind that green means it's a button.

Button on other pages
![green](https://user-images.githubusercontent.com/39722740/113279070-99746500-92da-11eb-8a60-d0f9a035ed37.PNG)

Button on news page
![grey](https://user-images.githubusercontent.com/39722740/113279122-aa24db00-92da-11eb-8762-ddd6fef6633c.PNG)

I think it will create better user experience.
If it's a good fix, then I would have it fixed.

lindaokorie27:
@dinakajoy I truly feel the coloring of the read more buttons is misleading. It initially struck me as a disabled button till I clicked on it. 

Although I just noticed that the linking of the "Read More" button is not the same as the title to the articles they are attached to. Did you notice this?

dinakajoy:
Nothing has been said yet by the maintainers on it. 
Regarding the link, can you drop a screenshot please because it was okay when I tried it except when you try to open in new tab.

dinakajoy:
I think the 'read more' button on the news can be improved. It is an accordion not a button, so I  suggest we indicate a drop down arrow with 'read more' text so as not to confuse users.  
When 'Read More' is clicked,  it opens up the remaining content and still shows other news topic after it which is confusing as users think they are on a new page not the current page.

nibble0101:
@dinakajoy  Personally I don't mind the button background color though there is no harm in changing it to be consistent with other buttons as well. What concerns me most is its positioning. Someone reading through will attempt to click the `...` after encountering it.  I believe it would make a lot more sense to place it immediately after `...` to convey to the user there is a hidden content which can be revealed by clicking the button.

pitag-ha:
@dinakajoy , thanks for filing the issue! I agree that there's room for improvement, we just need to discuss the details. It should be implemented by someone who hasn't worked on 2 or more simple issue yet though :)

lindaokorie27:
@nibble0101 if the read more button is placed immediately after the `...` then it would be better it is styled more as a link.

yashi-hub:
hey @pitag-ha , till now I have submitted 2 PRs. 
one for a good-first-issue https://github.com/ocaml/ocaml.org/issues/1258
and improved contribution.md.
[...]

PR #1500, opened 2021-04-09
Changed the color of the news page buttons 

oma16:
# Issue Description

Different button colors on different pages

The button color on other pages are green and text color are white while on news page the button color is grey and text color is deep grey

Before
Imgur

Imgur

Fixes # (issue)

     #1353

## Changes Made

Changed the button color on news page to green & the text-color to white (Read more and Hide) and added a down arrow to the read more button and up arrow to the hide button and tested on a desktop

After
Imgur

Imgur


site/community/planet/index.md
```diff
@@ -1,42 +1,3533 @@
-<!-- ((! set title OCaml Planet !)) ((! set news !)) -->
-<!-- ((! set advertise_rss true !)) -->
+<!DOCTYPE HTML>
```
patricoferris: This isn't right unfortunately. I think you may have copied the HTML output back into the markdown file in order to change the look of the button.

The original markdown has: 

```
((! cmd script/rss2html --posts -n 30 !))
```

Which generates the posts, you will probably need to have a look at https://github.com/ocaml/ocaml.org/blob/master/script/rss2html.ml#L297-L320 to make the desired changes and still allow the page to build and get the latest news and not be static.

patricoferris (changes requested):
Thanks for the PR :)) 

The implementation here isn't correct as it will stop the news page from being updated. Please have a look at my comment.

site/community/planet/index.md
oma16: Noted sir, I will make the desired changes right away. Thanks

PR #1509, opened 2021-04-12
Added TOC and modified  learn/success.fr.md page

shreyaswikriti:
# Issue Description
fixes issue #1454
Please include a summary of the issue.

- Added TOC to learn/success.fr.md page.

- Added missing sections on the learn/success.fr.md page and made it similar to the English page.

Fixes # (issue)
Fixes #1454
## Changes Made
**Before**
Screenshot from 2021-04-12 12-31-11
**After**

Screenshot from 2021-04-12 12-16-30

* **Please check if the PR fulfills these requirements**

- [x] PR is descriptively titled and links the original issue above
- [x] Before/after screenshots (if this is a layout change)
- [x] Details of which platforms the change was tested on (if this is a browser-specific change)
- [x] Context for what motivated the change (if this is a change to some content)

Issue #1454, opened 2021-04-06
'Content side menu ' is missing in learn/success stories in french page 


shreyaswikriti:
french page 
![image](https://user-images.githubusercontent.com/57635473/113739705-aac7d200-971d-11eb-8650-ffea3247c9d3.png)

english page 
![image](https://user-images.githubusercontent.com/57635473/113739767-bc10de80-971d-11eb-82b7-3643ce3bbee4.png)
@gs0510

pitag-ha:
This seems like a good issue. You can go ahead and work on it :) I'm not sure how easy/hard it will be.

shreyaswikriti:
Thank you @pitag-ha. I'm working on it.

shreyaswikriti:
@pitag-ha, kindly review the PR attached.

Issue #1413, opened 2021-04-05
Improving success.fr.html


gurleennsidhuu:
The English and the French translated pages look quite different.
The English translated page has table of contents whereas the French page doesn't.
Also, some of the stories are missing on the French translated page.
The page also has some syntax errors like: 
![Screenshot 2021-04-05 165109](https://user-images.githubusercontent.com/64313783/113568939-2d259880-962f-11eb-9493-b368f55692fb.png)
If approved, I would like to work on this page to make it look like the English translated page. :)

pitag-ha:
@gurleennsidhuu , do you know French? :)

gurleennsidhuu:
> @gurleennsidhuu , do you know French? :)

No, I don't know French.

pitag-ha:
I've just seen that you're the same person who's working on #1315. Thanks for trying to improve the French pages!

> Also, some of the stories are missing on the French translated page.

Were you thinking about fixing this without speaking French? Or was it more like pointing it out in general?

> The page also has some syntax errors like:

I'm not finding the syntax error there. Could you point me to it? Also, the screen shot you've added is from the English page. Was that intentional?

> The English translated page has table of contents whereas the French page doesn't.

Good catch, thanks! I'd say that's worth fixing for the matter of consistency. How many simple issue have you already worked on?

gurleennsidhuu:
> I've just seen that you're the same person who's working on #1315. Thanks for trying to improve the French pages!
> 
> > Also, some of the stories are missing on the French translated page.
> 
> Were you thinking about fixing this without speaking French? Or was it more like pointing it out in general?
> 
> > The page also has some syntax errors like:
> 
> I'm not finding the syntax error there. Could you point me to it? Also, the screen shot you've added is from the English page. Was that intentional?
> 
> > The English translated page has table of contents whereas the French page doesn't.
> 
> Good catch, thanks! I'd say that's worth fixing for the matter of consistency. How many simple issue have you already worked on?

I think I should not work on this issue because I don't know French and also I have worked on many simple issues. I would prefer leaving it for other Outreachy aspirants who have not yet got any `good first issues`. :)

Lontchi12:
Good evening, Please i'm an outreachy applicant and i would like to work on this page if accepted, i don't have any good first issues yet and i'm very good in French too

pitag-ha:
[...]

PR #1520, opened 2021-04-14
improved the site/learn/success.fr.html file translated the whole pag…

Lontchi12:
…e into french and adjusted what was lacking

# Issue Description
Improving the success.fr.html page
Quelques succès – OCaml - Google Chrome 4_14_2021 4_12_17 PM

The success.fr.html page didnot match with the english version of the page and some sections were not included so i fixed it
Fixes # (issue)
French translation
Content included
## Changes Made

Quelques succès – OCaml - Google Chrome 4_13_2021 10_00_37 PM
Quelques succès – OCaml - Google Chrome 4_13_2021 10_01_01 PM
Quelques succès – OCaml - Google Chrome 4_13_2021 10_01_51 PM

### DIFF
--- a/site/css/bootstrap.css
+++ b/site/css/bootstrap.css
@@ -6410,7 +6410,7 @@ body {
     width: 352px;
   }
   .span5 {
-    width: 290px;
+    width: 510px;
   }
   .span4 {
     width: 228px;
--- a/site/learn/success.fr.md
+++ b/site/learn/success.fr.md
@@ -3,6 +3,29 @@
 # Quelques succès
 *Contenu*
 
+## Jane Street
+[<img src='/img/users/jane-street.jpg' alt='Jane Street'
+style='clear: left; margin-right: 10px' />](http://janestreet.com/technology/)
+
+Jane Street est une société commerciale propriétaire qui utilise OCaml comme 
+plateforme de développement principale. Notre opération se déroule à grande échelle, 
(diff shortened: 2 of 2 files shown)


Lontchi12:
worked on this issue https://github.com/ocaml/ocaml.org/issues/1413 Your reviews will go a long way to help me improve it Thank you

site/css/bootstrap.css
```diff
@@ -6410,7 +6410,7 @@ body {
     width: 352px;
   }
   .span5 {
-    width: 290px;
+    width: 510px;
```
patricoferris, line 6413: I don't think we want to be doing this for two reasons: 

1. The `bootstrap.css` file is part of a CSS framework, and these are defined like this for a reason. It allows users to build grid like structures in CSS easily without having to worry about pixel widths, changing this would break that.
2. It's always better to tackle one single issue in a PR. It makes it easier to revert changes should we need to and (as is the case here) we don't want this addition so the PR is at least blocked until this is removed.

site/learn/success.fr.md
```diff
@@ -3,6 +3,29 @@
 # Quelques succès
 *Contenu*
 
+## Jane Street
```
patricoferris, line 6: Thanks! I will try to find someone who speaks better french than me that can actually review this bit of the PR :))

patricoferris (changes requested):
Thanks @Lontchi12 -- a few changes and suggestions in the review

shreyaswikriti:
@patricoferris @Lontchi12, I have already worked on this issue refer to #1454 . Also, I have already given permission to work on this and have raised PR #1509. Sorry Just saw this.

PR #1527, opened 2021-04-15
Improved accessibility on the Books page

### LINKED ISSUE #1439: Add alt text for images on the OCaml website
There is no alternative text on most of the images on the OCaml website. The Books page is an example, no of the images has an alt text. Adding alt text to the images will help improve accessibility for the visually impaired. In cases where the link to the image isn't correct or was changed and not updated, just the broken image icon gets displayed which doesn't communicate the content of the image that is supposed to be there to the visually impaired as seen in the image below:

no-image-sym

If this is approved, I can go ahead and add descriptive alt texts to the images.

Goodiec:
# Issue Description
Noticed that some images on the OCaml website didn't have the alt attribute so I measured its accessibility using Lighthouse and I saw that images not having the alt attribute affected the accessibility score. Having an alt text provides screen readers and search engines with a textual description of what is on your page.

Fixes # (issue)
Fixes #1439 
## Changes Made
Added alt attribute with appropriate descriptions for the images, added alt attributes with an empty value for decorative images or images that have texts that already describe them or are just on the page for visual purposes, for example on the books page, with the name of the book above the image and an adjacent description of the book, there is no need to add an additional description in the alt attribute to avoid confusing the readers. With an empty alt, screen readers will ignore it, putting the empty alt is better than no alt attribute at all because some screen readers will announce the file name of the image when the alt attribute is not provided. One more thing that happens with adding alt attributes is that they are displayed in case images fail to load for one reason or the other which is better than having the broken image icon on the page. For images with empty alt, nothing gets displayed as shown in the after image.

In addition to adding the alt attribute to images, I am also correcting typos and links on the affected pages.

Before
alt-1

After
alt-2

### DIFF
--- a/site/css/ocamlorg.css
+++ b/site/css/ocamlorg.css
@@ -611,10 +611,14 @@ ul.translations:after { content: "]"; }
   .books-span{
     margin: 5px;
   }
+  .books-row{
+    display: flex;
+  }
+  .books-row div.span2{
+    padding: 5px;
+  }
 }
-.books-row{
-  display: flex;
-}
+
 @media (max-width: 480px) {
   #footer .column {
     width: 98%;
--- a/site/learn/books.md
+++ b/site/learn/books.md
@@ -5,14 +5,14 @@
 # Books
 ## Books in English
 ###  The OCaml System: Documentation and User's Manual
-<img src="/img/colour-icon-170x148.png" style="float: left; margin-right: 15px; margin-bottom: 15px;"></img>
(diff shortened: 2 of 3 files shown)


Goodiec:
Realized that I made a mistake in pull request #1496, added the classes that I styled to fix the issue in the .html file instead of the .md file. After the request was merged I looked at the live site and saw the issue was still there and I realized my mistake. Decided to fix that in this pull request (the third commit), hope this is alright. I'm sorry about this.

PR #1536, opened 2021-04-15
http: don't use lwt.ignore_result

mseri:
`Lwt.ignore_result` is a broken version of `Lwt.async`. The equivalent of ignore is to bind unit. See https://ocsigen.org/lwt/5.2.0/api/Lwt#VALignore_result
Also removes some unnecessary `open Lwt`


avsm:
good "catch" (geddit?). Merging :)

PR #1538, opened 2021-04-16
Detecting and deleting dead scripts

### LINKED ISSUE #1531: Spring clean up: detect and delete dead scripts
There are a couple of scripts that seem to be dead code, e.g. `ocamlapplet.bash` and `link_blog_doc_ext.sh`. I propose that someone goes through all scripts in the `script/` folder and finds and deletes the ones that don't have an impact anymore.

### How
There are different ways to check if a script still has an impact (on the build). One option would be to build the page as is and save the result, then remove the script and build the page again, and then make a recursive diff on the two results.

### Watch out
In the first run, you might wrongly detect scripts that are meant for development as dead code. Have a look if the scripts you detect are used for development.

dldisha:
# Issue Description
There are a couple of scripts in the codebase that seem to be dead code and don't have an impact on the development.

Fixes #1531

## Changes Made
I have deleted all such scripts which are not being used for the development anymore. For resolving the issue, I first removed one script at a time and build the project again. If the build was successful without the script then that particular script is not being used for the development and is a dead code.

In total, I detected 7 such scripts that are dead code. 


patricoferris (changes requested):
Thanks for the PR. 

I think you've removed a little too much :)) For example: 

 - `graph.ml` and `test.ml` are both linked in the [gtk tutorial](https://github.com/ocaml/ocaml.org/blob/master/site/learn/tutorials/introduction_to_gtk.md) so they have to stay. 
 - similarly with `objcache.ml` only in the [garbage collection](https://github.com/ocaml/ocaml.org/blob/master/site/learn/tutorials/garbage_collection.md) tutorial.
  - `absolute_urls` this one appears to be used in [Makefile.opam.ocaml.org](https://github.com/ocaml/ocaml.org/blob/master/Makefile.opam.ocaml.org), I'm not entirely sure of the deployment process for opam.ocaml.org, but if it does still use this file then we probably shouldn't be removing it either.
  - `link_blog_doc_ext.sh` is also an `opam.ocaml.org` one which I would be fearful of removing
  
 Which only leaves `ocamlapplet.bash` which quite possibly could be removed, although it would be great if you could provide more information why. When you said:
 
 >  If the build was successful without the script then that particular script is not being used for the development and is a dead code.
 
I think you need to dig a little deeper than this, you can try using tools like [grep](https://man7.org/linux/man-pages/man1/grep.1.html) (or `cmd+shift+f` is VSCode) to search for uses of these files, try understanding what they might be used for and see if there are any uses cases. Most importantly, it would be good to relay this information in the PR itself as it will make for much quicker reviewing. 

I appreciate the hard work so far, thanks :))

dldisha:
Thankyou @patricoferris for the review and the explanation. I'll study the scripts more deeply and get back to this asap.

dldisha:
grep was quite helpful :) Helped me to undersnatd a lot about the file system.

Coming to the issue, I think we may not remove `ocamlapplet.bash` too because it is linked to [Makefile.common](https://github.com/ocaml/ocaml.org/blob/master/Makefile.common); a template file being used to generate certain configuration and run a certain task if something goes wrong with automatic config.

PR #1542, opened 2021-04-19
Dark Theme added

### LINKED ISSUE #1431: Dark extension for code examples (editors snaps where code is written)
Programmers often like the dark theme. I found this eye-catching.
Screenshot from 2021-04-06 13-35-28
rather than a light theme, we can convert the editor to a dark theme. I'll start working on this if you like. @patricoferris

shreyaswikriti:
# Issue Description
Added functionality of Theme-switching between Light Extention and Dark extension.

- Added a toggle button on Navbar that allows Users to change themes (dark or light) as per their preference.
- Light mode is the previous website theme while a new dark theme is added.

- whatever users prefer theme, that theme will get saved even after closing the browser. Also, they can change themes anytime.

Fixes # (issue)
Fixes #1431 
## Changes Made

- Added a Js file and two CSS files for light and dark mode changes.
- Done some changes in Bootstrap CSS files.

Before 
Screenshot from 2021-04-20 01-27-02
Screenshot from 2021-04-20 01-27-23
Screenshot from 2021-04-20 01-27-42

### DIFF
--- a/site/releases/3.12.1.md
+++ b/site/releases/3.12.1.md
@@ -1,5 +1,5 @@
 <!-- ((! set title OCaml 3.12.1 !)) -->
-
+*Table of contents*
 # OCaml 3.12.1
 **License**<br />
  The OCaml system is open source software: the compiler is distributed
--- a/site/releases/3.12.1.md
+++ b/site/releases/3.12.1.md
@@ -1,5 +1,5 @@
 <!-- ((! set title OCaml 3.12.1 !)) -->
-*Table of contents*
+
 # OCaml 3.12.1
 **License**<br />
  The OCaml system is open source software: the compiler is distributed
--- a/site/css/bootstrap.css
+++ b/site/css/bootstrap.css
@@ -5597,6 +5597,7 @@ body {
 .home-feature > a {
   float: left;
 }
+
 .home-feature h1 {
   font-size: 24px;
   line-height: 36px;
@@ -5633,7 +5634,13 @@ body {
   padding: 25px 0;
 }
 
-
+#home-learn a {
+  text-decoration: none;
+}
+#home-learn a:hover {
+  text-decoration: underline;
+  color: #c45641;
+}
 #home-learn img {
   vertical-align: middle;
   margin-right: 10px;
--- a/site/css/bootstrap.css
+++ b/site/css/bootstrap.css
@@ -7,6 +7,65 @@
  *
  * Designed and built with all the love in the world @twitter by @mdo and @fat.
  */
+
+ :root {
+  --light: #ffffff;
+  
+  --dark: #81899b;
+}
+/* Toggle Button for theme-change*/
+
+.toggle {
+  position : relative ;
+  display : inline-block;
+  width : 50px;
+  height : 25px;
+  background-color: white;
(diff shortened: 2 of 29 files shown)


site/releases/3.12.1.md
```diff
@@ -1,5 +1,5 @@
 <!-- ((! set title OCaml 3.12.1 !)) -->
-
+*Table of contents*
```
patricoferris: Would it be okay if you could remove these table of contents change please?

They could go in a separate PR but it just adds a lot of noise to this one PR (I think you touched maybe 4 actual files whereas the count is 29!). It's often best to tackle one single problem in the PR for the sake of the reviewers :))

patricoferris (changes requested):
First of all @shreyaswikriti let me just say this is really *excellent* work, thank you for taking the time to start this endeavour. 

One problem I think we'll encounter with this approach is handling the flickering effect you get when the order of operations is: 
  1. Render HTML with CSS 
  2. Run JS at the end of the body which notices dark mode from local storage
  3. Updates the CSS being used causing the style to re-render
  
Thus creating the so-called "dreaded flicker" effect which is a little disconcerting for users. When I implemented a similar feature I found [Josh Comeau's](https://www.joshwcomeau.com/react/dark-mode/#our-first-hurdle) tutorial very instructive, maybe we could do something like this? What do you think?

shreyaswikriti:
Hii @patricoferris, I solved the problem of flicker pages. Thank you so much for your tutorial and time. Also, I apologize for the inconvenience caused by the two different commits. I'll update the PR soon. Again thank you.

shreyaswikriti:
Hii @patricoferris, I have resolved the issue of the flicker of the web pages while loading. Kindly check the PR.

patricoferris:
Thanks @shreyaswikriti, this is starting to get in shape :tada:. A few more points and questions: 

 - Afaict, the toggle button is rendered always as light mode before the JS updates it based on local storage. The problem with this is if the user has set `dark-mode` then when they arrive at the page there is an awkward and attention-drawing slide of the slider. It would be good to mitigate this. I think there are probably two possible solutions: (a) remove the toggle and replace it with a button that doesn't change based on the theme, for example a simple `toggle theme` button (it should be obvious to the user what theme they are using!) or (b) a slightly more complicated approach would be to have the JS programmatically add in the button once the local storage has been checked and we know what theme to select. 
 - On the homepage there are two svgs that barely show up for the `dark-mode` because they are dark themselves so they'll need to be changed or given a light background or something (this might be the case for other pages, it would be good to check this).
 - Have you run an accessibility audit on the changes? Some of the contrasts don't look strong enough to meet the WCAG standards.

This is a fairly big change so this may take a while to get it into a mergeable state but thank you for persevering.

shreyaswikriti:
For the first point, I would propose using this kind of button. This kind of button does not cause any delay in changing state while page uploading also I think it's more suitable for the OCaml website since the toggle button is kind of animated and we have static pages. Happy to hear your thoughts on it.
image
For the second problem, We have to replace those two svgs with light color. And as far as I know, only the home page has this dark mode svgs that is causing the problem.
I have run the WAVE tool on the changes and found no contrast errors.

PR #1543, opened 2021-04-19
Fix the ordering of Blogs

### LINKED ISSUE #1534: Fix the ordering of blogs
There is a [FIXME](https://github.com/ocaml/ocaml.org/blob/master/script/rss2html.ml#L191) in the code ordering blogs by name concerning the following: a leading white space in the name of a blog will impact the ordering. So ideally leading white spaces should be removed before ordering, but arent' at the moment. I propose fixing that.

shreyaswikriti:
# Issue Description

Fixed the ordering of blogs ([FIXME](https://github.com/ocaml/ocaml.org/blob/master/script/rss2html.ml#L191))by removing the leading whitespace using String.trim function. Now leading whitespace won't impact ordering.

Fixes #1534

## Changes Made

### DIFF
--- a/script/rss2html.ml
+++ b/script/rss2html.ml
@@ -188,9 +188,8 @@ let get_opml () =
                                "OCaml Planet" in
   (* Broken feeds will be marked with [is_comment = true]. *)
   let opml = Opml1.of_atom ~head feeds in
-  (* Sort by name.  (FIXME: one may want to ignore spaces.) *)
-  let trimm str = String.trim str in
-  let by_name o1 o2 = String.compare (trimm o1.Opml1.text) (trimm o2.Opml1.text) in
+  (* Sort by name. *)
+  let by_name o1 o2 = String.compare (String.trim o1.Opml1.text) (String.trim o2.Opml1.text) in
   { opml with Opml1.body = List.sort by_name opml.Opml1.body }
 
 let opml fname =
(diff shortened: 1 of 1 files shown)


shreyaswikriti:
@pitag-ha, Can you please look into this?

script/rss2html.ml
```diff
@@ -189,7 +189,8 @@ let get_opml () =
   (* Broken feeds will be marked with [is_comment = true]. *)
   let opml = Opml1.of_atom ~head feeds in
   (* Sort by name.  (FIXME: one may want to ignore spaces.) *)
```
gs0510: You can remove the FIXME from this comment :)

script/rss2html.ml
```diff
@@ -189,7 +189,8 @@ let get_opml () =
...
-  let by_name o1 o2 = String.compare o1.Opml1.text o2.Opml1.text in
+  let trimm str = String.trim str in
```
gs0510: You can name this function trim :) Alternatively you don't need to define this and instead can use `String.trim` directly in line 193. :)

gs0510 (changes requested):
Thanks for your PR @shreyaswikriti! Looks good, some tiny changes needed :)

shreyaswikriti:
@gs0510, Thanks for the review, I'll push the suggested changes.

avsm:
This looks good, thank you for the contribution!

PR #1546, opened 2021-04-20
Reduced duplicated names in contributor list

### LINKED ISSUE #1421: Removal of duplicate names in contributors list
There are some persons in the contributors list whose names appear twice. Normally in the form of their Full name on one side and their github usernames elsewhere in the list. To show a more accurate list of contributors, the duplicates should be tracked down and removed.

lindaokorie27:
# Issue Description

The names in the contributors list are duplicated making the list very long

Fixes #1421 

## Changes Made

I checked the list using email addresses to make the search term more unique than using names

### DIFF
--- a/site/contributors.md
+++ b/site/contributors.md
@@ -44,7 +44,7 @@ The contributors to this site, extracted from the
 [Git log](https://github.com/ocaml/ocaml.org/commits/master), are:
 
 <ul id="contributors_list">
-((! cmd git log --format="%aE %aN" | sort | uniq | awk '{print "<li>"$2,"</li>"}' !))
+((! cmd git log --format="%aE %aN" | sort -u -k 1,1 | awk '{print "<li>"$2,$3,$4"</li>"}' !))
 </ul>
 
 
--- a/site/contributors.md
+++ b/site/contributors.md
@@ -44,7 +44,7 @@ The contributors to this site, extracted from the
 [Git log](https://github.com/ocaml/ocaml.org/commits/master), are:
 
 <ul id="contributors_list">
-((! cmd git log --format="%aN" | sort | uniq | awk '{print "<li>"$1,$2,$3"</li>"}' !))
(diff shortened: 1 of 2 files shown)


site/contributors.md
```diff
@@ -44,7 +44,7 @@ The contributors to this site, extracted from the
 [Git log](https://github.com/ocaml/ocaml.org/commits/master), are:
 
 <ul id="contributors_list">
-((! cmd git log --format="%aN" | sort | uniq | awk '{print "<li>"$1,$2,$3"</li>"}' !))
+((! cmd git log --format="%aE %aN" | sort | uniq | awk '{print "<li>"$2,"</li>"}' !))
```
gs0510: This results in printing only the first name of folks which isn't what we want. You want to list only the contributors using their email as a key for finding unique contributors.

gs0510 (commented):
Thanks for your PR @LindieK! Were you able to check if the changes you made resulted in less duplications on the contributors.html page? Can you add a screenshot? Thanks!

site/contributors.md
lindaokorie27: Oh sorry. Let me try again.

lindaokorie27: @gs0510 I have worked on it now. It displays contributor's full names. It removes names attached to the same email address.  There are a lot of contributors though that have multiple email addresses.

gs0510: @LindieK you can look into how to find only the uniq names in the list so awk can print only them. :)

lindaokorie27: Thanks for idea. I'll look into it right away

lindaokorie27: I sorted it further after the awk statement. The only thing left causing duplications is that the email and name documented on some commits are the person's github username (different from the full name used in other commit messages) and the auto-generated email from github that has the `@users.noreply.gtihub.com` extension.

gs0510 (commented):
Awesome, thanks! LGTM :tada:

Issue #1503, opened 2021-04-10
link to stackshare from industrial users


ghost:
In addition to the locally maintained lists, stackshare has a list of companies using ocaml

https://stackshare.io/ocaml

Consider linking to that website from the top or bottom of the industrial users page.

The API might also expose the data directly. If that is the case, one could get the names and urls for the companies for now, and create a follow up issue to generate a page with that data.

Bhavika-Gianey:
I would like to work on this.

Bhavika-Gianey:
@patricoferris  @pitag-ha, is this a medium-level issue? Shall I start working on this ??

pitag-ha:
@Bhavika-Gianey , I haven't had a look at the stackshare API yet, but you can have a look and see if the following is possible

> The API might also expose the data directly. If that is the case, one could get the names and urls for the companies

I think that would be a medium issue. Let us know :)

Bhavika-Gianey:
> @Bhavika-Gianey , I haven't had a look at the stackshare API yet, but you can have a look and see if the following is possible

Hii @pitag-ha , stackshre has a GraphQL API, I have never used that before but I will try to implement it and let you know in case of any difficulties :).

pitag-ha:
Perfect @Bhavika-Gianey! :) Do you know if using the stackshare API is for free?

Bhavika-Gianey:
> Perfect @Bhavika-Gianey! :) Do you know if using the stackshare API is for free?

I was going to discuss this with you @pitag-ha. It's free for the first 2 months allowing 100 API calls. And after that, there are monthly plans.

Bhavika-Gianey:
What do you think ?? Shall we proceed using this?

pitag-ha:
Thanks for investigating that, @Bhavika-Gianey! I think that should be ok, yes :) I don't think we would pay the monthly plan, but I'm hoping that 100 API calls will be enough to get the current list of names and urls of companies using OCaml (we simply wouldn't keep it updated).

ghost:
I spent a few minutes trying to find this data in the API and I didn't find it immediately. It may or may not exist there, so I would investigate that first and then come up with an implementation approach after finishing that research.

Bhavika-Gianey:
> I spent a few minutes trying to find this data in the API and I didn't find it immediately. It may or may not exist there, so I would investigate that first and then come up with an implementation approach after finishing that research.

[...]

PR #1572, opened 2021-04-28
Create script to fetch company data from stackshare api

Bhavika-Gianey:
# Issue Description
In addition to the locally maintained lists, stackshare has a list of companies using OCaml
https://stackshare.io/ocaml
The issue is to generate a JSON file from the stackshare API.

Work in progress  #1503 

## Changes Made
I tried to fetch data from the API but it is showing this error:

pp

### DIFF
--- a/script/test.ml
+++ b/script/test.ml
@@ -1,0 +1,43 @@
+#use "topfind";;
+#thread;;
+#require "lwt";;
+#require "cohttp-lwt-unix";;
+#require "ezjsonm";;
+
+open Lwt
+open Cohttp
+open Cohttp_lwt_unix
+open Printf
+
+let uri = "https://api.stackshare.io/graphql"
+let key = "hmTBPOT6qizPthv9BVFjFg"
+
+let query = {| query {
+leads(usingToolSlugs: ["ocaml"], toolMatch: "any",
+after:""){
+edges{
+node{
+companyId
+companyName
+domain
+}
+}
+}
+}|}
+
+let send_graphql_query query =
+let open Lwt.Infix in
+let headers =
+Cohttp.Header.of_list ["Content-type", "application/json"; "Accept", "application/json"; "x-api-key" , key]
+in
+let uri = Uri.of_string uri in
+let body =
+`O [("query", `String query)]
+in
+let serialized_body = Ezjsonm.to_string body in
+Cohttp_lwt_unix.Client.post ~headers ~body:(`String serialized_body) uri >>= fun (_resp, body) ->
+Cohttp_lwt.Body.to_string body >|= fun s ->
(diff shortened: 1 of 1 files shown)


script/test.ml
```diff
@@ -0,0 +1,43 @@
...
+}|}
+
+let send_graphql_query query =
+let open Lwt.Infix in
+let headers =
+Cohttp.Header.of_list ["Content-type", "application/json"; "Accept", "application/json"; "x-api-key" , key]
+in
+let uri = Uri.of_string uri in
+let body =
+`O [("query", `String query)]
```
Bhavika-Gianey, line 35: what is this piece of code doing?
and Why are we using `Ezjsonm.to_string`  body if the above expression is already converting the query to string form.??

Bhavika-Gianey:
Hii @patricoferris, can you please help me resolve this.

script/test.ml
gs0510: In this the json object is being constructed, so `body` is still [an ezyjsonm type](https://github.com/mirage/ezjsonm/blob/master/lib/ezjsonm.mli#L42) which is why you still need to convert body to a string type for sending the post request in line 38.

Also, if you do an `opam install ocamlformat` and run ` dune build @fmt --auto-promote .` it will format the code so it's more legible. :)

patricoferris:
As @gs0510 quite rightly mentioned `Ezjsonm.to_string` in converting our value to a **JSON** string, this can be transferred over the internet to the API. 

I suggest having a play around with it in `utop` for example: 

```ocaml
utop # Ezjsonm.to_string (`O ["hello", `String "world"]);;
- : string = "{\"hello\":\"world\"}"
```

Most GraphQL APIs (afaik) understand the body as a JSON object where they query is stored as a value associated with a key called `query` (there are other things you can add to like `variables` etc.) so that's what we're doing in that part of the code: 

```json
{
  "query": "query { ... }"
}
```

Also as a general rule it is pretty bad to commit API keys into any code. Other people could abuse it but the blame may come back to you since it is your API key. I don't think anyone here would do that of course, but just letting you know :)) You probably want to disable that API key and generate a new one.

Bhavika-Gianey:
Thanks, @gs0510, @patricoferris, I understood the use of  `Ezjsonm.to_string` and I will generate a new API Key. 
But I am still getting the error of no query string present.

pp. :((.

I think the error is caused due to the presence of [arguments inside the query](https://github.com/ocaml/ocaml.org/pull/1572/files#diff-ed977902fd4d7cf3852edc590ed0a8d33eb204ab0359703f813c7d801f1bf4e1R16) but I  don't know how to resolve that. Plse kindly look into this and help me.

PR #1574, opened 2021-04-30
html manual links changed from inria website to ocaml.org

### LINKED ISSUE #1131: html manuals still link to inria website instead of ocaml.org
in the `releases` directory, the `*refman.html`, `*.refman-html.zip`, and `refman-html.tar.gz` files point to the INRIA
website in the sentence

"This manual is also available in PDF. plain text, as a bundle of HTML files, and as a bundle of Emacs Info files."

instead of pointing to ocaml.org, while the links are correct in https://ocaml.org/releases/4.10/htmlman/index.html

This is related to https://github.com/ocaml/ocaml.org/pull/1110

gurleennsidhuu:
# Issue Description

in the `releases` directory, the `*refman.html`, `*.refman-html.zip`, and `refman-html.tar.gz` files point to the INRIA
website in the sentence

"This manual is also available in PDF. plain text, as a bundle of HTML files, and as a bundle of Emacs Info files."

instead of pointing to ocaml.org.

Fixes #1131 

Changed all the links in the above listed files from inria website to ocaml.org.


gurleennsidhuu:
I hope I understood the issue correctly. If not, I would love to make the required changes.

avsm:
This looks good to me. @Octachron are you ok with the historical manual links being changed?

Octachron (changes requested):
I am perfectly ok with updating html files.

There is however an issue with the *.tar.gz file: their size is increasing with a factor between 5 and 13.

Which brings me to the issue of reproducibility: it seems to me that it would be more maintainable (and safer) if we had a script to recreate the compressed archive from the html source.

PR #1576, opened 2021-05-01
Replacement "Data Types and Matching" tutorial

johnwhitington:
This PR replaces the "Data types and matching" tutorial.

The nice symbolic algebra example from the original tutorial is retained in modified form, but the rest is new.

We may need to rebase this after the new lists tutorial is merged, and we may need to think about changing the file name now that the title has changed (perhaps it should not be changed?). But the content may be reviewed now.

### DIFF
--- a/site/learn/tutorials/data_types_and_matching.md
+++ b/site/learn/tutorials/data_types_and_matching.md
@@ -112,16 +112,80 @@ let example = function
 
 ## Constructors with data
 
+We can do more than just enumerations. 
+
+```ocamltop
+type colour2 =
+  | Red
+  | Green
+  | Blue
+  | Yellow
+  | RGB of float * float * float
+  | Mix of float * colour2 * colour2
+```
+
 difference between a * b and (a * b)
 
-records in our own data type
+recursive:
 
-incomplete pattern matching
+```ocamltop
+type colour2 =
+  | Red
+  | Green
+  | Blue
+  | Yellow
+  | RGB of float * float * float
+  | Mix of float * colour2 * colour2
+```
 
+incomplete pattern matching example - what function?
 _ in pattern matching - reminder of when appropriate
 
+function to get rgb from any
+function to get any from rgb (yellow?)
+
+records in our own data type (put labels on existing data type, effectively *)
 
 ## Example: trees
+
+Data types may be polymorphic and recursive too. Here is an OCaml data type for
+a binary tree carrying any kind of data:
+
+```ocamltop
+type 'a tree =
+  | Lf
+  | Br of 'a tree * 'a * 'a tree;;
+
+let t = Br (Br (Lf, 1, Lf), 2, Br (Br (Lf, 3, Lf), 4, Lf));;
+```
+
+A `Lf` leaf holds no information, just like an empty list. A `Br` branch holds
+a left tree, a value of type `'a` and a right tree. Now we can write recursive
+and polymorphic functions over these trees, by pattern matching on our new
+constructors:
+
+```ocamltop
+let rec total t =
+  match t with
+  | Lf -> 0
+  | Br (l, x, r) -> total l + x + total r;;
+
+let rec flip t =
+  match t with
+  | Lf -> Lf
+  | Br (l, x, r) -> Br (flip r, x, flip l);;
+```
+
+Let's try our new functions out:
+
+```ocamltop
+let all = total t;;
+
+let flipped = flip t;;
+
+t = flip flipped;;
+```
+
 
 Fixed types vs paramerarized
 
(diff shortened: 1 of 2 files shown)


site/learn/tutorials/data_types_and_matching.md
```diff
@@ -1,289 +1,268 @@
...
+let rec rgb_of_colour = function
+  | Red -> (1.0, 0.0, 0.0)
+  | Green -> (0.0, 1.0, 0.0)
+  | Blue -> (0.0, 0.0, 1.0)
+  | Yellow -> (1.0, 1.0, 0.0)
+  | RGB (r, g, b) -> (r, g, b)
+  | Mix (p, a, b) ->
+      let r1, g1, b1 = rgb_of_colour a in
+      let r2, g2, b2 = rgb_of_colour b in
+        (r1 +. r2 /. 2.0, g1 +. g2 /. 2.0, b1 +. b2 /. 2.0)
```
xavierleroy: Wrong mixing formula!  You seem to assume `p = 0.5` and you forgot to parenthesize `(r1 +. r2) /. 2.`...

Simple solution: omit the `p` parameter to `Mix`, defaulting to 50-50 mixes, and fix the parentheses.

Alternative:
```suggestion
  | Mix (p, a, b) ->
      let r1, g1, b1 = rgb_of_colour a in
      let r2, g2, b2 = rgb_of_colour b in
      let mix x y = x *. p +. y *. (1.0 -. p) in
        (mix r1 r2, mix g1 g2, mix b1 b2)
```

site/learn/tutorials/data_types_and_matching.md
```diff
@@ -1,289 +1,268 @@
...
+for a binary tree carrying any kind of data:
 
 ```ocamltop
-type 'a binary_tree =
-  | Leaf of 'a
-  | Tree of 'a binary_tree * 'a binary_tree;;
+type 'a tree =
+  | Lf
+  | Br of 'a tree * 'a * 'a tree;;
+
```
xavierleroy: I prefer the "leafs and nodes" terminology, because in plants branches start at nodes.  Also, why abbreviate?  So I'd suggest `Leaf` and `Node`.

site/learn/tutorials/data_types_and_matching.md
```diff
@@ -339,33 +309,33 @@ let rec multiply_out e =
...
-expressions of the form `e1 * (e2 + e3)`. Now look at the right hand
-side of this first pattern match, and convince yourself that it is the
-equivalent of `(e1 * e2) + (e1 * e3)`.
-
-The second pattern does the same thing, except for expressions of the
-form `(e1 + e2) * e3`.
+expressions of the form `e1 * (e2 + e3)`. Now look at the right hand side of
+this first pattern match, and convince yourself that it is the equivalent of
+`(e1 * e2) + (e1 * e3)`. The second pattern does the same thing, except for
+expressions of the form `(e1 + e2) * e3`.
```
xavierleroy, line 328: The problem with this style of simplifications (match at top then recurse) is that it's often incomplete.  Here, `e1 * (e2 * (e3 + e4))` becomes `e1 * (e2 * e3 + e2 * e4)` instead of the expected `e1 * e2 * e3 + e1 * e2 * e4`.

xavierleroy (commented):
Nice work, thank you!  Minor comments below.

site/learn/tutorials/data_types_and_matching.md
xavierleroy: Also: I'd prefer parentheses around the patterns `(r1, g1, b1)` and `(r2, g2, b2)` for consistency with the parentheses around the expression `(mix r1 r2, ...)`

johnwhitington: Whoops. Committed your suggestion. Will add the parens in a moment.

site/learn/tutorials/data_types_and_matching.md
johnwhitington: This is content lightly edited from the old tutorial - I didn't write it, and I haven't taken the time to understand the implementation.

Is it so defective that it needs cutting?

site/learn/tutorials/data_types_and_matching.md
johnwhitington: Abbreviated just out of habit. It makes typing example trees faster in the top level.

Have changed to Leaf and Node, in this tutorial and in "A first hour with OCaml", which also uses trees.

site/learn/tutorials/data_types_and_matching.md
xavierleroy: No, it's not that bad.  But users who really want to do algebraic simplifications should not follow this approach...  Yet, the better approach (bottom-up rewriting via smart constructors) generally doesn't involve deep pattern matching, so it's not a good example for what you want to demonstrate here!  Let's keep it unchanged, then.

avsm:
Thanks! Read through it and no material comments, and Xavier's review was addressed in 7961c2ab08c63257ff689ee2fa7615f6b9c89cb0

PR #1579, opened 2021-05-02
Improved Books section

### LINKED ISSUE #1403: Organizing books.html page
The https://ocaml.org/learn/books.html page looks a little disorganized. For eg:

Here the sections for 2 different books overlap but they should look like below.

Also, the links to the books/pdfs are not on the same position. For eg:
In some books:

and in other books:

I would like to fix this if it is approved. :)

Riddhima23:
# Issue Description

The Books section was disorganised with irregularities in alignment of text with pictures and spacing between rows.

Before:

https://user-images.githubusercontent.com/58878761/116821644-525ae780-ab98-11eb-98f1-81a5e89b8d35.mp4

Fixes #1403 

## Changes Made

Aligned the text and images and fixed the spacing issue between two books.

After:

1-a
3-a
5-a
6-a
7-a

https://user-images.githubusercontent.com/58878761/116821648-57b83200-ab98-11eb-9fcd-c52ef17c1146.mp4

### DIFF
--- a/site/learn/books.md
+++ b/site/learn/books.md
@@ -5,7 +5,7 @@
 # Books
 ## Books in English
 ###  The OCaml System: Documentation and User's Manual
-<img src="/img/colour-icon-170x148.png" style="float: left; margin-right: 15px; margin-bottom: 15px;"></img>
+<img src="/img/colour-icon-170x148.png" style="float: left; margin-right: 15px; margin-bottom: 40px; margin-top: 18px"></img>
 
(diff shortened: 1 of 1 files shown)


patricoferris:
Hi @Riddhima23 this looks good thanks. I'm wondering if some of the styling can't be extracted into a class at this point to avoid all of the duplication in the `style` attributes?

Riddhima23:
Hi @patricoferris! Sincere apologies for replying late! Could you please elaborate a bit on the question, do you want there should be some  ``` div ``` class wherein style attributes are given and I should just include that class in every image tag?

PR #1580, opened 2021-05-04
deleted dead scripts from script/ and opam.ocaml.org/

### LINKED ISSUE #1531: Spring clean up: detect and delete dead scripts

gurleennsidhuu:
# Issue Description

There were some dead code which had no effect on the build process.
Files deleted:
opam.ocaml.org/contact.xhtml: This file was not linked in any other file. The file contained code for a list containing items such as feedback, find us on github, etc. which I assume was supposed to be a part the footer of the site. The current footer list contains some different items.
opam.ocaml.org/home.xhtml: This file was not linked in any other file too. This page had not effect on the build.
opam.ocaml.org/link_blog_doc_ext.sh
opam.ocaml.org/packages: It was a blank file.
opam.ocaml.org/universe.xhtml: This file was not linked in any other file too. This file contains code for a table which only contains header information but not data.
scripts/meetings.ml: This file was not linked in any other file and had no effect on the build.
scripts/ocamlapplet.bash
Fixes #1531


patricoferris:
A cursory glance suggests `meetings.ml` and the `ocamlapplet.bash` could go. I don't think we can remove anything in `opam.ocaml.org` because I think the deployment of that site lives elsewhere but may (or may not) use all of those files.

PR #1587, opened 2021-05-20
Fix liquidSoap link

xvw:
# Issue Description

When I reread the success stories I realised that the liquid-soap link was not targeting the right link. I didn't "make an issue" because the change seemed... trivial.


avsm:
Thanks!

PR #1590, opened 2021-06-24
99problems update (number 12)

crap0101:
# Issue Description

Fixes # (issue)

## Changes Made

### DIFF
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -274,16 +274,13 @@ problem, construct its uncompressed version.
 SOLUTION
 
 > ```ocamltop
-> let decode list =
->   let rec many acc n x =
->     if n = 0 then acc else many (x :: acc) (n - 1) x
->   in
->   let rec aux acc = function
->     | [] -> acc
->     | One x :: t -> aux (x :: acc) t
->     | Many (n, x) :: t -> aux (many acc n x) t
->   in
->     aux [] (List.rev list);;
+> let rec decode = function
+>     | [] -> []
+>     | h::t -> match h with
+>         | One x -> x :: (decode t)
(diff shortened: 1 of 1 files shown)


patricoferris:
Hi, thank you for the PR. 

I think your solution can handle less input than the current solution, this is because your proposed solution is not [_tail-recursive_](https://www.cs.cornell.edu/courses/cs3110/2019sp/textbook/data/tail_recursion.html) which means we build up a series of function calls before doing any calculation. The `aux` function in the current solution is crafted to build the solution as we go. 

Concretely, your proposed solution would `stack overflow` with the following program (at least on my machine) whereas the current solution does not:

```ocaml
let () = 
  let lst = List.init 1_000_000 (fun _ -> One "a") in 
  let s = decode lst in 
  List.iter print_endline s
```

Do you agree? If so, can we close this PR ? Thanks again for taking the time to suggest a solution :))

crap0101:
> Hi, thank you for the PR.
> 
> I think your solution can handle less input than the current solution, this is because your proposed solution is not [_tail-recursive_](https://www.cs.cornell.edu/courses/cs3110/2019sp/textbook/data/tail_recursion.html) which means we build up a series of function calls before doing any calculation. The `aux` function in the current solution is crafted to build the solution as we go.
> 
> Concretely, your proposed solution would `stack overflow` with the following program (at least on my machine) whereas the current solution does not:
> 
> ```ocaml
> let () = 
>   let lst = List.init 1_000_000 (fun _ -> One "a") in 
>   let s = decode lst in 
>   List.iter print_endline s
> ```
> 
> Do you agree? If so, can we close this PR ? Thanks again for taking the time to suggest a solution :))

sure! sorry

PR #1591, opened 2021-06-24
update 99problems.md (number 13)

crap0101:
# Issue Description

Fixes # (issue)

## Changes Made

### DIFF
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -300,15 +300,17 @@ duplicates, as in problem "[Pack consecutive duplicates of list elements into su
 SOLUTION
 
 > ```ocamltop
-> let encode list =
->   let rle count x = if count = 0 then One x else Many (count + 1, x) in
->   let rec aux count acc = function
->     | [] -> [] (* Can only be reached if original list is empty *)
->     | [x] -> rle count x :: acc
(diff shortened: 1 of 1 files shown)


patricoferris:
Thanks, this solution suffers from the same problems I outlined in #1590 i.e. it is not _tail-recursive_ so it "stack overflows" on large inputs, I tried for example:

```ocaml
let () = 
  let lst = List.init 1_000_000 (fun i -> if i mod 2 = 0 then "b" else "a") in 
  let s = encode lst in 
  List.iter (function One a -> print_endline a | Many (i, a) -> Printf.printf "%s %i\n" a i) s
```

If you agree can we close this PR?

crap0101:
> Thanks, this solution suffers from the same problems I outlined in #1590 i.e. it is not _tail-recursive_ so it "stack overflows" on large inputs, I tried for example:
> 
> ```ocaml
> let () = 
>   let lst = List.init 1_000_000 (fun i -> if i mod 2 = 0 then "b" else "a") in 
>   let s = encode lst in 
>   List.iter (function One a -> print_endline a | Many (i, a) -> Printf.printf "%s %i\n" a i) s
> ```
> 
> If you agree can we close this PR?

sure! sorry (again)

Issue #22, opened 2012-08-17
news.html: fix formatting


agarwal:
This page is linked to from the "More news" button on the front page. The formatting is completely messed up.

Chris00:
Fixed in commit fa9697c586087db8f30a281195cba03270e10aa6

PR #1592, opened 2021-06-29
99problems.md add solution for #22

crap0101:
# Issue Description

Fixes # (issue)

## Changes Made

alternative solution for  number 22


crap0101:
...a variant which do not stack overflow :-D

`let range start stop =
>    let op = if stop > start then (-) else (+) in
>    let rec aux s acc = function
>        | 0 -> s :: acc
>        | n -> aux (op s 1) (s  :: acc) (n - 1) in
>       aux stop [] (abs (start - stop));;
`

Issue #12, opened 2012-08-17
books.html: import pdfs directly to server


agarwal:
Write now links to books point to external URLs. It would be better to organize all PDFs into a location on the web server, and provide internal links.

Chris00:
Well, we will not be adding the books to the repository!  So what does this mean as long as we do not have a server...?
Also, one should be careful to ask the authors whether they agree we do that.

agarwal:
Right, this is a reminder to do it when we have a server.

agarwal:
Linking to external content is now widespread throughout ocaml.org. We have videos on youtube, cheat sheets on ocamlpro.com, etc. and so this request now seems outdated. It was motivated by a desire to have everything self contained, which is not feasible (a good sign that the content on the site has grown a lot). Thus, closing this issue as a _will-not-do_.

dsheets:
We might want to implement a broken link checker at some point in the future.

avsm:
We might even want to implement a working link checker at some point too.

_shows himself to the door_

PR #1593, opened 2021-07-05
Update solution for #12 to be more succinct

anukem:
# Issue Description

Updating the solution for #12 as either an alternative or replacement.
I had tried the other solution and it seemed more difficult to reason about because of the additional helper function, but would love to get thoughts on whether or not this implementation is useful to add.

Fixes # (issue)

## Changes Made

### DIFF
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -274,16 +274,13 @@ problem, construct its uncompressed version.
 SOLUTION
 
 > ```ocamltop
-> let decode list =
->   let rec many acc n x =
->     if n = 0 then acc else many (x :: acc) (n - 1) x
->   in
->   let rec aux acc = function
->     | [] -> acc
->     | One x :: t -> aux (x :: acc) t
->     | Many (n, x) :: t -> aux (many acc n x) t
->   in
->     aux [] (List.rev list);;
+> let decode l =
+>  let rec aux acc = function
+>    | [] -> acc
+>    | (One h)::t -> aux (h::acc) t
+>    | Many (c, v)::t -> if c > 0 then aux (v::acc) ((Many (c - 1, v))::t) else aux acc t
+>  in List.rev (aux [] l)
+
 > ```
 
 ```ocamltop


patricoferris:
Hi @anukem apologies for the delay. 

I had a look at your solution and it seems correct but there are two problems I see.

1. I'm not sure removing the helper function has made it much clearer (although it is quite subjective). The reason I say that is the recursion in your `| Many... ->` step is now performing both the adding of the "many" letters *and* (once `c = 0`) the moving along of the list of `'a rle`s. For me this is slightly more confusing. A compromise perhaps is to annotate the functions in the original solution with comments describing what they do?
2. I believe your solution does a fair amount of extra allocating. OCaml is a garbage-collected language and for most non-integer values they are stored in the heap (allocated). This includes things like `Many (1, "a")` and lists. In the original solution the `many` helper function adds the appropriate amount of `'a` to the list by recursively calling itself with smaller `n` values. Your solution allocates a new `Many (n - 1, v)` on to a list (see https://ocaml.org/manual/intfc.html#s:c-ocaml-datatype-repr for more details). A result of this is that your solution for bigger lists with large `Many`s ends up using quite a bit more memory and takes a bit longer I think.

I'm no expert so perhaps I've gotten point (2) wrong, but I ran some benchmarks using [bechamel](https://github.com/mirage/bechamel) which seem to back-up my claims  https://gist.github.com/patricoferris/951ad129daf14e285b545017d67a06ef. To reproduce you should be able to do (I attached some results from a particular run too):

```
opam install notty bechamel bechamel-notty
dune exec -- ./main.exe
```

anukem:
hey @patricoferris, just now seeing this as well. I appreciate the perf concerns and certainly didn't know the goal was the most efficient solution so apologies for that. As for your first comment about readability, I think my goal in submitting was to provide an alternative solution didn't require a helper. Happy to meet in the middle with the comments on top for further descriptions. Also, I had a smaller change that still hasn't been reviewed that I would love someone to take a look at #1594

PR #1595, opened 2021-07-10
Fix wrong description about currying

masmatsum:
# Issue Description

Fix wrong description about currying

## Changes Made

- Fix word misuse (currying -> partial application)
- Omit description about currying

### 1. These examples are about partial application, not currying.

They are totally different concepts.
Current description is wrong and misunderstands these concepts.

#### Rough explanation for functional-programming newbies

Suppose there are functions like these (the example is JavaScript):

```javascript

function plus(a, b) {
  return a + b;
}

function curried_plus(a, b) {
  return function(b) {
    return a + b;
  };
}

function plus_one(y) {
  return 1 + y;
}

// Usages
plus(5, 7) //=> 12
curried_plus(5)(7) //=> 12
plus_one(3) //=> 4
```

In the example above,
- "currying" is converting `plus` to `curried_plus`.
- "partial application" is converting `plus` to `plus_one` (which means, fixing the value of `plus`'s nth argument).

### 2. Usually we do not need manual currying

In OCaml, all functions are automatically curried when defined. 

For example, if you define a function whose arity is more than one:

```ocaml
let plus a b = a + b
```

then OCaml compiler automatically curries it. 
The function above is equivalent to:

```ocaml
let plus = fun a -> fun b -> a + b
```

Effectively, all OCaml functions are arity-1.
In normal use case, we do not need manual currying.

### 3. It is hard to explain (real) currying with OCaml code

Due to automatic conversion which is described above, it is a little bit difficult to explain (real) currying with OCaml code.

So, if you want to `avoid entering into the mathematics behind OCaml`, then you should just omit description about currying.
In any case, we should not have wrong description in **official tutorial**.

---

### DIFF
--- a/site/learn/tutorials/functional_programming.md
+++ b/site/learn/tutorials/functional_programming.md
@@ -103,7 +103,7 @@ that it's trying to save.
 Now look at the definition of `receiver_fn`. This function is a closure
 alright because it keeps a reference to `chan` from its environment.
 
-## Partial function applications and currying
+## Partial function applications
 Let's define a plus function which just adds two integers:
 
 ```ocamltop
@@ -168,14 +168,7 @@ rationale for the strange `->` arrow notation used for function types:
   plus 2 : int -> int
 plus 2 3 : int
 ```
-This process is called **currying** (or perhaps it's called
-**uncurrying**, I never was really sure which was which). It is called
-this after Haskell Curry who did some important stuff related to the
-lambda calculus. Since I'm trying to avoid entering into the mathematics
-behind OCaml because that makes for a very boring and irrelevant
-tutorial, I won't go any further on the subject. You can find much more
-information about currying if it interests you by [doing a search on
-Google](http://www.google.com/search?q=currying "http://www.google.com/search?q=currying").
+This process is called **partial application** of a function.
 
 Remember our `double` and `multiply` functions from earlier on?
 `multiply` was defined as this:
@@ -219,7 +212,7 @@ identical definition of the `plus` function as before:
 let plus = ( + );;
 plus 2 3;;
 ```
-Here's some more currying fun:
+Here's some more examples of partial application:
 
 ```ocamltop
 List.map (plus 2) [1; 2; 3];;


Octachron:
I am really not a fan of the tone and style of this tutorial but its use of `currying` is correct. (And it is not part of the reference manual).

Currying refers to the injection between `A x B -> C` and `A -> B -> C`.
Thus,
```ocaml
let plus a b = a + b
```
is the curryfied version of
```ocaml
let plus2 (a,b) = a + b
```
And indeed, the former can be constructed from the latter by using a `curry` function:
```ocaml
let curry f x y = f (x,y)
let plus = curry plus2
```
However, I agree that is probably a good idea to rewrite completely the rambling paragraph about `currying`.

masmatsum:
> Currying refers to the injection between `A x B -> C` and `A -> B -> C`.

I don't know what "injection" here means, but yes, converting `A * B -> C` to `A -> B -> C` is currying.

masmatsum:
Just for your information:

Note that the type `A * B` is a [tuple type](https://ocaml.org/manual/types.html#sss:typexpr-tuple).

And in @Octachron 's example,
```ocamltop
# let plus2 (a, b) = a + b;;
val plus2 : int * int -> int = <fun>
```

here, `int * int` is a tuple type which consists of 2 ints.
Threrefore, the `plus2` precisely is "a function which takes one tuple" (arity-1).

masmatsum:
> I am really not a fan of the tone and style of this tutorial but its use of `currying` is correct. 

Why correct??
The examples in the current tutorial are these:

```ocamltop
let plus a b =
  a + b
```

```ocamltop
let f = plus 2;;
f 10;;
f 15;;
f 99;;
```

and converting `plus` to `f` is partial application, not currying.
Current tutorial does not have examples of real currying.

Octachron:
The function `plus` is an example of currying, since it is perfectly possible to write SML-style functions (which generally uses tuple arguments) in OCaml.

masmatsum:
> The function `plus` is an example of currying, since it is perfectly possible to write SML-style functions (which generally uses tuple arguments) in OCaml.

Thanks for quick reply.
But, sorry, I can't figure out what you mean. Could you explain in detail?

The current tutorial page does not have examples of "function which takes one tuple", and therefore does not have any conversions from `A * B -> C` to `A -> B -> C`.

Octachron:
There is indeed not explicit conversion from `plus0 : int * int -> int`. But `plus: int -> int -> int` is the implicit curried version of this function.

Octachron:
I do agree that the rambling about currying is really at the wrong place, since the code above is demonstrating partial applications. Those partial application are made possible by the implictly curried definition of the function `plus`, but the role of currying is quite implicit at this point. And the non-explanation in the paragraph does not help.

masmatsum:
@Octachron  OK, I understood what you want to say.
Yes, the code itself is demonstrating partial applications.
So the sentences `This process is called currying` and `Here's some more currying fun:` in the page are very misleading.

masmatsum:
So, what should we do?
I think there are two options:

- a) Delete descriptions about currying
- b) Explain currying correctly, and give appropriate examples.

masmatsum:
>- b) Explain currying correctly, and give appropriate examples.

If we choose "b", we have to explain the following correctly:

- What is currying?
  - Definition
  - Comparison between functions "before currying" and "after currying"
- Implicit currying in OCaml
  - In OCaml, when you define a function (whose arity is greater than 1), the function is implicitly curried

Octachron:
In fact, rereading the text going with option 1 and deleting the mention of currying (like this PR is currently doing) might be for the best:
* the current description of currying is terrible
* the section is mostly about partial application
* I don't think that I can extend the current text while mimicking the current style.

masmatsum:
@Octachron
I agree with you. For this PR, I think it is better just deleting descriptions about currying.

If we explain currying correctly, we need to modify many sentences.
If we do this, it is better to create a subsequent PR.

masmatsum:
If there are no problems, could anyone merge this PR?

Octachron (approved):
I think that the change is good, and improve the current text. However, I dislike the style of this tutorial and a more neutral point of view would be useful.  Is there someone else willing to have a quick look at the change? Otherwise, I will merge the change sometimes next week.

patricoferris:
For what it is worth I think the changes are good and make it more readable. Also +1 on the overall tone of this tutorial being not as good as it could be in my opinion :))

masmatsum:
@Octachron It has already been 16 days since your last comment.
Could you merge this PR ?

The overall tone can be (and should be) fixed in a separate PR, so I think merging now is better.

Octachron:
Thanks for the reminder and the PR! This is merged now.

PR #1596, opened 2021-07-17
Rewrote set tutorial from scratch.

### LINKED ISSUE #598: Suggestion for the Set tutorial
If you search the web how to use Sets in OCaml you'll end up on the Set tutorial, which very clearly explain how to make a Set module for String-s like this:
module SS = Set.Make(String);;

Chances are the next thing the next thing you'll try is
module SI = Set.Make(int);;

Which won't work because int is not an OrderedType. It'd be nice to show how to make a set module for arbitrary non-ordered types. 

For example, by defining the corresponding OrderedType:

module Int = struct
  type t = int
  let compare = Pervasives.compare
end;;

NebuPookins:
# Issue Description

Fixes https://github.com/ocaml/ocaml.org/issues/598

## Changes Made

This version of the tutorial now demonstrates how to use Set with
arbitrary types. It also provides a demonstration on how to reason
about the behavior of functions based on their type signatures.

### DIFF
--- a/site/learn/tutorials/set.md
+++ b/site/learn/tutorials/set.md
@@ -3,74 +3,252 @@
 
 # Set
 
-## Module Set
-To make a set of strings:
+`Set` is a functor, which means that it is a module that is parameterized
+by another module. More concretely, this means you cannot directly create
+a set; instead, you must first specify what type of elements your set will
+contain.
+
+The `Set` functor provides a function `Make` which accepts a module as a
+parameter, and returns a new module representing a set whose elements have
+the type that you passed in. For example, if you want to work with sets of
+strings, you can invoke `Set.Make(String)` which will return you a new module
+which you can assign the name `SS` (short for "String Set"). Note: Be sure to
+pay attention to the case; you need to type `Set.Make(String)` and not
+`Set.Make(string)`. The reason behind this is explained in the
+"Technical Details" section at the bottom.
+
+Doing this in the OCaml's top level will yield a lot of output:
 
 ```ocamltop
 module SS = Set.Make(String);;
 ```
-To create a set you need to start somewhere so here is the empty set:
+
+What happened here is that after assigning your newly created module to the name
+`SS`, OCaml's top level then displayed the module, which in this case contains
+a large number of convenience functions for working with sets (for example `is_empty`
+for checking if you set is empty, `add` to add an element to your set, `remove` to
+remove an element from your set, and so on).
+
+Note also that this module defines two types: `type elt = String.t` representing
+the type of the elements, and `type t = Set.Make(String).t` representing the type of
+the set itself. It's important to note this, because these types are used in the
+signatures of many of the functions defined in this module.
+
+For example, the `add` function has the signature `elt -> t -> t`, which means
+that it expects an element (a String), and a set of strings, and will return to you
+a set of strings. As you gain more experience in OCaml and other function languages,
+the type signature of functions are often the most convenient form of documentation
+on how to use those functions.
+
+## Creating a Set
+
+You've created your module representing a set of strings, but now you actually want
+to create an instance of a set of strings. So how do we go about doing this? Well, you
+could search through the documentation for the original `Set` functor to try and
+find what function or value you should use to do this, but this is an excellent
+opportunity to practice reading the type signatures and inferring the answer from them.
+
+You want to create a new set (as opposed to modifying an existing set). So you should
+look for functions whose return result has type `t` (the type representing the set),
+and which *does not* require a parameter of type `t`.
+
+Skimming through the list of functions in the module, there's only a handful of functions
+that match that criteria: `empty: t`, `singleton : elt -> t`, `of_list : elt list -> t`
+and `of_seq : elt Seq.t -> t`.
+
+Perhaps you already know how to work with lists and sequences in OCaml or
+perhaps you don't. For now, let's assume you don't know, and so we'll focus
+our attention on the first two functions in that list: `empty` and `singleton`.
+
+The type signature for `empty` says that it simply returns `t`, i.e. an instance
+of our set, without requiring any parameters at all. By intuition, you might
+guess that the only reasonable set that a library function could return when
+given zero parameters is the empty set. And the fact that the function is named
+`empty` reinforces this theory.
+
+Is there a way to test this theory? Perhaps if we had a function which
+could print out the size of a set, then we could check if the set we get
+from `empty` has a size of zero. In other words, we want a function which
+receives a set as a parameter, and returns an integer as a result. Again,
+skimming through the list of functions in the module, we see there is a
+function which matches this signature: `cardinal : t -> int`. If you're
+not familiar with the word "cardinal", you can look it up on Wikipedia
+and notice that it basically refers to the size of sets, so this reinforces
+the idea that this is exactly the function we want.
+
+So let's test our hypothesis:
 
 ```ocamltop
 let s = SS.empty;;
+SS.cardinal s;;
 ```
-Alternatively if we know an element to start with we can create a set
-like
+
+Excellent, it looks like `SS.empty` does indeed create an empty set,
+and `SS.cardinal` does indeed print out the size of a set.
+
+What about that other function we saw, `singleton : elt -> t`? Again,
+using our intuition, if we provide the function with a single element,
+and the function returns a set, then probably the function will return
+a set containing that element (or else what else would it do with the
+parameter we gave it?). The name of the function is `singleton`, and
+again if you're unfamiliar with what word, you can look it up on
+Wikipedia and see that the word means "a set with exactly one element".
+It sounds like we're on the right track again. Let's test our theory.
 
 ```ocamltop
 let s = SS.singleton "hello";;
+SS.cardinal s;;
 ```
-To add some elements to the set we can do.
+
+It looks like we were right again!
+
+## Working with Sets
+
+Now let's say we want to build bigger and more complex sets. Specifically,
+let's say we want to add another element to our existing set. So we're
+looking for a function with two parameters: One of the parameters should
+be the element we wish to add, and the other parameter should be the set
+that we're adding to. For the return value, we would expect it to either
+return unit (if the function modifies the set in place), or it returns a
+new set representing the result of adding the new element. So we're
+looking for signatures that look something like `elt -> t -> unit` or
+`t -> elt -> unit` (since we don't know what order the two parameters
+should appear in), or `elt -> t -> t` or `t -> elt -> t`.
+
+Skimming through the list, we see 2 functions with matching signatures:
+`add : elt -> t -> t` and `remove : elt -> t -> t`. Based on their names,
+`add` is probably the function we're looking for. `remove` probably removes
+an element from a set, and using our intuition again, it does seem like
+the type signature makes sense: To remove an element from a set, you need
+to tell it what set you want to perform the removal on and what element
+you want to remove; and the return result will be the resulting set after
+the removal.
+
+Furthermore, because we see that these functions return `t` and not `unit`,
+we can infer that these functions do not modify the set in place, but
+instead return a new set. Again, we can test this theory:
 
 ```ocamltop
-let s =
-  List.fold_right SS.add ["hello"; "world"; "community"; "manager";
-                          "stuff"; "blue"; "green"] s;;
+let firstSet = SS.singleton "hello";;
+let secondSet = SS.add "world" firstSet;;
+SS.cardinal firstSet;;
+SS.cardinal secondSet;;
 ```
-Now if we are playing around with sets we will probably want to see what
-is in the set that we have created. To do this we can write a function
-that will print the set out.
+
+It looks like our theories were correct!
+
+## Sets of With Custom Comparators
+
+The `SS` module we created uses the built-in comparison function provided
+by the `String` module, which performs a case-sensitive comparison. We
+can test that with the following code:
 
 ```ocamltop
-(* Prints a new line "\n" after each string is printed *)
-let print_set s = 
-   SS.iter print_endline s;;
+let firstSet = SS.singleton "hello";;
+let secondSet = SS.add "HELLO" firstSet;;
+SS.cardinal firstSet;;
+SS.cardinal secondSet;;
 ```
-If we want to remove a specific element of a set there is a remove
-function. However if we want to remove several elements at once we could
-think of it as doing a 'filter'. Let's filter out all words that are
-longer than 5 characters.
 
-This can be written as:
+As we can see, the `secondSet` has a cardinality of 2, indicating that
+`"hello"` and `"HELLO"` are considered two distinct elements.
+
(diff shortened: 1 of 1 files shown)


site/learn/tutorials/set.md
```diff
@@ -3,74 +3,215 @@
...
+and `of_seq : elt Seq.t -> t`.
+
+Perhaps you already know how to work with lists and sequences in OCaml or
+perhaps you don't. For now, let's assume you don't know, and so we'll focus
+our attention on the first two functions in that list: `empty` and `singleton`.
+
+The type signature for `empty` says that it simply returns `t`, i.e. an instance
+of our set, without requiring any parameters at all. By intuition, you might
+guess that the only reasonable set that a library function could return when
+given zero parameters is the empty list. And the fact that the function is named
```
pmetzger: This should read `is the empty set`, not, `is the empty list`.

pmetzger (changes requested):
Generally, I like the things that you've added a lot, this is vastly clearer for a beginner and should be of great use to people starting out.

It might still be nice to include some of the information you've disposed of. For example, it would be nice to explain that Sets are pure functional (and that you can expect that from the type signatures), and a bit about what the provided operations are. Also it would be nice to explain the distinction between the `int` type and the `Int` module and why you can use the one but not the other.

site/learn/tutorials/set.md
NebuPookins: Good catch, will fix.

NebuPookins:
Thanks for the feedback. I think I've addressed them in this latest commit.

site/learn/tutorials/set.md
```diff
@@ -3,74 +3,252 @@
...
+The `Set` functor provides a function `Make` which accepts a module as a
+parameter, and returns a new module representing a set whose elements have
+the type that you passed in. For example, if you want to work with sets of
+strings, you can invoke `Set.Make(String)` which will return you a new module
+which you can assign the name `SS` (short for "String Set"). Note: Be sure to
+pay attention to the case; you need to type `Set.Make(String)` and not
+`Set.Make(string)`. The reason behind this is explained in the
+"Technical Details" section at the bottom.
+
+Doing this in the OCaml's top level will yield a lot of output:
```
pmetzger, line 20: If you're not displaying all the output, you may want to add ellipses here or some such.

pmetzger (changes requested):
One more small comment from me, but overall I think this looks good. It would be nice to get another opinion of course.

site/learn/tutorials/set.md
```diff
@@ -3,74 +3,252 @@
 
 # Set
 
-## Module Set
-To make a set of strings:
+`Set` is a functor, which means that it is a module that is parameterized
```
Octachron, line 6: This tutorial needs some introduction: it is reachable from the `Data structures` section of the tutorials. Thus when landing on the page the reader is probably expecting to learn about `Set` as a data structure. Starting with "`Set` is a functor" is thus doubly confusing: "What is a functor? Wasn't I promised a tutorial about a data structure?"

site/learn/tutorials/set.md
```diff
@@ -3,74 +3,252 @@
...
+given zero parameters is the empty set. And the fact that the function is named
+`empty` reinforces this theory.
+
+Is there a way to test this theory? Perhaps if we had a function which
+could print out the size of a set, then we could check if the set we get
+from `empty` has a size of zero. In other words, we want a function which
+receives a set as a parameter, and returns an integer as a result. Again,
+skimming through the list of functions in the module, we see there is a
+function which matches this signature: `cardinal : t -> int`. If you're
+not familiar with the word "cardinal", you can look it up on Wikipedia
```
Octachron, line 75: I don't think that sending the reader to a Wikipedia page is a good idea in general: either trust the reader to know the term or explain it. This particularly true for the mathematical section of Wikipedia which tends to be quite technical. Typically, the cardinal page of wikipedia cites the axiom of choice in its introduction.

site/learn/tutorials/set.md
```diff
@@ -3,74 +3,252 @@
...
+the set itself. It's important to note this, because these types are used in the
+signatures of many of the functions defined in this module.
+
+For example, the `add` function has the signature `elt -> t -> t`, which means
+that it expects an element (a String), and a set of strings, and will return to you
+a set of strings. As you gain more experience in OCaml and other function languages,
+the type signature of functions are often the most convenient form of documentation
+on how to use those functions.
+
+## Creating a Set
```
Octachron, line 43: The name of the section doesn't match the content. This section is describing how to analyze an unknown module, not how to create a set. I would propose to move this section later, and simply start by creating a set in the tutorial.

site/learn/tutorials/set.md
```diff
@@ -3,74 +3,252 @@
...
+remove an element from your set, and so on).
+
+Note also that this module defines two types: `type elt = String.t` representing
+the type of the elements, and `type t = Set.Make(String).t` representing the type of
+the set itself. It's important to note this, because these types are used in the
+signatures of many of the functions defined in this module.
+
+For example, the `add` function has the signature `elt -> t -> t`, which means
+that it expects an element (a String), and a set of strings, and will return to you
+a set of strings. As you gain more experience in OCaml and other function languages,
```
Octachron, line 39: function languages ⇒ functional languages

I am not that fond of meta comments that discourse about the expected proficiency of the reader.

site/learn/tutorials/set.md
```diff
@@ -3,74 +3,252 @@
...
+contain.
+
+The `Set` functor provides a function `Make` which accepts a module as a
+parameter, and returns a new module representing a set whose elements have
+the type that you passed in. For example, if you want to work with sets of
+strings, you can invoke `Set.Make(String)` which will return you a new module
+which you can assign the name `SS` (short for "String Set"). Note: Be sure to
+pay attention to the case; you need to type `Set.Make(String)` and not
+`Set.Make(string)`. The reason behind this is explained in the
+"Technical Details" section at the bottom.
```
Octachron, line 18: Delaying a simple explanation "`String` is a module" (?) to a technical section seems unnecessary?

site/learn/tutorials/set.md
```diff
@@ -3,74 +3,252 @@
...
 ```
-If we want to remove a specific element of a set there is a remove
-function. However if we want to remove several elements at once we could
-think of it as doing a 'filter'. Let's filter out all words that are
-longer than 5 characters.
 
-This can be written as:
+It looks like our theories were correct!
+
+## Sets of With Custom Comparators
```
Octachron, line 140: Sets with custom comparators?

Octachron:
After a quick glance, there are good ideas in the rewritten text. Nevertheless, I see two issues. First, the text is lacking an introduction. Second, there are two tutorials fighting each other currently: a "how to use a functor" tutorial and a set tutorial. Currently, the `Set` tutorial is losing in term of contents. The two tutorials need to be split.

site/learn/tutorials/set.md
NebuPookins: Not sure I understand this comment.

The original MD file just contains the one line you'd type into OCaml's top level:


       To make a set of strings:
       
       ```ocamltop
       module SS = Set.Make(String);;
       ```

But the output on https://ocaml.org/learn/tutorials/set.html contains not just the line you'd enter in, but also the output produced by OCaml's top level (I'm assuming whatever static site generator you're using is responsible for doing this). So my MD is doing something similar: I'm putting just the line that the user would enter in, and I'm expecting the static site generator to produce a huge amount of input.

The "Doing this in the OCaml's top level will yield a lot of output:" line I put there was intended to prepare new users to not panic in response to the huge block of text they're about to see (my reaction to seeing that initial huge block was panic).

NebuPookins:
Hi @Octachron 

I've skimmed through your various suggestions and I will attempt to address them in another revision of the pull request, but I'd like to address your "overarching feedback" here:

> There are two tutorials fighting each other currently: a "how to use a functor" tutorial and a set tutorial. Currently, the Set tutorial is losing in term of contents. The two tutorials need to be split.

This rewrite was done in response to https://github.com/ocaml/ocaml.org/issues/598 where someone pointed out that the original tutorial gave an example of creating a set of strings via `Set.Make(String)` and left it at that. They said that someone new to OCaml would probably then try something like `Set.Make(int)` which would fail and they would have no idea what when wrong.

I agree with their assessment, because as a new OCaml user, I was also surprised why one worked and the other didn't. This might be especially likely for people with a Java background or similar, where the type for String is called `String` (capital S) and the type for integers is called `int` (lowercase i).

After performing my own research on why `Set.Make(String)` worked but `Set.Make(int)` didn't, and then meditating on the best way to convey this to other new OCaml users, I think it is unavoidable to address the topic of types, modules, and functors.

This tutorial is primarily intended to be a tutorial on the using sets in OCaml with a target audience of people familiar with programming in general, but with zero OCaml experience (e.g. someone who has used set-like data structures in Java or other similar languages). I did not intend to focus specifically on how to use functors in the general case (to be honest, I don't know how to use functors in general in OCaml), but I do feel there is a minimum you need to know about functors in OCaml to actually use the Set module.

If there are more words spent on the topic of functors than on the topic of sets, it's probably because I expect most readers to know basically what a set is and how it works in other languages, but to be utterly confounded by this concept of "functor" which is likely to be totally alien to them. Especially since I don't expect the reader to have a good understanding of "modules" in the first place when they first stumble onto this page (I did not have a good understanding of modules when I first encountered this page).

Does that make sense?

Edit to add one more thought: One possibility is to tell the reader to first read the tutorials on modules and functors before reading this tutorial on sets, but I feel like that would be too discouraging to a new OCaml user. So I think it makes sense to do some minimal introduction to functors here, rather than ask the reader to learn about functors first.

Octachron:
I agree that the tutorial should not assume too much familiarity with functors. And that it is necessary to have good functor tutorial.  However, this is the `set` tutorial. The reader explicitly looked for explanations on sets in OCaml when they reached that page. You cannot decide in their stead that they are familiar with sets and just need to be told about functors. And your current explanation of functors is more on depth that the one in the functor tutorial. There is something amiss here. The set tutorial should focus on sets, maybe explain some details about to define set of custom types and link to the functor tutorial for further explanation. The detailed explanation on how to use a functor should go to the functor tutorial.

site/learn/tutorials/set.md
```diff
@@ -3,74 +3,252 @@
...
+Note also that this module defines two types: `type elt = String.t` representing
+the type of the elements, and `type t = Set.Make(String).t` representing the type of
+the set itself. It's important to note this, because these types are used in the
+signatures of many of the functions defined in this module.
+
+For example, the `add` function has the signature `elt -> t -> t`, which means
+that it expects an element (a String), and a set of strings, and will return to you
+a set of strings. As you gain more experience in OCaml and other function languages,
+the type signature of functions are often the most convenient form of documentation
+on how to use those functions.
```
Octachron, line 41: I skipped that part before, but no `types` are not the most convenient form of documentation. It is the most basic form of documentation.

NebuPookins:
>  The reader explicitly looked for explanations on sets in OCaml when they reached that page. You cannot decide in their stead that they are familiar with sets and just need to be told about functors. And your current explanation of functors is more on depth that the one in the functor tutorial. There is something amiss here. The set tutorial should focus on sets, maybe explain some details about to define set of custom types and link to the functor tutorial for further explanation. The detailed explanation on how to use a functor should go to the functor tutorial.

We might be at an impasse here. What's the conflict resolution strategy used for this repo if some people like the change and some people don't like it? Do we vote, or does "one dislike" veto the change, or what? I don't want to bother addressing the minor issues (e.g. the "function language" -> "functional language" typo) if the PR is likely to be rejected in the end.

In case it'll sway your vote, I claim that even if this tutorial isn't the ideal tutorial you have in mind, it better serves the goal of having a new OCaml programmer become comfortable with using Set than the previous tutorial, and so it'd be an overall positive if merged in. That's just my (probably biased) opinion, though.

Octachron:
I am not sure why you think that we are at an impasse? What is the issue with rerouting the functor related explanations to the functor tutorial?

site/learn/tutorials/set.md
pmetzger: Maybe the right thing then is to split this into two tutorials, one about sets-as-functors and another about how to use sets for their own sake?

pmetzger:
We could split this in two perhaps? Explain functors in one part, and the data structure itself in another, and advise readers to skip the functor section if and only if they're comfortable with functors?

pmetzger:
(BTW, this is a _very_ clear explanation of functors for the new user who thinks they're just interested in sets but needs to understand functors to make use of sets.)

NebuPookins:
> What is the issue with rerouting the functor related explanations to the functor tutorial?

I re-read the tutorial I wrote, and I actually don't think I spend very much time talking about functors at all. As far as I can see, this is the only explanation I give on functors:

> `Set` is a functor, which means that it is a module that is parameterized by another module.

Every other sentence in the tutorial is about Set.

The word "functor" appears a total of 4 times in the tutorial, including the sentence I quoted above which gives a one sentence definition of the term. The other 3 uses are always "the `Set` functor", meaning I'm talking about `Set`, but I thought just saying "the `Set`" would sound too awkward: For example, I think "The `Set` provides a function `Make` which..." doesn't sound right; `Set` is acting as an adjective here, and I need some noun. So I used "functor" as the noun, but the sentences are about Set, not about functors.

I could change it to "module", as in "The `Set` module provides a function `Make` which..." but then I think this would cause extra confusion to a beginner because I had previous said `Set` is a functor, and now I'm saying it's a module. It's true that functors are modules, but I don't want to distract the beginner with that detail, which is why I avoid bringing that up until we get to the "Technical Details" section.

So I guess my issue is that it's not clear to me what in the tutorial seems to be "about functors" that should be split off.

Octachron:
Hm, I think that I see what you mean. From my point of view, when you say

> you could search through the documentation for the original `Set` functor to try and
find what function or value you should use to do this, but this is an excellent
opportunity to practice reading the type signatures and inferring the answer from them.

you are discarding the set data structure and talking about how to make sense of the result of a generic functor.  Similarly, the sections `Creating a set` and `Working with a set` are not really about sets but how to construct a value from an abstract type. For instance, a section about creating a set should have created an empty set by the end of the first paragraph, but with the current focus on type-directed navigation of the signature, that only happens in the 7th paragraph. That makes sense for a module or functor tutorial, but that seems quite strange for a tutorial about sets.

Maybe another editing direction will be to start describing the basic set functions before diving in type exploration mode to discover more functions for creating or updating sets. I think that grounding a little more those sections in the data structure being described might remove my impression of reading a generic functor tutorial.

site/learn/tutorials/set.md
Octachron: The introduction should start by explaining what is a set from OCaml point of view : a collection of ordered elements. Then it would make sense to the reader that the `Set.Make` functor is building a set module from a module of ordered elements.

NebuPookins:
I think I better understand your point of view now. Let me think about this some more and I'll probably submit another revision.

PR #1597, opened 2021-07-25
Add a solution for "94. An arithmetic puzzle." in 99 Problems

dracconi:
## Changes Made

Added a solution to the 94th problem in 99 Problems.
Seems to work, but who knows :D

Done per request for commits "Please contribute more solutions or improve the existing ones."

### DIFF
--- a/site/learn/tutorials/99problems.md
+++ b/site/learn/tutorials/99problems.md
@@ -2785,7 +2785,87 @@ equations 2 - 3 + 5 + 7 = 11 or 2 = (3 * 5 + 7) / 11 (and ten others!).
 <!-- SOLUTION -->
 
 ```ocaml
-(* example pending *);;
+(* Simple bruteforce algorithm *)
+(* Produces duplicates sometimes *)
+type operand = Plus | Minus | Mult | Div;;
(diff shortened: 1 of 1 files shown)


patricoferris:
Hi @dracconi thanks for the PR, I haven't checked your solution (I'm not brilliant at algorithms) but could you maybe refactor the code to follow the pattern many of the others follow which uses the `>` to indent the solution and then reuses the implemented function on some of the examples in a code block below ? This will render quite nicely on the front-end then https://ocaml.org/learn/tutorials/99problems.html and also will render the output of a "top level" function that takes as input the arguments described in the text such as:

~~~
```ocamltop
f [2; 3; 5; 7; 11];;
```
~~~

This will then output on the site (if the solution is correct ;)) as something like: 

```ocaml
f [2; 3; 5; 7; 11];;
- : string = "2 - 3 + 5 + 7 = 11"
```

Thanks :))

dracconi:
Hi @patricoferris, I've commited the updated thing per request, though I'm not sure if the example should be exactly the same as listed in the example `[2; 3; 5; 7; 11]`, or can it be different?

PR #1609, opened 2021-09-27
Add Functori to list of companies

cagdasbozman:
This PR add Functori to list of companies.


patricoferris:
Thanks @cagdasbozman lgtm, whilst we're still transitioning from ocaml.org -> v3.ocaml.org it would be great to add a companion PR to Functori to v3 as well (in https://github.com/ocaml/v3.ocaml.org-server/tree/main/data/industrial_users/en) :))

PR #1612, opened 2021-10-13
bump opam-repo commit to include 4.13.0

patricoferris:
This bumps the opam-repository commit in order to pick up the 4.13.0 release. Note the commit is the exact commit when 4.13.0 was merged, more recent opam-repository commits would include 4.13.1 but the changelog/notes are not in ocaml.org yet so we would 404 on https://ocaml.org/releases/latest.

Thanks @psafont for noticing :))


patricoferris:
Handled in #1613

PR #1613, opened 2021-10-13
Update front page and create 4.13.1 releage page

Octachron:
This PR adds a release page for 4.13.1, updates the list of releases and the date of the last version of OCaml, and comments out the OCaml workshop announce in News on the frontpage.


Octachron:
I have added an updated Change notes too.

patricoferris:
Ah, this came in just after I PR-ed #1612 but now we can just use the latest opam-repository commit. @Octachron could you bump the repository commits in the Dockerfiles in this PR to the latest opam-repository commit (or at least one that comes after 4.13.1 was released) please? We need to do that because of how we learn of the latest release https://github.com/ocaml/ocaml.org/blob/eca0b931985ca4cf5c646224c2292e53a0d73604/Makefile.common 

Without this https://ocaml.org/releases/latest goes to 4.12.0 :))

Octachron:
Bumped. Sorry, I should have done that earlier, but I had switched on non-release tasks during last week.

patricoferris (approved):
Thanks! Once CI is happy should be good for merging :))

patricoferris:
Thanks again :))

PR #1615, opened 2021-10-26
Documentation: fix and clarify opam 2.1 switch creation

vphantom:
# Issue Description

The opam 2.1 switch creation examples in the release notes for OCaml 4.12.1+ included an incorrect extra `-` and did not match the <2.1 example with two options.

## Changes Made

In the three release notes' opam 2.1 example, I removed the extra `-` and added the missing `nnpchecker` option to match the other example and show how multiple options are now separated by spaces instead of commas.

- [x] ❗ If the PR changes a markdown document in `site/`, a comment was added in https://github.com/ocaml/ood/issues/52 with a link to the PR

### DIFF
--- a/site/releases/4.12.1.md
+++ b/site/releases/4.12.1.md
@@ -34,7 +34,7 @@ opam switch create 4.12.1+flambda+nnpchecker --package=ocaml-variants.4.12.1+opt
 or with opam 2.1:
 
 ```
-opam switch create 4.12.1+flambda -ocaml-variants.4.12.1+options ocaml-option-flambda
+opam switch create 4.12.1+flambda+nnpchecker ocaml-variants.4.12.1+options ocaml-option-flambda ocaml-option-nnpchecker
 ```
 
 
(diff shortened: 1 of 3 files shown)


Octachron (approved):
The typo fix is correct. And I did not think about the comma versus space change in opam 2.1, good catch! 

(For people wondering about the 4.12.0 release page: this page is anterior to the full release of opam 2.1 and does not need to be updated).

patricoferris:
Thanks, the CI errors are unrelated I think. I think opam.ocaml.org went through some maintenance and https://opam.ocaml.org/json/last10_updates.json was probably down. Merging now (thanks for the v3 pointer too)

vphantom:
> Thanks, the CI errors are unrelated I think.

I think the problem is that I edited my message twice very shortly after submitting it and it seems to have triggered three checks in quick succession. Had I known that editing the message caused this, I would've proof-read a few more times before my initial submission. 😉

patricoferris:
Ah no, I think it is because `opam.ocaml.org` is now fully `https` so the `curl` returns nothing:
 - `curl http://opam.ocaml.org/json/last10_updates.json` is empty
 - `curl https://opam.ocaml.org/json/last10_updates.json` works

See https://deploy.ci3.ocamllabs.io/job/2021-10-27/134900-ocluster-build-14a940

patricoferris:
Your changes are live @vphantom, thanks again :)) https://ocaml.org/releases/4.13.1.html

PR #1616, opened 2021-10-27
Update last10_updates to https

patricoferris:
# Issue Description

I believe opam.ocaml.org has been fully swapped to `https` now (i.e. http redirects to https) which means the `curl` for the latest updates no longer works causing the build to fail.

## Changes Made

Change http to https


patricoferris:
Looks like this fixed the `curl` error seen in #1615, merging to unblock the deployment :))

PR #1619, opened 2021-11-22
Added bsc to the list of alternative compilers.

fccm:
Added `bsc` to the list of alternative compilers.

bsc (the rescript compiler) is also able to translate ocaml code to javascript.

### DIFF
--- a/site/releases/4.13.0.md
+++ b/site/releases/4.13.0.md
@@ -86,6 +86,9 @@ targets traditionally associated with other languages:
 
 * [Js_of_ocaml](http://ocsigen.org/js_of_ocaml/) is a stable OCaml
   to JavaScript compiler.
+* [bsc](https://rescript-lang.org/) is a stable OCaml/ReScript
+  to JavaScript compiler
+  (easy to use, well documented and with a supporting community.)
 
 ![](../img/doc.gif "") User's manual
 ------------------------------------


Octachron:
As far as I know, the rescript compiler does not support OCaml 4.13. Moreover, one of the motivations for the rescript split was to free rescript from the grip of OCaml backward compatibility.

fccm:
I just put it there because there was `js_of_ocaml`, so I thought it would be natural to put `bsc` next to it in case someone want a Javascript backend.

`bsc` is far more easy to use than `js_of_ocaml`, so maybe it would be worth talking about it somewhere on the website.

On which page would you recommend to talk about it?

Octachron:
Rescript may support an OCaml syntax (as a legacy feature) but it is not a javascript backend for OCaml, contrarily to js_of_ocaml. Thus Rescript cannot be easier to use than `js_of_ocaml`, if the aim is to compile OCaml code.

It might make sense to have a page describing languages inspired or derived from OCaml like F# or Rescript.

fccm:
This is only a personal opinion. I've tried hard several times to learn `js_of_ocaml` but never succeed to make even a hello-world work. There was no documentation, no simple example to start with, the js output is cryptic, and there is no community to get some help with. There aren't also libs that come with it.

This is all the opposite for `bsc`, there's lot of docs, lots of examples and articles, lots of libs, and there's a community that helps you to start. The output is very readable. Beginning with `bsc` and getting things work was not painful for me, it was enjoyable.

This is only a personal opinion, you don't have to share it. But I know that I'm not the only one with such experience, because I read so several times on the forum about it. @lefessan also shared with me that his students have difficulties to use js_of_ocaml.

I don't want to discredit `js_of_ocaml` on `ocaml.org`, I just thought that it would be interesting to share that it's possible to **translate ocaml to javascript** with `bsc` in case some people could be interested.

Maybe a full page with an article might make sense, to explain more details than on the release page?

Octachron:
Using hyperbolic arguments (js_of_ocaml has documentations, libraries, hello world examples and a community) doesn't work really well. 

More importantly, you are missing the point that compiling OCaml code is not a supported feature of ReScript.
There is a reason why OCaml is never mentioned on the main landing page of https://rescript-lang.org . It would be a disservice to ReScript developers to mention ReScript as a javascript backend for OCaml: it is not the intended goal of Rescript since the split between Reason and ReScript.

dbuenzli:
> This is only a personal opinion. I've tried hard several times to learn `js_of_ocaml` but never succeed to make even a hello-world work.

@fccm did you try to use it with [`brr`](https://erratique.ch/software/brr) and `dune`'s [support](https://dune.readthedocs.io/en/stable/jsoo.html#jsoo) ? Brr's docs have a [web page howto](https://erratique.ch/software/brr/doc/web_page_howto.html) with a hello world example. 

There are also a few manuals [here](https://erratique.ch/software/brr/doc/index.html#manuals) that detail the OCaml to JavaScript FFI in other terms than `js_of_ocaml`'s website. Brr doesn't try to encode JavaScript objects as phantom types which I (and others it seems) find much more usable. 

> no community to get some help with

People may be less web focused on [`discuss.ocaml.org`](https://discuss.ocaml.org/). But most technical questions about `js_of_ocaml`  do eventually get an answer.

fccm:
Hi Daniel, `brr` looks great, as everything you do!
My opinion on the state of using ocaml and js was indeed outdated and stuck in the past.
I saw there's `melange` and `ocaml-wasm` today in the links in `CAMLBOY`.
It's not really clear which version of rescript melange is a fork from.
https://anmonteiro.com/2021/03/on-ocaml-and-the-js-platform/
it look like the direction I will take after bsc don't support ocaml anymore (I'm still using bsc 8.4.2 and I'm happy with it).

PR #1628, opened 2022-10-12
OCaml 5.0 manual

Octachron:
This is the beta version of OCaml 5.0 manual, except for the binary files that I will update with the actual release of OCaml 5.0 .


tmattio:
Thanks @Octachron!

tmattio:
The deployment is failing because `make production` fails with:
```
Error: Could not reach the end of block ending with "</span>!}}".
```

This seems to be coming from `<span class="syntax-token">{{!</span>` and possibly caused because `{{!` is interpreted by [MPP](https://github.com/ocaml/MPP-language-blender), which expects a closing tag `!}}`.

Before I investigate further, do you know the build was passing before? (I see that it's also present in the 4.14 manual)

Octachron:
The issue is that the manual pattern which is used to exclude the manual files from the MPP processing
```
MANUAL_DIRS=site/releases/[1-9].[0-9][0-9]
```
doesn't work with the `5.0` version scheme.

PR #1629, opened 2022-10-17
manual: fix manual_dirs patterns for 5.0

Octachron:
This is a simple fix which is expand the pattern used to detect the manual repositories to work with the 5.0 manual.


tmattio (approved):
Thanks @Octachron!

I had tried `([0-9]?)[0-9]` and `[0-9]+`, neither of which worked, but it seems that `[0-9]*` does the trick.

PR #1632, opened 2023-01-10
OCaml 5: update default links and redirection

Octachron:
This PR updates the version selector in the documentation pages and the `manual` symbolic link to point to the 5.0 version.

### DIFF
--- a/site/api
+++ b/site/api
@@ -1,1 +1,1 @@
-releases/4.14/api
+releases/5.0/api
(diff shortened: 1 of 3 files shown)


tmattio:
Thanks @Octachron! I think the symlink of `api` would need to be updated as well, right?

Octachron:
Indeed, fixed, thanks!

tmattio (approved):

PR #1636, opened 2023-09-15
Bump opam-repository and OCaml compiler version in Dockerfiles

tmattio:
See https://github.com/ocaml/infrastructure/issues/72.


Octachron:
Thanks a lot for taking the time to fix the issue that I had missed!

PR #1637, opened 2023-12-06
Add production robots.txt to stop indexing v2 pages

tmattio:
From a user who was Googling for OCaml learning resources and ended up on V2. Reported by @Khady. cc @sabine

image


Khady:
Also about `v2`, it’s not obvious that it’s about the version of ocaml.org and not about the version of the ocaml language for someone who is starting with ocaml now.

sabine:
@Khady Yes, that's why we're currently moving the manual from v2.ocaml.org to ocaml.org

tmattio:
@sabine do you have a rough timeline for that? If it's going to happen in the next few weeks, I suggest we don't merge this PR and just stop indexing all of v2 when the manual + stdlib is in v3.

sabine:
I expect the move will happen this year.

